PR #2, opened 2024-02-13
Restructure | Bump RET & Scrypto & UniFFI version | Add CI (test.yml)

CyonAlexRDX:
* Bump RET version
* Bump Scrypto version (same as used by new RET)
* Remove unneeded crate dependency `transaction` (part of Scrypto)
* Remove Workspace => mono crate (already was mono crate)
* Restructure files
* Changed naming convention of imported Radix types to `Scrypto` prefix for types from radixdlt/radixdlt-scrypto package and `RET` prefix for types from radix-engine-toolkit package.
* Replace `rust.yml` with `test.yml` that is based on the same [contents as RET](https://github.com/radixdlt/radix-engine-toolkit/blob/main/.github/workflows/test.yml), but with `tarpaulin` and `upload to codecov` added


uniffi.toml
CyonAlexRDX, line 1: unchanged, just moved and removed Python bindings...

GhenadieVP (approved):

src/profile/v100/networks/network/network_id.rs
```diff
@@ -114,40 +114,40 @@ impl std::fmt::Display for NetworkID {
...
-            Zabanet => NativeNetworkDefinition::zabanet(),
-            Enkinet => NativeNetworkDefinition {
+            Mainnet => ScryptoNetworkDefinition::mainnet(),
+            Stokenet => ScryptoNetworkDefinition::stokenet(),
+            Adapanet => ScryptoNetworkDefinition::adapanet(),
+            Nebunet => ScryptoNetworkDefinition::nebunet(),
+            Kisharnet => ScryptoNetworkDefinition::kisharnet(),
+            Ansharnet => ScryptoNetworkDefinition::ansharnet(),
+            Zabanet => ScryptoNetworkDefinition::zabanet(),
+            Enkinet => ScryptoNetworkDefinition {
```
micbakos-rdx, line 130: Don't such definitions exist in RET? This is why you added some custom ones?

uniffi.toml
```diff
@@ -0,0 +1,53 @@
+[bindings.swift]
+module_name = "Sargon"
+
+[bindings.kotlin]
+package_name = "sargon"
```
micbakos-rdx, line 5: 👍

src/profile/v100/networks/network/network_id.rs
Sajjon: mm does not exist

PR #4, opened 2024-02-14
Swift Build Script | CD release pipeline

CyonAlexRDX:


src/core/types/decimal.rs
```diff
@@ -11,26 +11,26 @@ use radix_engine_common::math::Decimal as ScryptoDecimal;
     DeserializeFromStr,
     uniffi::Record,
 )]
-pub struct Decimal {
+pub struct Decimal192 {
```
micbakos-rdx, line 14: @CyonAlexRDX Why is this needed?

CyonAlexRDX: Clash with Swift Foundations type with same name. So in iOS app in a file where we import
Both Sargon and Foundation it does not compile. And I don't wanna have to write Sargon.Decimal all over the place.

PR #10, opened 2024-02-20
Wrap ret 2 addresses

CyonAlexRDX:


apple/Sources/Sargon/Extensions/Methods/Addresses+Wrap+Functions.swift
Sajjon, line 1: 🎉

apple/Sources/UniFFI/Sargon.swift
CyonAlexRDX, line 1: I realize we might actually `git rm --cached Sargon.swift` this and then gitignore it. Because it is added part of all releases, and will be present locally...

src/core/types/decimal192.rs
```diff
@@ -19,35 +19,6 @@ impl crate::UniffiCustomTypeConverter for InnerDecimal {
...
-    Copy,
-    Debug,
-    PartialEq,
-    Eq,
-    PartialOrd,
-    Ord,
-    Hash,
-    derive_more::Display,
-)]
-pub struct InnerDecimal(pub(crate) ScryptoDecimal192);
```
CyonAlexRDX, line 40: I had this because I (incorrectly) thought we could not add custom UniFFI conversion for external types, but we can, so no need for this level of indirection.

src/profile/v100/address/accesscontroller_address.rs
```diff
@@ -0,0 +1,217 @@
...
+    Eq,
+    Hash,
+    derive_more::FromStr,
+    derive_more::Display,
+    SerializeDisplay,
+    DeserializeFromStr,
+    uniffi::Record,
+)]
+#[display("{secret_magic}")]
+pub struct AccessControllerAddress {
```
CyonAlexRDX, line 33: Declare "manually" and NOT by macro so we can add documentation to it! which is not possible with `macro_rules!` (or at least too hard for me...).

src/profile/v100/address/wrap_ret_address.rs
```diff
@@ -0,0 +1,150 @@
...
+};
+
+pub trait AddressViaRet: Sized {
+    fn new(
+        node_id: impl Into<ScryptoNodeId>,
+        network_id: NetworkID,
+    ) -> Result<Self>;
+}
+
+macro_rules! decl_ret_wrapped_address {
```
CyonAlexRDX, line 27: This macro creates the needed UniFFI converter for each RET address, using `String` as `Builtin` type. It impl `From` the wrapped RET address type into the Sargon address type. and all UniFFI exported function

src/sargon.udl
CyonAlexRDX, line 1: unfortunate but, c'est la vie, declaring the Ret canonical address as UniFFI convertible via String make implementation of our wrapping of these types easy and allows for safe Rust code, and safe Swift/Kotlin models, so best of two worlds, at the cost of having to declare them here in the UDL file...

src/profile/v100/address/wrap_ret_address.rs
```diff
@@ -0,0 +1,150 @@
...
+        }
+    };
+}
+
+decl_ret_wrapped_address!(
+    AccessControllerAddress,
+    RetAccessControllerAddress,
+    accesscontroller
+);
+decl_ret_wrapped_address!(AccountAddress, RetAccountAddress, account);
```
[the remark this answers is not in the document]
CyonAlexRDX: Ah wrong! we CAN include doc strings in macros, change incoming.

src/profile/v100/entity/abstract_entity_type.rs
```diff
@@ -19,37 +19,8 @@ use radix_engine_common::types::EntityType as ScryptoEntityType;
 )]
 #[repr(u32)] // it is u32 since used in Derivation Paths (CAP26) where each component is a u32.
 pub enum AbstractEntityType {
```
GhenadieVP, line 21: So this is meant only to describe an Entity in the sense of either Account or Identity, but not other address kind?

apple/Sources/Sargon/Extensions/Methods/Addresses+Wrap+Functions.swift
```diff
@@ -0,0 +1,662 @@
...
+				of.stokenetGC,
+				of.stokenetCandy,
+			]
+		}
+	}
+#endif
+
+#if DEBUG
+	/// TODO: Declare in Rust land? Make non-DEBUG?
+	public enum Address: Hashable, Equatable, Sendable {
```
GhenadieVP, line 289: This will most probably be needed in the iOS Wallet, where we do need to merge together different kinds of addresses in some places.

GhenadieVP (approved):
Very very nice 👏!

_typos.toml
```diff
@@ -9,6 +9,7 @@ extend-exclude = [
 
 [default.extend-identifiers]
 inout = "inout"
+pool_tdx_2_1c4ml86h8lvfk7jma0jy0vksh8srcxhmtax8nd3aur29qtd2k2wmlzk = "pool_tdx_2_1c4ml86h8lvfk7jma0jy0vksh8srcxhmtax8nd3aur29qtd2k2wmlzk"
```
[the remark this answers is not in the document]
CyonAlexRDX: Typos CLI tool corrected that address 😅😅 so had to ignore it

apple/Sources/Sargon/Extensions/Methods/Addresses+Wrap+Functions.swift
CyonAlexRDX: Will be added in Rust land immediately when I see we really need it. I still - naive perhaps - think we can avoid having it in wallets. Or at least a smaller tagged union (fewer cases)

Sajjon:
UniFFI macro invocations introduces a lot of false negatives, decreasing the accuracy in reported code coverage, I've asked about if we can improve accuracy in https://github.com/xd009642/tarpaulin/issues/351#issuecomment-1957736118

Will merge this PR as is, nothing we can do.

PR #13, opened 2024-02-23
Android building and CI

micbakos-rdx:
1. Added most of the steps needed to develop, build and run locally the android libraries.
2. The jvm project is gradle based so all scripts are enclosed in the `build.gradle.kts` files. 
3. Two modules are produced: `sargon-android` and `sargon-desktop-bins`. The latter is a test runtime dependency clients need to include in order for junit tests to run.
4. Unit tests will only run successfully in an apple silicon machine. In the future we should attempt expanding the support for more architectures.
5. All binaries are published with maven in Github's maven repo (only for internal use).

Last but not least, the CD workflows currently publish libraries with version names in the form of:
```
<version-name>-<commit-sha>
```
This was needed in order for me to test CD and publish libraries with the same version name.


jvm/app/src/main/java/com/radixdlt/sargon/android/MainActivity.kt
```diff
@@ -0,0 +1,227 @@
...
+                        modifier = Modifier.fillMaxWidth(),
+                        value = accountName,
+                        onValueChange = { accountName = it },
+                        label = {
+                            Text(text = "New Account Name")
+                        },
+                        singleLine = true,
+                        keyboardOptions = KeyboardOptions(imeAction = ImeAction.Done),
+                        keyboardActions = KeyboardActions(onDone = {
+                            walletState?.createAndSaveNewAccount(
```
CyonAlexRDX, line 114: 🚀

CyonAlexRDX (approved):
nice, great job!!! ⭐️ 🚀

PR #18, opened 2024-03-06
Kotlin extensions set 1

micbakos-rdx:
Extension methods, properties and factories are added to most of the objects Uniffi exports that also exist in `apple/Sources`. Some additional extensions added to also support some unit tests. I will add more in the future when @CyonAlexRDX does the same for swift.

Also created sample classes that are exposed either in unit tests (with `@VisibleForTesting`) or for compose previews with `PreviewParameterProvider` (Although previews on compose are not working with sargon yet, these may become useful when we can make Android studio work). Samples are not exposed in production code.

All tests written for bindgen and some more now exist in `sargon-android` library. I had to do some changes on how the desktop binaries are built either when we publish them in maven (using `sargon-desktop-release`) or when we use them for local development (using `sargon-desktop-debug`). For this I had to share the build logic for desktop binaries by creating a custom plugin living in `jvm/buildSrc` called `cargoDesktop`. This detail is not important to the developer of the library as they can always use the same commands to run unit tests like:

```sh
cd jvm
# For running unit tests with sargon built in debug mode
./gradlew sargon-android:testDebugUnitTest

# For running unit tests with sargon built in release mode
./gradlew sargon-android:testReleaseUnitTest
```


jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/BagOfBytes.kt
```diff
@@ -0,0 +1,25 @@
...
+
+package com.radixdlt.sargon.extensions
+
+import com.radixdlt.sargon.BagOfBytes
+import com.radixdlt.sargon.newBagOfBytesFrom
+import kotlin.random.Random
+
+fun bagOfBytesOf(byteArray: ByteArray) = newBagOfBytesFrom(bytes = byteArray)
+
+fun randomBagOfBytes(byteCount: Int) = Random.nextBytes(size = byteCount).toBagOfBytes()
```
CyonAlexRDX, line 11: Is this CSRNG (**Cryptographically Secure** Random Number Generator)? I think not? Hmm maybe best put `randomBagOfBytes` being VisibilityForTesting (only) or otherwise DEBUG only compilation flag? What we need to protect against is using non-secure Randomness for Mnemonics. 

I dont think there is any API surface yet where you can pass a BagOfBytes for entropy which needs to be CSRNG but just want us to be mindful about that.

jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/Decimal192.kt
```diff
@@ -0,0 +1,114 @@
...
+
+operator fun Decimal192.plus(other: Decimal192): Decimal192 = decimalAdd(lhs = this, rhs = other)
+
+operator fun Decimal192.minus(other: Decimal192): Decimal192 = decimalSub(lhs = this, rhs = other)
+
+operator fun Decimal192.times(other: Decimal192): Decimal192 = decimalMul(lhs = this, rhs = other)
+
+operator fun Decimal192.div(other: Decimal192): Decimal192 = decimalDiv(lhs = this, rhs = other)
+
+operator fun Decimal192.compareTo(other: Decimal192): Int {
```
CyonAlexRDX, line 107: is this needed? I thought UniFFI would generate `compareTo` for Kotlin? Or was that genenerated version bad?

jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/PublicKey.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.radixdlt.sargon.ed25519PublicKeyToBytes
+import com.radixdlt.sargon.ed25519PublicKeyToHex
+import com.radixdlt.sargon.newEd25519PublicKeyFromBytes
+import com.radixdlt.sargon.newEd25519PublicKeyFromHex
+import com.radixdlt.sargon.newSecp256k1PublicKeyFromBytes
+import com.radixdlt.sargon.newSecp256k1PublicKeyFromHex
+import com.radixdlt.sargon.secp256k1PublicKeyToBytes
+import com.radixdlt.sargon.secp256k1PublicKeyToHex
+
+fun PublicKey.Ed25519.Companion.init(hex: String): PublicKey.Ed25519 =
```
CyonAlexRDX, line 14: does UniFFI created the nested type `PublicKey.Secp256k1` for you?

CyonAlexRDX, line 14: does UniFFI created the nested type `PublicKey.Secp256k1` for you?

CyonAlexRDX (approved):
Nice, good job! feel free to merge and bear the CSRNG remarks in mind and ensure to be addresses later when Mnemonics are created (either explicitly or implicitly)

CyonAlexRDX (approved):
Nice, good job! feel free to merge and bear the CSRNG remarks in mind and ensure to be addresses later when Mnemonics are created (either explicitly or implicitly)

jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/BagOfBytes.kt
micbakos-rdx: Right will do that

jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/Decimal192.kt
micbakos-rdx: No Uniffi only produces records that have equals and hashcode. But in order to make something abide by <, >, <=, >= you need to make it extend the `operator fun compareTo`. This has to return an int. So my only option was to use the subtraction in order to find the diff.

jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/PublicKey.kt
micbakos-rdx: Yes

micbakos-rdx: Yes

jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/BagOfBytes.kt
micbakos-rdx: @CyonAlexRDX I could use [SecureRandom](https://developer.android.com/reference/kotlin/java/security/SecureRandom) provided by the android platform, but since this is only used in tests, I think it suffices to make it visible for tests only

PR #23, opened 2024-03-13
Wrap RET: batch 8 (format Decimal)

CyonAlexRDX:
Add three format methods to `Decimal192` - also UniFFI exported and sugared in Swift.
- `formatted_plain`
- `formatted_engineering_notation`
- `formatted` (using the previously mentioned)

I've used @kugel3 s implementation in Swift as ref, it is almost a 1:1 mapping. I've fixed a crash caused by an overflow bug the Swift impl has, when formatting Decimal::max. I will follow up with  bug fix in Swift.

Further changes:
- Changed impl of `abs` to never panic, in case of `Decimal::min().abs()` instead of panic I return `Decimal::max()` which seems like the correct thing to do???
- For TEST ONLY: Impl `From<str>` for Decimal and Address types, greatly simplifying writing of tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,7 +73,6 @@ bip39 = { version = "2.0.0", features = ["serde"] }
 time-util = { version = "0.3.4", features = ["chrono"] }
 assert-json-diff = "2.0.2"
 url = { version = "2.5.0", features = ["serde"] }
-num-format = "0.4.4"
 paste = "1.0.14"
 clap = { version = "4", default-features = false, features = ["std", "derive"] }
 camino = "1.0.8"
--- a/scripts/ios/release.sh
+++ b/scripts/ios/release.sh
@@ -20,7 +20,8 @@ function last_tag() {
     local out=`git tag --sort=taggerdate | tail -1`
     echo $out
 }
-echo "🚢 🏷️  Last tag: $(last_tag)"
+LAST_TAG=$(last_tag)
+echo "🚢 🏷️  Last tag: $LAST_TAG"
 
 # one liner from: https://stackoverflow.com/a/8653732
 NEXT_TAG=$(echo $(last_tag) | awk -F. -v OFS=. 'NF==1{print ++$NF}; NF>1{if(length($NF+1)>length($NF))$(NF-1)++; $NF=sprintf("%0*d", length($NF), ($NF+1)%(10^length($NF))); print}')
@@ -55,7 +56,7 @@ echo "🚢 🏷️ 📡 Pushing tag: $(NEXT_TAG), but only tag, not commit."
 SWIFT_SARGON_BINARY_ASSET_NAME="libsargon-rs.xcframework.zip" 
 
 GH_RELEASE_TITLE="Sargon Swift Only v$NEXT_TAG"
-RELEASE_CMD="gh release create $NEXT_TAG '$XCFRAME_ZIP_PATH#$SWIFT_SARGON_BINARY_ASSET_NAME' --generate-notes --title '$GH_RELEASE_TITLE'"
+RELEASE_CMD="gh release create $NEXT_TAG '$XCFRAME_ZIP_PATH#$SWIFT_SARGON_BINARY_ASSET_NAME' --generate-notes --notes-start-tag $LAST_TAG --title '$GH_RELEASE_TITLE'"
 eval $RELEASE_CMD
 
 echo "🚢  End of release script ✅"
--- a/src/core/types/decimal192.rs
+++ b/src/core/types/decimal192.rs
@@ -441,21 +441,6 @@ fn trailing_zero_count_of(s: impl AsRef<str>) -> usize {
         .unwrap_or(str.len())
 }
 
-#[cfg(test)]
-#[test]
-fn test_trailing_zero_count_of() {
-    let test = |s: &str, exp: usize| assert_eq!(trailing_zero_count_of(s), exp);
-
-    test("", 0);
-    test("1", 0);
-    test("0", 1);
-    test("100", 2);
-    test("1001", 0);
-    test("90000", 4);
-    test("9000.0", 1);
-    test("9.000", 3);
-}
-
 fn insert_grouping_separator_into(s: &mut String, separator: String) {
     let digits = s.len();
     let zeroes_per_thousand = 3;
@@ -472,28 +457,6 @@ fn insert_grouping_separator_into(s: &mut String, separator: String) {
     }
 }
 
-#[cfg(test)]
-#[test]
-fn test_insert_grouping_separator_into() {
-    let test_w = |s: &str, exp: &str, sep: char| {
-        let mut string = s.to_owned();
-        insert_grouping_separator_into(&mut string, sep.to_string());
-        assert_eq!(string, exp.to_owned())
-    };
-    let test = |s: &str, exp: &str| test_w(s, exp, ' ');
-
-    test("", "");
-    test("1", "1");
-    test("22", "22");
-    test("333", "333");
-    test("4444", "4 444");
-    test("123456789", "123 456 789");
-    test("12345678987654321", "12 345 678 987 654 321");
-
-    test_w("123456789", "123.456.789", '.');
-    test_w("123456789", "123,456,789", ',');
-}
-
 fn split_str(s: impl AsRef<str>, after: i8) -> (String, String) {
     let mut s = s.as_ref().to_owned();
     if after <= 0 {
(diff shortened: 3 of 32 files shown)


src/core/types/decimal192.rs
```diff
@@ -207,38 +413,118 @@ impl Decimal {
...
+    /// assert_eq!(SUT::two().neg().abs(), SUT::two());
+    /// assert_eq!(SUT::three().abs(), SUT::three());
+    /// assert_eq!(SUT::max().abs(), SUT::max());
+    /// ```
+    ///
+    #[inline]
     pub fn abs(&self) -> Self {
-        self.native().checked_abs().expect("Expected clients of Sargon to not use so large negative numbers (Self::MIN).").into()
+        if self == &Self::min() {
+            Self::max()
```
Sajjon, line 507: Is this reasonable to do?! I think it is! We now return `Decimal::max()` instead of CRASH if we call `abs` on `Decimal::min()`, they are aaaaaaalmost eq. We also note that it is probably never going to happen that our Wallets use `Decimal::min()`, we almost never use negative numbers at all...

src/core/types/decimal192.rs
```diff
@@ -270,16 +613,24 @@ impl Decimal {
     /// - Panic if the number of decimal places is not within [0..SCALE(=18)]
     pub fn round_with_mode(
         &self,
-        decimal_places: i32,
```
Sajjon, line 273: does not make sense to round with a NEGATIVE number of many decimal places :P ... I used `i32` since that is what Scrypto uses... strange!

src/core/types/decimal192.rs
```diff
@@ -270,16 +613,24 @@ impl Decimal {
...
             .ok_or(CommonError::DecimalError)
             .map(Self::from)
     }
 }
 
+#[cfg(test)]
+impl From<&str> for Decimal192 {
+    /// TEST ONLY
+    fn from(value: &str) -> Self {
+        value.parse().unwrap_or_else(|_| panic!("Test failed since the passed in str is not a valid Decimal192: '{}'", value))
```
Sajjon, line 630: N.B. TEST ONLY! Very nice to have this, since we can just use "123.456" for any method/function which accepts `impl Into<Decimal>`, which allows for us to skip writing the very much more verbose "123.456".parse::<Decimal192>.unwrap().

src/core/types/decimal192.rs
```diff
@@ -325,11 +676,280 @@ impl Decimal192 {
...
+    /// ```
+    /// extern crate sargon;
+    /// use sargon::prelude::*;
+    /// #[allow(clippy::upper_case_acronyms)]
+    /// type SUT = Decimal192;
+    ///
+    /// assert_eq!("12345678.975".parse::<SUT>().unwrap().formatted(LocaleConfig::default(), 8, true), "12.345679 M");
+    /// ```
+    ///
+    pub fn formatted(
```
Sajjon, line 927: The most important method to review....

src/core/types/decimal192.rs
```diff
@@ -770,6 +1413,441 @@ mod test_decimal {
...
+        test("123", "123");
+
+        test("123456.4321", "123,456.43");
+        test("12345.4321", "12,345.432");
+        test("1234.4321", "1,234.4321");
+        test("123.4321", "123.4321");
+    }
+
+    #[test]
+    fn format_decimal() {
```
Sajjon, line 1631: [Gustafs tests in Swift](https://github.com/radixdlt/babylon-wallet-ios/blob/main/RadixWalletTests/EngineKitTests/DecimalTests.swift#L276-L495) + some more

src/core/types/decimal192.rs
```diff
@@ -770,6 +1413,441 @@ mod test_decimal {
...
+        let test_ = |decimal: SUT, exp: &str| {
+            let locale = LocaleConfig::english_united_states();
+            let actual = decimal.formatted(locale, 8, false);
+            assert_eq!(actual, exp);
+        };
+        let test = |x: &str, exp: &str| test_(SUT::from(x), exp);
+
+        test_(SUT::max(), "3.138e39");
```
Sajjon, line 1639: `SUT::max()` test added by me, crashed in Swift, will fix in a PR.

src/core/types/locale_config.rs
```diff
@@ -16,46 +16,30 @@ impl LocaleConfig {
...
     }
 }
 
-impl LocaleConfig {
-    /// Tries to create a
-    /// A BCP-47 language identifier such as `"en_US_POSIX"`, `"sv_FI"` or `"zh_Hant_MO"`,
-    /// see: [list][list]
-    ///
-    /// [list]: https://docs.rs/num-format/0.4.4/src/num_format/locale.rs.html#5565-6444
-    pub fn from_identifier(identifier: impl AsRef<str>) -> Result<Self> {
```
Sajjon, line 42: We will just let Swift / Android create an instance of `LocaleConfig`...

src/profile/v100/address/non_fungible_local_id.rs
```diff
@@ -132,6 +132,14 @@ impl FromStr for NonFungibleLocalId {
     }
 }
 
+#[cfg(test)]
+impl From<&str> for NonFungibleLocalId {
+    /// TEST ONLY
+    fn from(value: &str) -> Self {
+        value.parse().unwrap_or_else(|_| panic!("Test failed since the passed in str is not a valid NonFungibleLocalId: '{}'", value))
```
Sajjon, line 139: TEST ONLY! Greatly simplifies creation from `str` in tests :)

src/profile/v100/address/non_fungible_resource_address.rs
```diff
@@ -88,6 +88,14 @@ macro_rules! decl_specialized_address {
                 }
             }
 
+            #[cfg(test)]
+            impl From<&str> for $specialized_address_type {
+                /// TEST ONLY
+                fn from(value: &str) -> Self {
+                    value.parse().expect(&format!("Test failed since the passed in str is not a valid address: '{}'", value))
```
Sajjon, line 95: TEST ONLY! Greatly simplifies creation from `str` in tests :)

src/profile/v100/address/wrap_ret_address.rs
```diff
@@ -87,6 +87,14 @@ macro_rules! decl_ret_wrapped_address {
                 }
             }
 
+            #[cfg(test)]
+            impl From<&str> for [< $address_type:camel Address >] {
+                /// TEST ONLY
+                fn from(value: &str) -> Self {
+                    value.parse().expect(&format!("Test failed since the passed in str is not a valid address: '{}'", value))
```
Sajjon, line 94: TEST ONLY! This is for ALL our Address types. Greatly simplifies creation from `str` in tests :)

src/wrapped_radix_engine_toolkit/high_level/manifest_building/manifest_assets_transfers.rs
```diff
@@ -513,14 +355,14 @@ mod tests {
         CALL_METHOD
             Address("account_rdx16xlfcpp0vf7e3gqnswv8j9k58n6rjccu58vvspmdva22kf3aplease")
             "withdraw_non_fungibles"
-            Address("resource_rdx1tknxxxxxxxxxradxrdxxxxxxxxx009923554798xxxxxxxxxradxrd")
```
Sajjon, line 516: ops... I was using a FUNGIBLE ResourceAddress in a Non-Fungible context...

src/wrapped_radix_engine_toolkit/high_level/sargon_specific_types/assets_transfers/per_asset/per_asset_fungible_resource.rs
```diff
@@ -3,13 +3,13 @@ use crate::prelude::*;
 #[derive(Clone, Debug, PartialEq, Eq, Hash, uniffi::Record)]
 pub struct PerAssetFungibleResource {
     pub resource_address: ResourceAddress,
-    pub divisibility: Option<i32>,
```
Sajjon, line 6: does not make sense for divisibility to be negative...

src/wrapped_radix_engine_toolkit/high_level/sargon_specific_types/assets_transfers/per_asset/per_asset_transfers_of_non_fungible_resource.rs
```diff
@@ -41,7 +41,7 @@ impl PerAssetTransfersOfNonFungibleResource {
 impl PerAssetTransfersOfNonFungibleResource {
     pub(crate) fn sample_mainnet() -> Self {
         Self::new(
-            ResourceAddress::sample_mainnet_xrd(),
+            NonFungibleResourceAddress::sample_mainnet(),
```
Sajjon, line 44: ops... I was using a FUNGIBLE ResourceAddress in a Non-Fungible context...

src/wrapped_radix_engine_toolkit/low_level/transaction_manifest/execution_summary/transaction_manifest_execution_summary.rs
Sajjon, line 1: drastically decreased amount of code thanks to `From<str>` + enum constructors! :)

src/wrapped_radix_engine_toolkit/low_level/transaction_manifest/execution_summary/transaction_manifest_execution_summary.rs
```diff
@@ -169,9 +169,9 @@ mod tests {
                 ],
                 FeeLocks::default(),
                 FeeSummary::new(
-                    "0.4077857".parse::<Decimal>().unwrap(),
-                    "0.1307814".parse::<Decimal>().unwrap(),
-                    "0.52433013014".parse::<Decimal>().unwrap(),
+                    "0.4077857",
```
Sajjon, line 172: 🎉

Cargo.toml
```diff
@@ -73,7 +73,6 @@ bip39 = { version = "2.0.0", features = ["serde"] }
 time-util = { version = "0.3.4", features = ["chrono"] }
 assert-json-diff = "2.0.2"
 url = { version = "2.5.0", features = ["serde"] }
-num-format = "0.4.4"
```
CyonAlexRDX, line 76: wont use identifier based, will let iOS  / Android specify decimal separator and grouping separator.

scripts/ios/release.sh
```diff
@@ -55,7 +56,7 @@ echo "🚢 🏷️ 📡 Pushing tag: $(NEXT_TAG), but only tag, not commit."
 SWIFT_SARGON_BINARY_ASSET_NAME="libsargon-rs.xcframework.zip" 
 
 GH_RELEASE_TITLE="Sargon Swift Only v$NEXT_TAG"
-RELEASE_CMD="gh release create $NEXT_TAG '$XCFRAME_ZIP_PATH#$SWIFT_SARGON_BINARY_ASSET_NAME' --generate-notes --title '$GH_RELEASE_TITLE'"
+RELEASE_CMD="gh release create $NEXT_TAG '$XCFRAME_ZIP_PATH#$SWIFT_SARGON_BINARY_ASSET_NAME' --generate-notes --notes-start-tag $LAST_TAG --title '$GH_RELEASE_TITLE'"
```
CyonAlexRDX, line 59: right not the notes are always from first commit to new tag, instead of from last tag to new tag.

micbakos-rdx (approved):

PR #24, opened 2024-03-15
Wrap RET: batch 9 (ManifestSummary)

CyonAlexRDX:
(Target branch is `Batch 8`, merge #23 first )

Impl of `ManifestSummary`, you probably want to review only the first commit.

The second commit just makes Address types impl Copy, so we can drop a lot of `clone()` calls.

The third commit is smaller and similar, adds Copy for PublicKeys

### DIFF
--- a/src/wrapped_radix_engine_toolkit/low_level/manifest_summary.rs
+++ b/src/wrapped_radix_engine_toolkit/low_level/manifest_summary.rs
@@ -1,33 +1,72 @@
 use crate::prelude::*;
 
+/// A summary of the manifest
 #[derive(Clone, Debug, PartialEq, Eq, uniffi::Record)]
 pub struct ManifestSummary {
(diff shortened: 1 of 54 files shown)


src/wrapped_radix_engine_toolkit/low_level/manifest_summary.rs
```diff
@@ -1,33 +1,72 @@
 use crate::prelude::*;
 
+/// A summary of the manifest
 #[derive(Clone, Debug, PartialEq, Eq, uniffi::Record)]
 pub struct ManifestSummary {
-    pub addresses_of_accounts_deposited_into: Vec<AccountAddress>,
+    /// Addresses of accounts withdraws from in the manifest.
```
micbakos-rdx, line 6: ✏️  -> `withdrawn`

micbakos-rdx (approved):

PR #27, opened 2024-03-19
Mix (mostly Swift Test)

CyonAlexRDX:
* Rust + UniF: Extract the `#[uniffi::export]` into specific file, for *some* types (we should do it for *all*!), e.g. `bag_of_bytes_uniffi_fn.rs`, we SHOULD use the same suffix: `_uniffi_fn` for all types, which we can add a RegExp for in our Precommit hook and only run Swift / Kotlin tests if Rust files ending with `_uniffi_fn.rs` has changed (CI always runs all tests ofc...), this would allow faster commits, e.g. if I just add a new method in `bag_of_bytes.rs` but have not touched UniFFI, then no need to run Swift / Kotlin tests....
* Rust + UniF: Make it possible to construct a `Secp256k1PublicKey` from uncompressed bytes
* Rust + UniF: Make it possible to get uncompressed bytes from `Secp256k1PublicKey`
* UniFFI: Hide Decimal192 func `formatted_engineering_notation`, it was never meant to be used "publically" - also document that it will panic for numbers less than 1 (after `abs`)
* UniFFI: Add some more sample values for addresses, ensuring ALL address types have these four sample values: `sampleMainnet`, `sampleMainnetOther`, `sampleStokenet`, `sampleStokenetOther`
* UniFFI: Profile - add `to_string` and `to_debug_string` funcs 
* Swift: Define a `SargonModel` protocol
* Swift: Add `Test` test base class which is generic over `SUT: SargonModel` with impl of equality, hash tests, making it super easy to test SargonModels!
* Swift: Add `AddressTest` which is generic `SUT: AddressProtocol` which inherits from `Test<SUT>` and tests bech32 roundtrip, making it super easy to test Address types!
* Swift: Add the four sample values requirement on `AddressProtocol`, and auto impl of `CaseIterable`
* Swift: Add `CustomDump` dependency and using `XCTAssertNoDifference` instead of `XCTAssertEqual` in Swift Sargon tests for prettier diffing
* Swift: Data + Hex utils


apple/Sources/Sargon/Extensions/Methods/Crypto/Secp256k1PublicKey+Wrap+Functions.swift
```diff
@@ -0,0 +1,56 @@
...
+    /// Use `uncompressedData` for uncompressed format (65 bytes)
+	public var compressedData: Data {
+		// Rust: `secp256k1_public_key_to_bytes`
+		secp256k1PublicKeyToBytes(publicKey: self)
+	}
+    
+    /// Returns the key on **uncompressed** form (65 bytes)
+    ///
+    /// Use `compressedData` for compressed format (33 bytes)
+    public var uncompressedData: Data {
```
CyonAlexRDX, line 44: new

micbakos-rdx (approved):

PR #31, opened 2024-03-22
Create maven deployment on each push on develop

micbakos-rdx:
Android and desktop binaries were not being built as they were expecting a tag to be created. For now CD workflows will start on each develop push


CyonAlexRDX (approved):
LGTM

PR #33, opened 2024-03-22
address map_to_network & codecov.yml

CyonAlexRDX:
(100 misleading source branch name)

# Infra
Added `codecov.yml`, trying to see if I can get badges for flags to work....

# Rust/UniFFI Changes
* Add & UniFFI export `network_ids_all` 
* UniFFI export: samples for `StakeClaim`
* Add & UniFFI export `*_address_map_to_network` which translates an address to the specified network. This works for e.g. XRD address on mainnet -> translated to XRD address on Stokenet. This SHOULD probably be kept "DEBUG ONLY" or "somewhat hidden", I'm using them for tests, since...
* Scrypto does not care about networks of addresses! Scrypto is 100% address agnostic! what that means is that, when building a manifest, we can input addresses on **different networks** to the `ScryptoManifestBuilder`, and when called `build(network_id)` all address will be mapped to that network! which might be surprising. We can change the manifest building methods to either return `Result` instead, and we would have to "manually" validate that input to all manifest building methods use same network... or... we could panic! or we can keep it as is (I've added a warning for `stake_claim` one, could be added to all...). I think it might makes sense to keep them "non throwing", since we control the manifest we build, and wallets validate network_id of incoming requests anyway...
* ... I've made use of the `*_map_to_network` method in a `stake_claim` unit test, make sure to read that and ask me if you dont understand it.
* 

# Swift
* Add `mapTo` network method on address for `DEBUG` **only**.
* Add more manifest intstruction tests.
* Ensure `NetworkID.allCases` uses new `networkIdsAll` from UniFFI


README.md
```diff
@@ -1,6 +1,6 @@
 # Sargon
 
-[![codecov](https://codecov.io/github/radixdlt/Sargon/graph/badge.svg?token=EQYDU0XPMX)](https://codecov.io/github/radixdlt/Sargon)
+[![codecov](https://codecov.io/gh/radixdlt/Sargon/graph/badge.svg?token=8QPKIUSAQD)](https://codecov.io/github/radixdlt/Sargon)
```
Sajjon, line 3: we were using the wrong token, but it seems it did not matter. I still cannot produce badges for specific flags: `kotlin` / `swift` / `rust`, but this PR adds a `codecov.yml` file which *might* fix that (Im guessing here...).

apple/Sources/Sargon/Protocols/AddressProtocol.swift
```diff
@@ -37,4 +39,10 @@ extension AddressProtocol {
 	public var description: String {
 		address
 	}
+	
+	/// Returns the`ResourceAddress` of `XRD` on the same network
+	/// as this address.
+	public var xrd: ResourceAddress {
+		ResourceAddress.xrd(on: networkID)
```
Sajjon, line 46: might come in handy...

codecov.yml
```diff
@@ -0,0 +1,18 @@
...
+    wait_for_ci: true
+  max_report_age: off
+  require_ci_to_pass: true
+comment:
+  behavior: default
+  layout: "reach, diff, flags, files"
+  show_carryforward_flags: false
+coverage:
+  precision: 1
+  range: 97...98 # red -> yellow (the inside range) -> green
```
Sajjon, line 12: less than 97: 🔴
97 -> 98: 🟡
99 -> 100: 🟢

codecov.yml
```diff
@@ -0,0 +1,18 @@
...
+flag_management:
+  default_rules:
+    carryforward: true
```
Sajjon, line 15: not sure about this... https://docs.codecov.com/docs/carryforward-flags

src/profile/v100/address/non_fungible_resource_address_uniffi_fn.rs
Sajjon, line 1: split put UniFFI fn

src/wrapped_radix_engine_toolkit/high_level/manifest_building/manifests.rs
```diff
@@ -279,4 +287,43 @@ mod tests {
             "#,
         );
     }
+
+    #[test]
+    fn manifest_builder_switches_network_of_validators_to_that_of_account_address(
```
CyonAlexRDX, line 292: Make sure to fully understand what is going on in this test

CyonAlexRDX: Hmm or maybe we should make it throwing 🤔🤔🤔

CyonAlexRDX: lets defer it, keep it non throwing for now.

micbakos-rdx (approved):

PR #44, opened 2024-03-27
Sendable Swift Types

CyonAlexRDX:
Use `experimental_sendable_value_types` (written by my, in PR https://github.com/mozilla/uniffi-rs/pull/2045 ), which marks structs and enum's `Sendable` inside Sargon (bindgen generated Swift file), so that we can drop LOTS of `@unchecked Sendable`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,6 +3,11 @@
 
 import PackageDescription
 
+let swiftSettings: [SwiftSetting] = [
+	.enableExperimentalFeature("StrictConcurrency"),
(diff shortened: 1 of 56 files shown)


Package.swift
```diff
@@ -49,15 +54,17 @@ let package = Package(
 		.target(
```
CyonAlexRDX, line 54: We want to use Swift settings in SargonUniFFI target as well but we cannot due to https://github.com/mozilla/uniffi-rs/issues/2046

micbakos-rdx (approved):

PR #49, opened 2024-03-29
Codable Addresses, Decimal192, NonFunIDs

CyonAlexRDX:
Swift only:
* move Codable impl of `Decimal192`, all addresses and `NonFungible`IDs into Sargon, needed by iOS wallet due to CacheClient 
* add casting methods on Address
* add `asDouble` which seems it cannot fail, so made into computed property. Should be moved into Rust Sargon...

### DIFF
--- a/apple/Sources/Sargon/Extensions/Methods/Prelude/NonFungibleLocalID+Wrap+Functions.swift
+++ b/apple/Sources/Sargon/Extensions/Methods/Prelude/NonFungibleLocalID+Wrap+Functions.swift
@@ -1,14 +1,13 @@
 import SargonUniFFI
 
 extension NonFungibleLocalID {
-	public func toString() -> String {
+	public func toRawString() -> String {
 		nonFungibleLocalIdAsStr(id: self)
 	}
 	
 	public init(integer value: UInt64) {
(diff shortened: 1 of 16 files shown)


apple/Sources/Sargon/Extensions/Methods/Prelude/NonFungibleLocalID+Wrap+Functions.swift
```diff
@@ -24,18 +23,18 @@ extension NonFungibleLocalID {
...
     
     /// Tries to decode a `NonFungibleLocalID.string`
     ///
-    /// Not to be confused with `init(localId:)` which tries to decode
-    /// a string into any case of `NonFungibleLocalID` 
-	public init(string: String) throws {
-		self = try newNonFungibleLocalIdString(string: string)
+    /// Not to be confused with `init(:String)` which tries to decode
+    /// a string into any case of `NonFungibleLocalID`
+	public static func stringID(_ string: String) throws -> Self {
```
micbakos-rdx, line 36: Exactly 👌 The naming is far better now. I will change it too.

micbakos-rdx (approved):

PR #50, opened 2024-03-29
Kotlin extensions & random samples for addresses

micbakos-rdx:
* Added missing extensions and samples regarding the latest additions.
* Added random sample for each Scrypto address type using macro.

### DIFF
--- a/jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/NonFungibleLocalId.kt
+++ b/jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/NonFungibleLocalId.kt
@@ -1,13 +1,17 @@
 package com.radixdlt.sargon.extensions
 
+import com.radixdlt.sargon.AddressFormat
 import com.radixdlt.sargon.BagOfBytes
+import com.radixdlt.sargon.NonFungibleGlobalId
 import com.radixdlt.sargon.NonFungibleLocalId
 import com.radixdlt.sargon.newNonFungibleLocalIdBytes
 import com.radixdlt.sargon.newNonFungibleLocalIdFromString
 import com.radixdlt.sargon.newNonFungibleLocalIdInt
 import com.radixdlt.sargon.newNonFungibleLocalIdRuid
 import com.radixdlt.sargon.newNonFungibleLocalIdString
+import com.radixdlt.sargon.nonFungibleGlobalIdFormatted
 import com.radixdlt.sargon.nonFungibleLocalIdAsStr
+import com.radixdlt.sargon.nonFungibleLocalIdFormatted
 
 @Throws(SargonException::class)
 fun NonFungibleLocalId.Companion.init(localId: String): NonFungibleLocalId =
@@ -30,3 +34,7 @@ fun NonFungibleLocalId.Companion.fromString(string: String): NonFungibleLocalId 
 
 val NonFungibleLocalId.string: String
     get() = nonFungibleLocalIdAsStr(id = this)
+
+fun NonFungibleLocalId.formatted(
+    format: AddressFormat = AddressFormat.DEFAULT
+): String = nonFungibleLocalIdFormatted(id = this, format = format)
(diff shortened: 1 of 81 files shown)


src/profile/v100/address/wrap_ret_address.rs
```diff
@@ -89,6 +89,15 @@ macro_rules! decl_ret_wrapped_address {
                 address.formatted(format)
             }
 
+            /// Returns a random address in `network_id` as Network
+            #[uniffi::export]
+            pub fn [<new_ $address_type:snake _address_sample_random>](network_id: NetworkID) -> [<$address_type:camel Address >] {
```
Sajjon, line 94: good job impl this! But this violates our principle of just delegating from a `#[uniffi::exported]` function into a non exported on. We can benefit - in the realm of Rust - of having this on each address type!

So split it into non-uniffi, and then just delegate/forward call to it from the `#[uniffi::exported]` function :)

you will also need the double unit test, unit test of non-uniffi func is most important. The uniffi unit test ought to only do a quick "soundness check".

src/profile/v100/address/wrap_ret_address.rs
```diff
@@ -218,6 +227,14 @@ macro_rules! decl_ret_wrapped_address {
                     let sut = SUT::sample();
                     assert_eq!([<$address_type:snake _address_map_to_network>](&sut, sut.network_id()), sut); // unchanged
                 }
+
+                #[test]
+                fn random_address() {
+                    for network_id in NetworkID::all() {
+                        let sut = [<new_ $address_type:snake _address_sample_random>](network_id);
+                        assert_eq!(SUT::from_str(&sut.to_string()).unwrap(), sut); // unchanged
```
Sajjon, line 235: add a comment that the string roundtrip ensures that we have used to correct entity => correct Bech32 HRP, thus ensuring correctness of `EntityType`

I'd also like a `let n = 1000'; ... hashset.len() == n` test, ensuring we are indeed using randomness as the remamning 29 bytes in the node id.

jvm/sargon-android/src/main/java/com/radixdlt/sargon/samples/ResourceAddressSample.kt
```diff
@@ -24,6 +26,10 @@ object ResourceAddressSampleMainnet: Sample<ResourceAddress> {
 
     override fun other(): ResourceAddress = candy
 
+    override fun random(): ResourceAddress = newResourceAddressSampleRandom(
```
Sajjon, line 29: maybe it will be usefull with a `fun random(networkID: NetworkID)` too? And then `random` can call it with Mainnet as default?

jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/NonFungibleLocalId.kt
```diff
@@ -8,25 +9,65 @@ import com.radixdlt.sargon.newNonFungibleLocalIdInt
...
+ * Creates a [NonFungibleLocalId.Str] id. Not to be confused with
+ * [NonFungibleLocalId.Companion.init]
+ *
+ * For example creates an id from: "foo" => `Str(<foo>)`
+ *
+ * @throws SargonException if the [value] bytes are invalid
+ */
 @Throws(SargonException::class)
-fun NonFungibleLocalId.Companion.fromString(string: String): NonFungibleLocalId =
+fun NonFungibleLocalId.Companion.stringId(string: String): NonFungibleLocalId =
```
Sajjon, line 65: 👍

src/profile/v100/address/wrap_ret_address.rs
micbakos-rdx: Right will do

micbakos-rdx: @Sajjon Where are the non-uniffi tests in this file?

CyonAlexRDX (approved):
Nice!

PR #53, opened 2024-04-01
Sargon `main` branch

CyonAlexRDX:
Sargon 1.0.0 will be used for the first Wallet App release we make that uses Sargon. It makes sense to create main branch (and maybe obsolete develop?).

The purpose of this PR is too gather input about some designs on how we have wrapped Radix Engine Toolkit in Sargon, specifically which functions have been made non-throwing (made to panic) - I've chosen to NOT-throwing (i.e. panic instead) in a few places, and then gather input from builders team before we merge this PR. Quiet likely code review of this PR will result in us changing some non-throwing into throwing for some RET wrapped APIs.

PR #54, opened 2024-04-01
SBOR depth validation

CyonAlexRDX:
Validate SBOR depth in constructors of:
* `Instructions`
* `TransactionIntent`
* `SignedIntent`
* `NotarizedTransaction`

Hide Rust constructors of:
* `CompiledNotarizedIntent`
* `NotarizedTransaction`
* `SignedIntent`
* `TransactionIntent`
* `Instructions`
* `InstructionsSecretMagic`
* `TransactionManifest`

*Thus ensuring we always use Failable constructors ("throwing initializers") of these types - where we have asserted SBOR depth*.

Result:
* All seven instances where I've chosen to `panic` (crash if err) in PR https://github.com/radixdlt/sargon/pull/53 are safe to do so, since we have protected against all failable execution paths.
* `debug_print_compiled_notarized_intent` is no longer throwing, since it is not possible to construct an invalid `CompiledNotarizedIntent`.

Extra:
* UniFFI: exported sample value of TransactionHeader
* Rust: possible to create an address from node_id bytes (random uses it), to be able to create some address on some network with some bytes
* bump `typos` version
* Fix broken Swift post processor (after bump to UniFFI 0.27 the format of inits changed)


src/wrapped_radix_engine_toolkit/low_level/transaction_intent.rs
```diff
@@ -36,26 +39,52 @@ impl TransactionIntent {
...
-                value.manifest.instructions().clone(),
-            ),
-            blobs: value.manifest.blobs().clone().into(),
-            message: value.message.into(),
-        }
+        into_scrypto(&value.header, &value.manifest, &value.message)
     }
 }
+
+fn into_scrypto(
```
CyonAlexRDX, line 53: cannot use `From<TransactionIntent>` inside the constructor of `TransactionIntent`... duh... so we use this global fn, which is also used by our `impl From<TransactionIntent> for ScryptoIntent` implementation

src/wrapped_radix_engine_toolkit/low_level/transaction_intent.rs
```diff
@@ -86,6 +111,7 @@ impl HasSampleValues for TransactionIntent {
             TransactionManifest::sample(),
             Message::sample(),
         )
+        .unwrap()
```
CyonAlexRDX, line 114: sample values, safe to unwrap.

src/wrapped_radix_engine_toolkit/high_level/manifest_building/manifest_assets_transfers.rs
```diff
@@ -62,12 +62,7 @@ impl TransactionManifest {
             }
         }
 
-        let scrypto_manifest = builder.build();
-
-        TransactionManifest::from_scrypto(
-            scrypto_manifest,
-            from_account.network_id(),
-        )
+        TransactionManifest::sargon_built(builder, from_account.network_id())
```
CyonAlexRDX, line 65: this unwraps the failable constructor - which is OK, since this manifest is built by US in Sargon, and we know the SBOR depth.

src/wrapped_radix_engine_toolkit/high_level/manifest_building/modify_manifest.rs
```diff
@@ -149,10 +149,10 @@ impl TransactionManifest {
             }
         };
 
-        let instructions = Instructions::from_scrypto(
-            ScryptoInstructions(instructions),
+        let instructions = Instructions::try_from((
+            instructions.as_ref(),
             self.network_id(),
-        );
+        )).expect("Should not have changed depth of SBOR value, thus inserting an instruction should never fail.");
```
CyonAlexRDX, line 155: we are not inserting any Instruction with too great an SBOR depth (we insert CallMethod lock fee, and TX Guarantees (WORKTOP ASSERTS))

src/wrapped_radix_engine_toolkit/low_level/compiled_notarized_intent.rs
```diff
@@ -13,7 +13,7 @@ use crate::prelude::*;
     uniffi::Record,
 )]
 pub struct CompiledNotarizedIntent {
-    pub(crate) secret_magic: BagOfBytes,
```
CyonAlexRDX, line 16: hide constructor, ensuring we always go via `TryFrom<ScryptoNotarizedTransaction>`

src/wrapped_radix_engine_toolkit/low_level/notarized_transaction.rs
```diff
@@ -2,63 +2,106 @@ use crate::prelude::*;
 
 #[derive(Debug, Clone, Eq, PartialEq, uniffi::Record)]
 pub struct NotarizedTransaction {
-    pub signed_intent: SignedIntent,
+    signed_intent: SignedIntent,
```
CyonAlexRDX, line 5: Hide constructor so that we ensure we always go via `new` which now is a failable constructor, performing `compile` - checking SBOR depth.

src/wrapped_radix_engine_toolkit/low_level/sbor_depth_validation.rs
```diff
@@ -0,0 +1,73 @@
...
+    SCRYPTO_SBOR_V1_MAX_DEPTH,
+};
+use radix_engine_interface::prelude::ScryptoValue as ScryptoScryptoValue;
+use sbor::{
+    CustomValue as ScryptoCustomValue,
+    CustomValueKind as ScryptoCustomValueKind, Value as ScryptoValue,
+};
+
+#[cfg(test)]
+fn sbor_value_with_depth<X, Y>(depth: usize) -> ScryptoValue<X, Y>
```
CyonAlexRDX, line 14: Thank you @0xOmarA for providing this!

src/wrapped_radix_engine_toolkit/low_level/signed_intent.rs
```diff
@@ -1,25 +1,72 @@
 use crate::prelude::*;
 
+pub use radix_engine_toolkit::functions::signed_intent::compile as RET_signed_intent_compile;
+
 #[derive(Clone, Debug, PartialEq, Eq, uniffi::Record)]
 pub struct SignedIntent {
-    pub intent: TransactionIntent,
+    intent: TransactionIntent,
```
CyonAlexRDX, line 7: Hide constructor so that we ensure we always go via `new` which now is a failable constructor, performing `compile_signed_intent_with ` - checking SBOR depth.

src/wrapped_radix_engine_toolkit/low_level/transaction_intent.rs
```diff
@@ -4,29 +4,32 @@ use crate::prelude::*;
 #[debug("header:\n{:?}\n\nmessage:\n{:?}\n\nmanifest:\n{}\n\n", self.header, self.message, self.manifest.instructions_string())]
 pub struct TransactionIntent {
     pub header: TransactionHeader,
-    pub manifest: TransactionManifest,
+    manifest: TransactionManifest,
```
CyonAlexRDX, line 7: Hide constructor so that we ensure we always go via `new` which now is a failable constructor, performing `compile_intent_with ` - checking SBOR depth.

src/wrapped_radix_engine_toolkit/low_level/transaction_manifest/instructions/instructions.rs
```diff
@@ -1,36 +1,79 @@
 use crate::prelude::*;
 
+use radix_engine::types::MANIFEST_SBOR_V1_MAX_DEPTH;
+use radix_engine_toolkit::functions::address::decode as RET_decode_address;
+
 #[derive(Clone, Debug, PartialEq, Eq, derive_more::Display, uniffi::Record)]
 #[display("{}", self.instructions_string())]
 pub struct Instructions {
-    pub(crate) secret_magic: InstructionsSecretMagic, // MUST be first prop, else you break build.
+    secret_magic: InstructionsSecretMagic, // MUST be first prop, else you break build.
```
CyonAlexRDX, line 9: Hide constructor so that we ensure we always go via `TryFrom<(&[ScryptoInstruction], NetworkID)> ` which calls `scrypto_decompile` this checking SBOR depth

apple/Tests/TestCases/RET/TransactionManifestTests.swift
```diff
@@ -38,4 +38,106 @@ final class TransactionManifestTests: Test<TransactionManifest> {
         
         XCTAssertNoDifference(summary.addressesOfAccountsRequiringAuth, ["account_tdx_2_1288efhmjt8kzce77par4ex997x2zgnlv5qqv9ltpxqg7ur0xpqm6gk"])
     }
+	
+	func test_from_instructions_string_with_max_sbor_depth_is_ok() throws {
+		let instructionsString = """
+CALL_METHOD
```
micbakos-rdx, line 44: Shouldn't those tests be saved in an rtm file too?

micbakos-rdx (approved):

apple/Tests/TestCases/RET/TransactionManifestTests.swift
Sajjon: sure, can do it after merge, just wanted to get approval from you and Omar first. Also should generalize file location of RTM files, now there were tied to `execution_summary` (in that folder), but should probably be some layer up, maybe... top level? `$PROJECT_ROOT/examples/transactions` or is it strange to put (new folder) `transactions` next to the iOS example app? speaking of which, is it possible to put the Android example app in that folder? 

if not `$PROJECT_ROOT/examples/transactions`, maybe ``$PROJECT_ROOT/resources/transactions` ? `$PROJECT_ROOT/fixtures/transaction` and also put JSON test vectors in `fixtures/testvectors`??

src/wrapped_radix_engine_toolkit/low_level/transaction_manifest/instructions/instructions.rs
```diff
@@ -1,36 +1,79 @@
...
+    type Error = CommonError;
+
+    fn try_from(
+        value: (&[ScryptoInstruction], NetworkID),
+    ) -> Result<Self, CommonError> {
+        let scrypto = value.0;
+        let network_id = value.1;
+
+        // Verify that the instructions has acceptable depth and are compatible
+        _ = instructions_string_from(scrypto, network_id)?;
```
Sajjon, line 51: @0xOmarA maybe actually it would be enough to create a DUMMY `NotarizedTransaction` here and compile it? If that works, we know that this `Instructions` can be used... so no *need* (but we still can...to be 100% correct and safe) to compile and validate SBOR length in other types (`TransactionIntent`, `SignedIntent`, `NotarizedIntent`, since they all use `Instructions`. 

What do you think?

apple/Tests/TestCases/RET/TransactionManifestTests.swift
micbakos-rdx: Unfortunatelly the android example app cannot go there. It has to be under the jvm directory with its own build.gradle file.

Regarding the manifests and dat files, the `testvectors` is something I remember @0xOmarA was following with kotlin engine toolkit when we were developing on that repo.

0xOmarA (commented):
Looks good to me! I've left some mostly minor comments and another on the compiled types.

src/wrapped_radix_engine_toolkit/low_level/transaction_manifest/instructions/instructions.rs
0xOmarA: My personal preference for this would be to keep the check in this way and to have `Instructions` fairly isolated from the other types and not make assumptions about where it will be used. Other types would have their own depth check that would eventually involve checking the depth of the instructions.

src/wrapped_radix_engine_toolkit/low_level/compiled_notarized_intent.rs
```diff
@@ -23,30 +23,36 @@ impl CompiledNotarizedIntent {
...
-        })
-        .and_then(TryInto::<NotarizedTransaction>::try_into)
+    pub fn bytes(&self) -> BagOfBytes {
+        self.secret_magic.clone()
     }
-}
 
-impl TryFrom<ScryptoNotarizedTransaction> for NotarizedTransaction {
-    type Error = crate::CommonError;
+    pub fn decompile(&self) -> NotarizedTransaction {
```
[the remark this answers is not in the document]
CyonAlexRDX: Right, so the thinking here is that the constructor of `CompiledNotarizedIntent` is hidden in Rust, Swift and Kotlin, so it is not possible to create an instance of it, thus, all instances are valid, since they have been created by calling `compile()` on `NotarizedTransaction` - where `NotarizedTransaction` itself is guaranteed to be valid.

0xOmarA:
Another comment on this, it seems like there is a common pattern for multiple types where we care about the SBOR depth and want to always ensure that they're below a certain depth. I think that it might be beneficial to do this as a wrapper type where we know that encoding this type is 100% safe and impossible to unwrap. I have an example of such a type here:

```rust
use scrypto::prelude::*;
use std::{borrow::Borrow, ops::Deref};

/// Represents an SBOR type whose depth has been validated. None of the interfaces allow for
/// mutating the inner value as any mutation to it will require another depth check.
#[derive(ManifestSbor, Clone, Debug)]
pub struct DepthValidated<T, const MAX_DEPTH: usize>(T);

impl<T, const MAX_DEPTH: usize> DepthValidated<T, MAX_DEPTH>
where
    T: ManifestEncode + ManifestDecode,
{
    fn new(value: T) -> Option<DepthValidated<T, MAX_DEPTH>> {
        // Attempt to encode with the specified depth.
        drop(manifest_encode_with_depth_limit(&value, MAX_DEPTH).ok()?);

        // Encoding succeeded, value is below the depth limit.
        Some(Self(value))
    }

    fn encode(&self) -> Vec<u8> {
        manifest_encode_with_depth_limit(&self.0, MAX_DEPTH).expect("Check was done in constructor")
    }

    fn decode(slice: &[u8]) -> Option<DepthValidated<T, MAX_DEPTH>> {
        manifest_decode_with_depth_limit(slice, MAX_DEPTH)
            .ok()
            .map(Self)
    }

    fn into_inner(self) -> T {
        self.0
    }
}

impl<T, const MAX_DEPTH: usize> AsRef<T> for DepthValidated<T, MAX_DEPTH> {
    fn as_ref(&self) -> &T {
        &self.0
    }
}

impl<T, const MAX_DEPTH: usize> Borrow<T> for DepthValidated<T, MAX_DEPTH> {
    fn borrow(&self) -> &T {
        &self.0
    }
}

impl<T, const MAX_DEPTH: usize> Deref for DepthValidated<T, MAX_DEPTH> {
    type Target = T;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

// Just an example.
type DepthValidatedInstructions = DepthValidated<Vec<InstructionV1>, MANIFEST_SBOR_V1_MAX_DEPTH>;
```

This type gives us some nice properties:

1. It's a standard way of marking that a type has had its SBOR depth checked. 
2. It supports any SBOR depth you may want and you're no longer limited to just the max when doing the encoding.
3. Encoding this type is guaranteed to be 100% safe as it has a small well-defined interface and its well understood that it's immutable.

I mention mutability quickly in one of the comments, but something like this can't be mutable since each mutation would require another encode check.

PR #55, opened 2024-04-02
Some more extensions + `@UsesSampleValues`

micbakos-rdx:
* New annotation named `@UsesSampleValues` which replaces the `@VisibleForTesting`. It is more versatile since if a sample is used by src code it will **fail during compile time** except if we specifically opt-in using them (which we have to do in composable previews). 
* Some more missing extensions and tests


CyonAlexRDX (approved):
LGTM, nice with custom annotations!

PR #64, opened 2024-04-05
Rename into() -> asSpecific and embed() -> asGeneral

CyonAlexRDX:
# Swift only
Rename method on *specific* Addresses which was name ~func embed()~ -> `var asGeneral` (a computed property)

Rename method ~`func into()`~ on `Address` -> `func asSpecific()`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1959,7 +1959,7 @@ dependencies = [
 
 [[package]]
 name = "sargon"
-version = "0.6.9"
+version = "0.6.10"
 dependencies = [
  "assert-json-diff",
  "bip32",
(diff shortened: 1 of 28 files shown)


apple/Sources/Sargon/Extensions/Swiftified/Prelude/NonFungibleGlobalID+Swiftified.swift
```diff
@@ -1,6 +1,12 @@
+import SargonUniFFI
+
 public typealias NonFungibleGlobalID = NonFungibleGlobalId
 
-extension NonFungibleGlobalID: IdentifiableByStringProtocol {}
+extension NonFungibleGlobalID: IdentifiableByStringProtocol {
+    public var localID: NonFungibleLocalID {
+        nonFungibleLocalId
```
CyonAlexRDX, line 7: @micbakos-rdx @kugel3 prefers using `localID` so I crated this "alias"

Cargo.lock
```diff
@@ -1959,7 +1959,7 @@ dependencies = [
 
 [[package]]
 name = "sargon"
-version = "0.6.9"
+version = "0.6.10"
```
CyonAlexRDX, line 1962: lets see now if my recent git tag change of Swift release script works :D previously git tags became `0.7.0` but I hope when Swift CD finishes once I merge tihs PR that the next release is indeed `0.6.10`...

PR #65, opened 2024-04-05
Kotlin batch

micbakos-rdx:
* Decimal192 from Double
* Component address extensions
* Android app in examples dir
* Create asGeneral() for addresses, signatures and public keys micbakos-rdx A minute ago


CyonAlexRDX (approved):
LGTM

PR #67, opened 2024-04-07
Gateway single crate

CyonAlexRDX:
Introduce `GatewayClient` and `NetworkAntenna`, with implementation in Rust, Swift and Kotlin.

This contains the bear minimum implementation of models - manually (not generated from schema since openapi generation with Rust is broken).

Using those models, the following endpoints have been implemented:
* (partial implementation) state entity details
* transaction/construct to get current epoch


jvm/sargon-android/src/test/java/com/radixdlt/sargon/NetworkAntennaTest.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    @Test
+    @Tag("IntegrationTests")
+    fun testNetworkFailure() = runBlocking {
+        val client = GatewayClient(SargonNetworkAntenna(okHttpClient), NetworkId.MAINNET)
+
+        val xrdBalance = client.xrdBalanceOfAccountOrZero(address = AccountAddress.sampleMainnet())
+
+        assertTrue(xrdBalance >= 1.toDecimal192())
+    }
```
micbakos-rdx, line 35: This is now a duplicate and can be deleted

micbakos-rdx (approved):
👏

PR #74, opened 2024-04-09
Pin Crates: Batch 2

CyonAlexRDX:
Continue "pinning" crates by usage of explicit commit hashes (`rev`) - soon all crates have been pinned.

These crates was removed all together:
- ed25519-dalek (the Ed25519PublicKey uses Ed25519 inside of IOTA Crypto crate)
- nutype
- schemars

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,9 @@ derive_more = { git = "https://github.com/JelteF/derive_more", rev = "1196b2dd7a
 serde = { version = "1.0.193", features = ["derive", "rc", "std"] }
 serde_json = { version = "1.0.108", features = ["preserve_order"] }
 serde_with = { version = "3.4.0" }
-serde_repr = "0.1.17"
+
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -22,49 +22,97 @@ path = "src/bindgen/bin.rs"
 required-features = ["build-binary"]
 
 [dependencies]
+
 log = "0.4.20"
-pretty_env_logger = "0.5.0"
-derive_more = { version = "1.0.0-beta.6", features = [
+# log = { git = "https://github.com/rust-lang/log/", rev = "4708f1484c7e6b8d4418b571c05e613b18e57673" }
```
GhenadieVP, line 27: commented? shouldn't it replace the `log` above or maybe you meant to comment out the one above :)?

Sajjon: Does not work.  Will need help from builders team

PR #76, opened 2024-04-09
Export serde methods

micbakos-rdx:


CyonAlexRDX (approved):
Nice! Looks great! Maybe add some unit test for unhappy path.

Maybe you can write "Profile" as type name, no need to read the name of the type since we know it is Profile?

micbakos-rdx:
@CyonAlexRDX Added some extensions too

src/profile/v100/profile.rs
```diff
@@ -390,6 +415,24 @@ mod tests {
...
 
+    #[test]
+    fn to_json_bytes_new_from_json_bytes() {
+        let sut = SUT::sample();
+
+        let encoded = sut.to_json_bytes().unwrap();
+        let profile_result = SUT::new_from_json_bytes(encoded).unwrap();
+        assert_eq!(profile_result, sut);
+
+        let malformed_profile_snapshot = BagOfBytes::from("{}".as_bytes());
```
Sajjon, line 426: can you split the happy path and unhappy path unit tests in two independent tests?

src/wrapped_radix_engine_toolkit/low_level/transaction_classes/detailed_manifest_class.rs
```diff
@@ -84,7 +84,7 @@ pub enum DetailedManifestClass {
         pool_addresses: Vec<PoolAddress>,
 
         /// The redemptions observed in the transaction
-        pool_contributions: Vec<TrackedPoolRedemption>,
+        pool_redemptions: Vec<TrackedPoolRedemption>,
```
Sajjon, line 87: thx for fixing my copy paste error!

PR #82, opened 2024-04-15
Create SupportedCurves and OtherGateways through macro

micbakos-rdx:
* Identified array of `[SLIP10Curve]` can now be referred to as `SLIP10Curves`
* `other: IdentifiedVecVia<Gateway>` is now called as `OtherGateways`. Could not refer to it as `Gateways` since Gateways is another struct with current inside. So I had to change the `decl_...` api a bit.
* The `impl` methods has a small renaming. For structs like `SLIP10Curve` the method that was creating the struct with just one curve would be named `with_s_l_i_p_10_curve()`. IMHO the name `just` for a method that instantiates an array with just one element inside was easier to remember. Also I just kept `from_iter` instead of redefining another method since it was doing the same thing.

### DIFF
--- a/src/core/error/common_error.rs
+++ b/src/core/error/common_error.rs
@@ -430,6 +430,9 @@ pub enum CommonError {
 
     #[error("Submitted transaction was duplicate.")]
     GatewaySubmitDuplicateTX { intent_hash: String } = 10119,
+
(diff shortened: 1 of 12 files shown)


CyonAlexRDX (approved):
Good job!

PR #88, opened 2024-04-17
Fix ios release properly

CyonAlexRDX:
I made iOS build artifacts unusable in last PR

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -10,9 +10,13 @@ comment:
 coverage:
   precision: 1
   range: 97...98 # red -> yellow (the inside range) -> green
+  status:
+    patch:
+      default:
+        target: auto
+        threshold: 80%
+        base: auto
+        only_pulls: true
 flag_management:
   default_rules:
     carryforward: true
-
-
-
--- a/scripts/ios/build-sargon.sh
+++ b/scripts/ios/build-sargon.sh
@@ -45,6 +45,18 @@ generate_ffi() {
     local TARGET_FOR_DYLIB_PATH="aarch64-apple-ios"
   fi 
   cargo run --features build-binary --bin sargon-bindgen generate --library target/$TARGET_FOR_DYLIB_PATH/release/lib$1.dylib --language swift --out-dir target/uniffi-xcframework-staging
+
+  if $release; then
+    echo "🚢 Ensuring Sargon build for release - that it will work for e.g. iOS wallet to archive."
+    sed -i '' 's/let useLocalFramework = false/let useLocalFramework = true/' Package.swift
+    swift build -c release || exit $?
+    echo "🚢 Swift Sargon builds for release ✅"
+
+    # Prepare for release
+    sed -i '' 's/let useLocalFramework = true/let useLocalFramework = false/' Package.swift
+  fi
+
+
   mkdir -p apple/Sources/UniFFI/
   mv target/uniffi-xcframework-staging/*.swift apple/Sources/UniFFI/
   mv target/uniffi-xcframework-staging/$1FFI.modulemap target/uniffi-xcframework-staging/module.modulemap  # Convention requires this have a specific name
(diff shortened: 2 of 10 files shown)


apple/Sources/UniFFI/Sargon.swift
Sajjon, line 1: revert

scripts/ios/release.sh
```diff
@@ -60,6 +49,9 @@ eval $GIT_COMMIT_CMD
 echo "🚢 🏷️ 📡 Pushing tag: $(NEXT_TAG), but only tag, not commit."
 `git push origin $NEXT_TAG`
 
+# if run locally, revert
+sed -i '' 's/let useLocalFramework = false/let useLocalFramework = true/' Package.swift
```
Sajjon, line 53: only needed if script is run locally (which we almost never do...)

scripts/ios/release.sh
```diff
@@ -35,17 +35,6 @@ XCFRAME_ZIP_PATH=`echo "$OUTPUT_OF_BUILD" | cut -d ";" -f 2` || exit $?
 echo "🚢  CHECKSUM: $CHECKSUM"
 echo "🚢  XCFRAME_ZIP_PATH: $XCFRAME_ZIP_PATH"
 
-echo "🚢 ensuring Swift Sargon build for release"
```
Sajjon, line 38: Moved into `release.sh` - my mistake here was that the `Packag.swift` was copied inside of the `release.sh` script, incorrectly containing `let useLocalFramework = true`.

scripts/ios/build-sargon.sh
```diff
@@ -45,6 +45,18 @@ generate_ffi() {
     local TARGET_FOR_DYLIB_PATH="aarch64-apple-ios"
   fi 
   cargo run --features build-binary --bin sargon-bindgen generate --library target/$TARGET_FOR_DYLIB_PATH/release/lib$1.dylib --language swift --out-dir target/uniffi-xcframework-staging
+
+  if $release; then
+    echo "🚢 Ensuring Sargon build for release - that it will work for e.g. iOS wallet to archive."
```
Sajjon, line 50: We must try the `swift build -c release` after FFI bindgen has run, and we try it before we create the XCFramework.

scripts/ios/build-sargon.sh
```diff
@@ -45,6 +45,18 @@ generate_ffi() {
...
+    sed -i '' 's/let useLocalFramework = false/let useLocalFramework = true/' Package.swift
+    swift build -c release || exit $?
```
Sajjon, line 52: must have `let useLocalFramework = true` when trying to do `swift build -c release` .... no new build artifacts on GH yet....

scripts/ios/build-sargon.sh
```diff
@@ -45,6 +45,18 @@ generate_ffi() {
...
+    echo "🚢 Swift Sargon builds for release ✅"
+
+    # Prepare for release
+    sed -i '' 's/let useLocalFramework = true/let useLocalFramework = false/' Package.swift
```
Sajjon, line 56: And we ensure to set `let useLocalFramework = false` in the Package.swift BEFORE we copy it over! ensuring it is part of the release.

codecov.yml
```diff
@@ -10,9 +10,13 @@ comment:
 coverage:
   precision: 1
   range: 97...98 # red -> yellow (the inside range) -> green
+  status:
+    patch:
+      default:
+        target: auto
+        threshold: 80%
```
Sajjon, line 17: the PATCH check is too aggressive, lowering to 80. The important threshold is not PATCH it is PROJECT. Previously if we added 50 new lines of code and one of was not unit tested, the whole PR check would fail, that is too aggressive IMO. With this change "only" 4/5 lines need tested for the PATCH to pass the PR chech.

micbakos-rdx (approved):

GhenadieVP (approved):
LGTM

PR #89, opened 2024-04-17
Fix ios cd again

CyonAlexRDX:


scripts/ios/build-sargon.sh
```diff
@@ -45,18 +45,6 @@ generate_ffi() {
     local TARGET_FOR_DYLIB_PATH="aarch64-apple-ios"
   fi 
   cargo run --features build-binary --bin sargon-bindgen generate --library target/$TARGET_FOR_DYLIB_PATH/release/lib$1.dylib --language swift --out-dir target/uniffi-xcframework-staging
-
-  if $release; then
-    echo "🚢 Ensuring Sargon build for release - that it will work for e.g. iOS wallet to archive."
-    sed -i '' 's/let useLocalFramework = false/let useLocalFramework = true/' Package.swift
-    swift build -c release || exit $?
```
Sajjon, line 52: bah. ofc this did not work, I've only built Rust project but not Xcframeworks yet. In fact, safest to not modify the `build` script at all, so I've reverted back to what it was a couple of days ago...

scripts/ios/release.sh
```diff
@@ -12,6 +12,9 @@ cd "../../" # go to parent of parent, which is project root.
 echo "🚢 Start of '$me' (see: '$DIR/$me')"
 echo "🚢 PWD: $PWD"
 
+echo "🚢 Switch 'useLocalFramework' to 'false' in Package.swift for release"
+sed -i '' 's/let useLocalFramework = true/let useLocalFramework = false/' Package.swift
```
Sajjon, line 16: reverted back to what we had a couple of days ago...

PR #91, opened 2024-04-18
gateway and gateways 

CyonAlexRDX:
UF export:
- Index of Bip44 path 
- lookup GW from NetworkId
- new GW from URL and NetworkId
- Well known GWs
- GW get id 
- GW toString 
- GWS (Gateways) default / preset 
- GWS get all elements 
- GWS change current GW


apple/Sources/Sargon/Extensions/Methods/Prelude/Decimal192+Wrap+Functions.swift
```diff
@@ -123,19 +123,11 @@ extension Decimal192 {
 extension Decimal192 {
 
 	private func rounded(decimalPlaces: UInt8, roundingMode: RoundingMode) -> Self {
-		precondition(
-			decimalPlaces <= Decimal192.maxDivisibility,
-			"Decimal places MUST be 0...18, was: \(decimalPlaces)"
```
micbakos-rdx, line 128: Should I remove this from Kotlin too?

micbakos-rdx (approved):

apple/Sources/Sargon/Extensions/Methods/Prelude/Decimal192+Wrap+Functions.swift
Sajjon: I just removed it because it lowered code coverage - we panic in Rust anyway, so not needed to have it there.

Sajjon: but we still should improve rounding methods a bit :) can do it later. create some throwing and some non throwing (using mode `towardsZero` never fails). And perhaps even make number of places an enum?

PR #101, opened 2024-04-23
Batch 10: BIP39Entropy UF exp and more

CyonAlexRDX:
# Rust
- Exactly N Bytes is now `json_string_convertible!` (Swift uses it to make `Exactly32Bytes` as `Swift.Codable` - allowing us to replace iOS wallet type Hex32BytesCodable with it)
- `secret_bytes!` macro now UF exports the type => `BIP39Entropy` is now UF exported
- `BIP39Word` UF exported samples (samples are new)
- `Mnemonic` new UF exported construtor from `entropy: BIP39Entropy, language`, used to generate new mnemonic using FFI side provided entoropy of fixed BIP39 lengths.
- `AppearanceID` FIX BUG where `11` could not be used as ID
- `AppearanceID` UF export `from_number_of_accounts_on_network` constructor
- `FactorSourceCryptoParameter` UF export `preset_babylon_only`, `preset_olympia_only` and `preset_babylon_olympia_compatibe` and `_supports_olympia` and `_supports_babylon` functions + samples
- `FactorSourceIDFromHash` UF export `from_mnemonic_with_passphrase` constructor

# Swift
Swiftified all of the above.

- Added `FactorSourceIDProtocol` (migrated from iOS app)
- Renamed `FactorSourceProtocol` -> `BaseFactorSourceProtocol` and added (Specific)`FactorSourceProtocol` with missing functionalitly for `embed`/`extract` from iOS wallet


src/core/types/secret_bytes.rs
```diff
@@ -11,26 +11,74 @@ macro_rules! decl_secret_bytes {
...
             $(
                 #[doc = $expr]
             )*
-            #[derive(Zeroize, ZeroizeOnDrop, derive_more::Debug, derive_more::Display)]
+            #[derive(Zeroize, derive_more::Debug, derive_more::Display, uniffi::Record)]
             #[debug("OBFUSCATED")]
             #[display("OBFUSCATED")]
-            pub struct $struct_name(Box<[u8; Self::LENGTH]>);
+            pub struct $struct_name {
+                secret_magic: [< $struct_name SecretMagic >]
```
Sajjon, line 48: using the `secret_magic` trick as usual, making the `$struct_name` its own concrete type, rather than a typealias, on the FFI side.

src/hierarchical_deterministic/bip39/bip39_entropy.rs
```diff
@@ -21,7 +21,7 @@ macro_rules! entropy_with_byte_counts {
             $(
                 #[doc = $expr]
             )*
-            #[derive(Zeroize, ZeroizeOnDrop)]
+            #[derive(Zeroize, uniffi::Enum)]
```
Sajjon, line 24: Have to skip `ZeroizeOnDrop` due to https://github.com/mozilla/uniffi-rs/issues/2080

I will give this some more thought, we might wanna ensure we pass secrets by mutable reference, so that we can call `.zeroize()` on them.

src/profile/v100/entity/account/appearance_id.rs
```diff
@@ -32,9 +32,7 @@ impl AppearanceID {
     }
 
     pub fn from_number_of_accounts_on_network(n: usize) -> Self {
-        Self {
-            value: (n % (Self::MAX as usize)) as u8,
-        }
+        Self::new((n % ((Self::MAX + 1) as usize)) as u8).unwrap()
```
Sajjon, line 35: `+1` needed, to fix the bug! otherwise `Self::MAX % Self::MAX == 0`, but we want it to be  `Self::MAX % Self::MAX == Self::MAX`.

src/profile/v100/entity/account/appearance_id.rs
```diff
@@ -110,59 +108,68 @@ impl From<AppearanceID> for u8 {
 #[cfg(test)]
 mod tests {
 
-    use crate::prelude::*;
+    use super::*;
+
+    #[allow(clippy::upper_case_acronyms)]
+    type SUT = AppearanceID;
```
Sajjon, line 114: @jakub-rdx @micbakos-rdx @GhenadieVP When I started in November I did not use the `SUT` pattern, but I've been moving over to it gradually. So if you ever see a `mod tests` which does NOT use `type SUT =`, please help convert to it.  `SUT` is great since it makes writing macros much easier to!

Also I did not know I could use `use super::*;`, so if you see ` use crate::prelude::*;` in mod tests, please change to `use super::*;`

src/core/types/exactly_n_bytes.rs
```diff
@@ -103,6 +103,9 @@ macro_rules! decl_exactly_n_bytes {
                 secret_magic: [<Exactly $byte_count Bytes SecretMagic>],
             }
 
+            // Make it JSON String convertible in Swift/Kotlin
+            json_string_convertible!([<Exactly $byte_count Bytes>]);
```
Sajjon, line 107: A macro calling a macro calling a macro 🤯 

As @micbakos-rdx said, compared to Rust other languages are toy languages 🤩

micbakos-rdx (approved):

PR #102, opened 2024-04-24
[ABW-3207] app link parsing

jakub-rdx:
- add parsing of app link and translate it into Radix Connect request

### DIFF
--- a/src/core/error/common_error.rs
+++ b/src/core/error/common_error.rs
@@ -459,7 +459,10 @@ pub enum CommonError {
     RadixConnectMobileInvalidOrigin { bad_value: String } = 10128,
 
     #[error("Failed to create Session (UUID) from string: {bad_value}")]
-    RadixConnectInvalidSessionID { bad_value: String } = 10129,
+    RadixConnectMobileInvalidSessionID { bad_value: String } = 10129,
+
+    #[error("Failed to create InteractionID (UUID) from string: {bad_value}")]
+    RadixMobileInvalidInteractionID { bad_value: String } = 10130,
 }
 
 #[uniffi::export]
--- a/src/core/error/common_error.rs
+++ b/src/core/error/common_error.rs
@@ -452,11 +452,11 @@ pub enum CommonError {
     #[error("Invalid LedgerHardwareWalletModel, bad value: {bad_value}")]
     InvalidLedgerHardwareWalletModel { bad_value: String } = 10126,
 
-    #[error("RadixConnectMobile invalid URL")]
-    RadixConnectMobileInvalidRequestUrl = 10127,
+    #[error("RadixConnectMobile invalid URL, bad value: {bad_value}")]
+    RadixConnectMobileInvalidRequestUrl { bad_value: String } = 10127,
 
-    #[error("RadixConnectMobile invalid URL")]
-    RadixConnectMobileInvalidOrigin = 10128,
+    #[error("RadixConnectMobile invalid origin, bad value: {bad_value}")]
+    RadixConnectMobileInvalidOrigin { bad_value: String } = 10128,
 }
 
 #[uniffi::export]
--- a/src/core/error/common_error.rs
+++ b/src/core/error/common_error.rs
@@ -451,6 +451,12 @@ pub enum CommonError {
 
     #[error("Invalid LedgerHardwareWalletModel, bad value: {bad_value}")]
     InvalidLedgerHardwareWalletModel { bad_value: String } = 10126,
+
+    #[error("RadixConnectMobile invalid URL")]
+    RadixConnectMobileInvalidRequestUrl = 10127,
+
+    #[error("RadixConnectMobile invalid URL")]
+    RadixConnectMobileInvalidOrigin = 10128,
 }
 
 #[uniffi::export]
--- a/src/core/error/common_error.rs
+++ b/src/core/error/common_error.rs
@@ -457,6 +457,9 @@ pub enum CommonError {
 
     #[error("RadixConnectMobile invalid origin, bad value: {bad_value}")]
     RadixConnectMobileInvalidOrigin { bad_value: String } = 10128,
+
+    #[error("Failed to create Session (UUID) from string: {bad_value}")]
+    RadixConnectInvalidSessionID { bad_value: String } = 10129,
 }
 
 #[uniffi::export]
--- a/src/radix_connect/app_link_parser.rs
+++ b/src/radix_connect/app_link_parser.rs
@@ -2,82 +2,72 @@ use crate::prelude::*;
 use url::form_urlencoded;
 use url::Url;
 
-use super::{DappRequest, LinkRequest, MobileConnectRequest};
+const CONNECT_URL_PARAM_SESSION_ID: &str = "sessionId";
+const CONNECT_URL_PARAM_ORIGIN: &str = "origin";
+const CONNECT_URL_PARAM_INTERACTION_ID: &str = "interactionId";
+const CONNECT_URL: &str = "https://d1rxdfxrfmemlj.cloudfront.net";
 
-#[uniffi::export]
-pub fn parse_request_url(url: String) -> Result<MobileConnectRequest> {
+pub fn parse_mobile_connect_request(
+    url: String,
+) -> Result<MobileConnectRequest> {
     let connect_url = Url::parse(CONNECT_URL).unwrap();
-    match Url::parse(url.as_str()) {
-        Ok(parsed_url) => {
-            if parsed_url.host_str() != connect_url.host_str()
-                || parsed_url.scheme() != connect_url.scheme()
-            {
-                return Err(CommonError::RadixConnectMobileInvalidRequestUrl);
-            }
-
-            let query_parameters = parsed_url
-                .query_pairs()
-                .into_owned()
-                .collect::<HashMap<String, String>>();
-            if let Some(session_id) =
-                query_parameters.get(CONNECT_URL_PARAM_SESSION_ID)
-            {
-                match query_parameters.get(CONNECT_URL_PARAM_ORIGIN) {
-                    Some(origin) => {
-                        if let Ok(url) = Url::parse(origin) {
-                            return Ok(MobileConnectRequest::Link(
-                                LinkRequest {
-                                    origin: url,
-                                    session_id: SessionID(
-                                        session_id.to_owned(),
-                                    ),
-                                },
-                            ));
-                        }
-                        Err(CommonError::RadixConnectMobileInvalidOrigin)
-                    }
-                    None => {
-                        match query_parameters
-                            .get(CONNECT_URL_PARAM_INTERACTION_ID)
-                        {
-                            Some(interaction_id) => {
-                                Ok(
-                                    MobileConnectRequest::DappInteraction(
-                                        DappRequest {
-                                            session_id: SessionID(
-                                                session_id.to_owned(),
-                                            ),
-                                            interaction_id: WalletInteractionId(
-                                                interaction_id.to_owned(),
-                                            ),
-                                        },
-                                    ),
-                                )
-                            }
-                            None => {
-                                Err(CommonError::RadixConnectMobileInvalidRequestUrl)
-                            }
-                        }
-                    }
-                }
-            } else {
-                Err(CommonError::RadixConnectMobileInvalidRequestUrl)
-            }
+    let parsed_url = Url::parse(url.as_str()).map_err(|_| {
+        CommonError::RadixConnectMobileInvalidRequestUrl {
+            bad_value: url.clone(),
         }
-        Err(_) => Err(CommonError::RadixConnectMobileInvalidRequestUrl),
+    })?;
+    if parsed_url.host_str() != connect_url.host_str()
+        || parsed_url.scheme() != connect_url.scheme()
+    {
+        return Err(CommonError::RadixConnectMobileInvalidRequestUrl {
+            bad_value: url.clone(),
+        });
+    }
+
+    let query_parameters = parsed_url
+        .query_pairs()
+        .into_owned()
+        .collect::<HashMap<String, String>>();
+    let session_id = query_parameters.get(CONNECT_URL_PARAM_SESSION_ID).ok_or(
+        CommonError::RadixConnectMobileInvalidRequestUrl {
+            bad_value: url.clone(),
+        },
+    )?;
+
+    match query_parameters.get(CONNECT_URL_PARAM_ORIGIN) {
+        Some(origin) => Url::parse(origin)
+            .map_err(|_| CommonError::RadixConnectMobileInvalidOrigin {
+                bad_value: origin.to_owned(),
+            })
+            .map(|url| {
+                MobileConnectRequest::Link(LinkRequest {
+                    origin: url,
+                    session_id: SessionID(session_id.to_owned()),
+                })
+            }),
+        None => {
+            let interaction_id = query_parameters
+                .get(CONNECT_URL_PARAM_INTERACTION_ID)
+                .ok_or(CommonError::RadixConnectMobileInvalidRequestUrl {
+                    bad_value: url.clone(),
+                })?;
+            Ok(MobileConnectRequest::DappInteraction(DappRequest {
+                session_id: SessionID(session_id.to_owned()),
+                interaction_id: WalletInteractionId(interaction_id.to_owned()),
+            }))
+        }
     }
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::prelude::*;
 
     #[test]
     fn parse_url_into_link_request_origin() {
         let connect_url =
             CONNECT_URL.to_owned() + "/?sessionId=123&origin=radix%3A%2F%2Fapp";
-        let result = parse_request_url(connect_url);
+        let result = parse_mobile_connect_request(connect_url);
         assert!(result.is_ok());
         match result.unwrap() {
             MobileConnectRequest::Link(link_request) => {
@@ -99,7 +89,7 @@ mod tests {
     #[test]
     fn parse_url_into_dapp_interaction() {
         let url = CONNECT_URL.to_owned() + "/?sessionId=123&interactionId=456";
-        let result = parse_request_url(url);
+        let result = parse_mobile_connect_request(url);
         assert!(result.is_ok());
         match result.unwrap() {
             MobileConnectRequest::DappInteraction(dapp_request) => {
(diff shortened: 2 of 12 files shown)


src/core/error/common_error.rs
```diff
@@ -451,6 +451,12 @@ pub enum CommonError {
 
     #[error("Invalid LedgerHardwareWalletModel, bad value: {bad_value}")]
     InvalidLedgerHardwareWalletModel { bad_value: String } = 10126,
+
+    #[error("RadixConnectMobile invalid URL")]
+    RadixConnectMobileInvalidRequestUrl = 10127,
+
```
Sajjon, line 457: copy paste typo in description, and also you might wanna include a `{ bad_value: String }`, see `InvalidLedgerHardwareWalletModel` above

src/radix_connect/app_link_parser.rs
```diff
@@ -0,0 +1,154 @@
+use crate::prelude::*;
+use url::form_urlencoded;
+use url::Url;
+
+use super::{DappRequest, LinkRequest, MobileConnectRequest};
```
GhenadieVP, line 5: this potentially should not be needed, as `use crate::prelude::*` should expose these

src/radix_connect/app_link_parser.rs
```diff
@@ -0,0 +1,154 @@
...
+
+#[uniffi::export]
+pub fn parse_request_url(url: String) -> Result<MobileConnectRequest> {
+    let connect_url = Url::parse(CONNECT_URL).unwrap();
+    match Url::parse(url.as_str()) {
+        Ok(parsed_url) => {
+            if parsed_url.host_str() != connect_url.host_str()
```
Sajjon, line 12: Better to avoid "pyramid of doom", by returning early, like so

```rust
   let parsed_url = Url::try_from(&url)
            .map_err(|_| CommonError::RadixConnectMobileInvalidRequestUrl{ bad_value: url })?;
```

Maybe best if you create a pub global utility function for that, and use it here and in `Gateway`. **Notice the `?` in the end, it will return from the whole function if err**, lets for the sake of this review call the function `parse_url`, so:

Somewhere, probably in `src/core/utils` a new file perhaps...? or in `string_utils`:

```rust
pub fn parse_url(s: impl AsRef<str>) -> Result<Url> {
     Url::try_from(url.as_str())
            .map_err(|_| CommonError::InvalidURL { bad_value: url })
}
```

and then here:

```rust
let parsed_url = parse_url(url).map_err(|_| RadixConnectMobileInvalidRequestUrl { bad_value: url })?;
...
```

src/radix_connect/app_link_parser.rs
```diff
@@ -0,0 +1,154 @@
...
+                return Err(CommonError::RadixConnectMobileInvalidRequestUrl);
+            }
+
+            let query_parameters = parsed_url
+                .query_pairs()
+                .into_owned()
+                .collect::<HashMap<String, String>>();
+            if let Some(session_id) =
+                query_parameters.get(CONNECT_URL_PARAM_SESSION_ID)
+            {
```
Sajjon, line 24: same here, avoid indentation and return early. *We ALWAYS wanna return early, make all code easier to read :)*, like so:

```rust
let session_id = query_parameters.get(CONNECT_URL_PARAM_SESSION_ID). ok_or(CommonError::RadixConnectMobileInvalidRequestUrl)?;
```

src/radix_connect/app_link_parser.rs
```diff
@@ -0,0 +1,154 @@
+use crate::prelude::*;
+use url::form_urlencoded;
+use url::Url;
+
+use super::{DappRequest, LinkRequest, MobileConnectRequest};
+
+#[uniffi::export]
+pub fn parse_request_url(url: String) -> Result<MobileConnectRequest> {
+    let connect_url = Url::parse(CONNECT_URL).unwrap();
+    match Url::parse(url.as_str()) {
```
GhenadieVP, line 10: to reduce nesting I think you can this:

```rust
let parsed_url = Url::parse(url.as_str())
        .map_err(|_| CommonError::RadixConnectMobileInvalidRequestUrl)?;
```
this will throw immediately if the error is encountered. with this you can remove the whole `match`

src/radix_connect/app_link_parser.rs
```diff
@@ -0,0 +1,154 @@
...
+                || parsed_url.scheme() != connect_url.scheme()
+            {
+                return Err(CommonError::RadixConnectMobileInvalidRequestUrl);
+            }
+
+            let query_parameters = parsed_url
+                .query_pairs()
+                .into_owned()
+                .collect::<HashMap<String, String>>();
+            if let Some(session_id) =
```
GhenadieVP, line 22: same here, something like this would remove nesting
```
 let session_id = query_parameters.get(CONNECT_URL_PARAM_SESSION_ID)
        .ok_or(CommonError::RadixConnectMobileInvalidRequestUrl)?
```

src/radix_connect/app_link_parser.rs
```diff
@@ -0,0 +1,154 @@
...
+                query_parameters.get(CONNECT_URL_PARAM_SESSION_ID)
+            {
+                match query_parameters.get(CONNECT_URL_PARAM_ORIGIN) {
+                    Some(origin) => {
+                        if let Ok(url) = Url::parse(origin) {
```
GhenadieVP, line 27: you could probably do something like:
```rust
Url::parse(origin)
   .map_error(|_| CommonError::RadixConnectMobileInvalidOrigin)
   .and_then(|url| {
      Ok(MobileConnectRequest::Link(
                                LinkRequest {
                                    origin: url,
                                    session_id: SessionID(
                                        session_id.to_owned(),
                                    ),
                                },
                            ));
   })
```
I don't recall the exact syntax, but something along these lines :)

src/radix_connect/app_link_parser.rs
```diff
@@ -0,0 +1,154 @@
...
+                                    session_id: SessionID(
+                                        session_id.to_owned(),
+                                    ),
+                                },
+                            ));
+                        }
+                        Err(CommonError::RadixConnectMobileInvalidOrigin)
+                    }
+                    None => {
+                        match query_parameters
```
GhenadieVP, line 40: you could use `ok_or` here as well to remove nesting.

src/radix_connect/app_link_parser.rs
Sajjon, line 27: use `parse_url` and return early...:

```rust
let url = parse_url(origin).map_err(|_| CommonError::RadixConnectMobileInvalidOrigin { bad_value: origin })?;
```

src/radix_connect/app_link_parser.rs
```diff
@@ -0,0 +1,154 @@
...
+                            .get(CONNECT_URL_PARAM_INTERACTION_ID)
```
Sajjon, line 41: return early:

```rust
let interaction_id  = query_parameters
    .get(CONNECT_URL_PARAM_INTERACTION_ID)
    .ok_or(CommonError::RadixConnectMobileInvalidRequestUrl)?;
```

src/radix_connect/mobile_connect_request.rs
```diff
@@ -0,0 +1,19 @@
+use crate::prelude::*;
+
+#[derive(Debug, uniffi::Enum)]
+pub enum MobileConnectRequest {
```
GhenadieVP, line 4: Sample values for these models would be nice to have

src/radix_connect/app_link_parser.rs
```diff
@@ -0,0 +1,154 @@
...
+            String::from("https://d1rxdfxrfmemlj.cloudfront.net/?invalid=1");
+        let err = parse_request_url(url).err().unwrap();
+        assert!(matches!(
+            err,
+            CommonError::RadixConnectMobileInvalidRequestUrl
+        ));
+    }
+}
+
+const CONNECT_URL_PARAM_SESSION_ID: &str = "sessionId";
```
GhenadieVP, line 151: I think these need to be moved up top.

GhenadieVP (commented):
Amazing start 👍, left some comments about some code improvements.

CyonAlexRDX (changes requested):
Good start!

src/radix_connect/app_link_parser.rs
Sajjon: ha I should have use "Add single comment", my suggestion is better than this one :) even more generalized :)

src/radix_connect/app_link_parser.rs
Sajjon: :D we wrote the same thing :D

GhenadieVP: at the same time :D

src/radix_connect/app_link_parser.rs
Sajjon: you can use `map` instead of `and_then`, removing need for `Ok`.

@jakub-rdx also call `cargo clippy --fix --allow-staged`, it can fix some stuff automatically...

src/core/error/common_error.rs
Sajjon: ofc making use of `bad_value` in description as seen in:

```rust
  #[error("Invalid LedgerHardwareWalletModel, bad value: {bad_value}")]
```

src/radix_connect/app_link_parser.rs
Sajjon: and of course write unit tests for `parse_url` function in that file

src/radix_connect/mobile_connect_request.rs
```diff
@@ -0,0 +1,19 @@
...
+    Link(LinkRequest),
+    DappInteraction(DappRequest),
+}
+
+#[derive(Debug, uniffi::Record)]
+pub struct LinkRequest {
```
Sajjon, line 10: move `LinkRequest` out to separate file

src/radix_connect/mobile_connect_request.rs
```diff
@@ -0,0 +1,19 @@
...
+    pub origin: Url,
+    pub session_id: SessionID,
+}
+
+#[derive(Debug, uniffi::Record)]
+pub struct DappRequest {
```
Sajjon, line 16: move `DappRequest` to separate file

src/radix_connect/session_id.rs
```diff
@@ -0,0 +1,43 @@
+use crate::prelude::*;
+
+uniffi::custom_newtype!(SessionID, String);
+
+#[derive(
+    Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Ord, PartialOrd, Hash,
+)]
+pub struct SessionID(pub String);
```
Sajjon, line 8: hmm what is the SessionID though? Is it a string? is it bytes, or is it a `UUID`?

src/radix_connect/mobile_connect_request.rs
Sajjon: Yes! ALWAYS SampleValues for ALL types :) and then test:

(here, example: `SargonBuildInformation`)
```rust

#[cfg(test)]
mod tests {
    use super::*;

    type SUT = SargonBuildInformation;

    #[test]
    fn equality() {
        assert_eq!(SUT::sample(), SUT::sample());
        assert_eq!(SUT::sample_other(), SUT::sample_other());
    }

    #[test]
    fn inequality() {
    }
}

```

src/radix_connect/app_link_parser.rs
```diff
@@ -0,0 +1,154 @@
+use crate::prelude::*;
+use url::form_urlencoded;
+use url::Url;
+
+use super::{DappRequest, LinkRequest, MobileConnectRequest};
+
+#[uniffi::export]
+pub fn parse_request_url(url: String) -> Result<MobileConnectRequest> {
```
Sajjon, line 8: oh also, you violate an important principle here, our UniFFI exported method should NEVER have logic in them, they should ALWAYS just contain one single trivial line calling another, non UniFFI exported function.

So change to let `MobileConnectRequest` impl `FromStr` and use it:

(inside the 
```rust
impl FromStr for MobileConnectRequest {
    type Err = CommonError;
    fn from_str(s: &str) -> Result<Self> { /* this function below... */}
}
```

and then this UniFFI exported function should be just:

```rust
#[uniffi::export]
pub fn new_mobile_connect_request(url: String) -> Result<MobileConnectRequest> {
    MobileConnectRequest::from_str(&url)
}

#[cfg(test)]
mod uniffi_test { 
    use super::*;

    type SUT = MobileConnectRequest;
   #[test]
   fn test_new_mobile_connect_request() {
        assert!(new_mobile_connect_request("some example").is_ok()) // at least this test
   }
}
```

src/radix_connect/app_link_parser.rs
```diff
@@ -0,0 +1,154 @@
...
+            }
+        }
+        Err(_) => Err(CommonError::RadixConnectMobileInvalidRequestUrl),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::prelude::*;
```
Sajjon, line 74: no need for: `use crate::prelude::*;` it is included by ` use super::*;`

src/radix_connect/mobile_connect_request.rs
Sajjon: ofc with `HasSampleValues` and with unit tests, like:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    type SUT = SargonBuildInformation;

    #[test]
    fn equality() {
        assert_eq!(SUT::sample(), SUT::sample());
        assert_eq!(SUT::sample_other(), SUT::sample_other());
    }

    #[test]
    fn inequality() {
    }
}

```

src/radix_connect/mobile_connect_request.rs
Sajjon: ofc with `HasSampleValues` and with unit tests, like:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    type SUT = SargonBuildInformation;

    #[test]
    fn equality() {
        assert_eq!(SUT::sample(), SUT::sample());
        assert_eq!(SUT::sample_other(), SUT::sample_other());
    }

    #[test]
    fn inequality() {
    }
}

```

src/radix_connect/session_id.rs
Sajjon: If it is a UUID, take a look at: `profile_id.rs` ! and use same pattern as that perhaps?

src/radix_connect/app_link_parser.rs
jakub-rdx: I return early now but was not able to add `parse_url` so far

src/radix_connect/session_id.rs
jakub-rdx: it's uuid, I will update it

src/radix_connect/app_link_parser.rs
```diff
@@ -0,0 +1,154 @@
...
+use url::form_urlencoded;
+use url::Url;
+
+const CONNECT_URL_PARAM_SESSION_ID: &str = "sessionId";
+const CONNECT_URL_PARAM_ORIGIN: &str = "origin";
+const CONNECT_URL_PARAM_INTERACTION_ID: &str = "interactionId";
+const CONNECT_URL: &str = "https://d1rxdfxrfmemlj.cloudfront.net";
+
+pub fn parse_mobile_connect_request(
+    url: String,
```
Sajjon, line 11: Better if `parse_mobile_connect_request` takes an `impl AsRef<str>`, that way you can:

```rust
pub fn parse_mobile_connect_request(
    url: impl AsRef<Str>, // 🔮
) -> Result<MobileConnectRequest> {
    let url = url.as_ref(); // 🔮
    let connect_url = Url::parse(CONNECT_URL).unwrap();
    let parsed_url = Url::parse(url).map_err(|_| { // 🔮 no need to use `.as_str()` (you should have used there btw `&` 😊)
        CommonError::RadixConnectMobileInvalidRequestUrl {
            bad_value: url.to_owned(), // 🔮 have to make it owned (&str -> String conversion)
        }
    })?;
```

This have the advantage that we can now parse *either* a `String` or ` &str`.

And down below, in your tests, instead of:

```rust
 let connect_url = CONNECT_URL.to_owned() + "/?sessionId=123&origin=radix%3A%2F%2Fapp";
```

do this:

```rust
let connect_url = concat!(CONNECT_URL, "/?sessionId=123&origin=radix%3A%2F%2Fapp");
```

and the big win is the rest of the tests:

```rust
    #[test]
    fn url_does_not_match_expected() {
        let url = String::from("https://example.com");
        let err = parse_mobile_connect_request(url.clone()).err().unwrap();
        assert_eq!(
            err,
            CommonError::RadixConnectMobileInvalidRequestUrl { bad_value: url }
        );
    }
```

this become this:

```rust
    #[test]
    fn url_does_not_match_expected() {
        let url = "https://example.com";
        let err = parse_mobile_connect_request(url).err().unwrap(); // 🔮 No need to clone!!!
        assert_eq!(
            err,
            CommonError::RadixConnectMobileInvalidRequestUrl { bad_value: url.to_owned() } //  🔮 must call `to_owned()`
        );
    }
```

src/radix_connect/app_link_parser.rs
```diff
@@ -0,0 +1,173 @@
...
+        .ok_or(CommonError::RadixConnectMobileInvalidRequestUrl {
+            bad_value: url.clone(),
+        })?;
+    let session_id = SessionID::from_str(session_id_string).map_err(|_| {
+        CommonError::RadixConnectMobileInvalidRequestUrl {
+            bad_value: url.clone(),
+        }
+    })?;
+
+    match query_parameters.get(CONNECT_URL_PARAM_ORIGIN) {
```
Sajjon, line 42: This last match here, can be reworked to use `let else` which makes a more readable, and this solution also scales better, moving heavy lifting done here *ad-hoc* into reusable code:

something like: 

```rust
{ ...

let Some(origin) = query_parameters.get(CONNECT_URL_PARAM_ORIGIN) else {
    query_parameters
        .get(CONNECT_URL_PARAM_INTERACTION_ID)
        .ok_or(CommonError::RadixConnectMobileInvalidRequestUrl {
            bad_value: url.clone(),
        })
        .and_then(|interaction_id| {
            DappRequest::try_with_interaction_id_and_session_id(
                interaction_id,
                session_id
            )
        })
        .map(MobileConnectRequest::DappInteraction)
};

LinkRequest::try_with_origin_and_session_id(origin, session_id)
        .map(MobileConnectRequest::Link)

}
```

Inside `link_request.rs` file

```rust
impl LinkRequest {
    pub(crate) fn try_with_origin_and_session_id(
        origin: impl AsRef<str>,
        session_id: SessionD
    ) -> Result<Self> {
        let origin = Url::parse(origin) // 🔮 you MIGHT have to call `.as_ref()` 
            .map_err(|_| CommonError::RadixConnectMobileInvalidOrigin {
                bad_value: origin.to_owned(),
            })?;

        OK(Self::new(origin, session_id)) // 🔮 write a constructor `new`
    }
}
```

and inside `dapp_request.rs` file

```rust
impl DappRequest {
    pub(crate) fn try_with_interaction_id_and_session_id(
        interaction_id: impl AsRef<str>,
        session_id: SessionID
    ) -> Result<Self> {

        // 🔮 WalletInteractionId is probably a Uuid...
        // https://github.com/radixdlt/sargon/pull/98/files#r1578324966
        // so must TRY parse it with `FromStr` (returns Result)
        let interaction_id = WalletInteractionId.from_str(interaction_id)?;
        Ok(Self::new(interaction_id, session_id)) // 🔮 write a constructor `new`
    }
}
```

src/radix_connect/app_link_parser.rs
Sajjon: what do you mean by "was not able to add `parse_url`" ? :) did you make it `pub fn`? if you add a `pub fn` into e.g. `string_utils.rs` and as long as you always in the top of each file do `use crate::prelude::*;` it should be visible :) 

the advantage of that small function is that it returns `Result<Url>` which just is an alias for `Result<Url, CommonError>` - which has the advatange of us being able in other functions returning `Result<T>` we can use just `parse_url(url)?;` to return early, **no need** to map the error type, since it is our own `CommonError` :) 

Ofc in some cases we *might* wanna return a specific error, relating to radix mobile connect. but maybe not?

src/radix_connect/app_link_parser.rs
Sajjon: @jakub-rdx sorry, I meant `as_ref()` ofc :)

jakub-rdx: declared function as 
`pub fn parse_url(s: impl AsRef<str>) -> Result<Url, CommonError> {
    Url::try_from(s.as_ref()).map_err(|_| CommonError::InvalidURL {
        bad_value: s.as_ref().to_owned(),
    })
}` and using it like 
`let parsed_url = parse_url(url).map_err(|_| RadixConnectMobileInvalidRequestUrl { bad_value: url })?;`
produces compiler error

src/radix_connect/dapp_request.rs
```diff
@@ -0,0 +1,125 @@
...
+                    bad_value: session_id.as_ref().to_owned(),
+                }
+            })?;
+        Ok(DappRequest::new(interaction_id, session_id))
+    }
+}
+
+impl HasSampleValues for DappRequest {
+    fn sample() -> Self {
+        Self {
```
CyonAlexRDX, line 42: You can use the constructor for sample values

src/radix_connect/link_request.rs
```diff
@@ -0,0 +1,77 @@
...
+                CommonError::RadixConnectMobileInvalidSessionID {
+                    bad_value: session_id.as_ref().to_owned(),
+                }
+            })?;
+        Ok(LinkRequest::new(origin, session_id))
+    }
+}
+
+impl HasSampleValues for LinkRequest {
+    fn sample() -> Self {
```
CyonAlexRDX, line 36: Use constructor :)

GhenadieVP (approved):
LGTM 💪!

PR #104, opened 2024-04-25
Add AddressFormat.Middle support

matiasbzurovski:
## Changelog
The wallet apps will need to colorise the part of an address that is normally hidden. The new `AddressFormat::Middle` will return such part of a given address.

## Related issues
[ABW-3212](https://radixdlt.atlassian.net/browse/ABW-3212)

[ABW-3212]: https://radixdlt.atlassian.net/browse/ABW-3212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/profile/v100/address/address_format.rs
+++ b/src/profile/v100/address/address_format.rs
@@ -14,4 +14,5 @@ pub enum AddressFormat {
     Full,
     Raw,
     Default,
+    Hidden,
 }
--- a/src/profile/v100/address/address_format.rs
+++ b/src/profile/v100/address/address_format.rs
@@ -14,5 +14,5 @@ pub enum AddressFormat {
     Full,
     Raw,
     Default,
-    Hidden,
+    Middle,
 }
--- a/apple/Tests/TestCases/Address/AccountAddressTests.swift
+++ b/apple/Tests/TestCases/Address/AccountAddressTests.swift
@@ -29,6 +29,10 @@ final class AccountAddressTests: AddressTest<AccountAddress> {
 	func test_short() {
 		XCTAssertEqual(SUT.sample.shortFormat, "acco...nvjdwr")
 	}
+	
+	func test_middle() {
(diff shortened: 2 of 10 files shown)


src/profile/v100/address/address_format.rs
```diff
@@ -14,4 +14,5 @@ pub enum AddressFormat {
     Full,
     Raw,
     Default,
+    Hidden,
```
CyonAlexRDX, line 17: Can we try finding a more descriptive name of the variant (Rust calls "cases" "variants") than `Hidden`? I think `Hidden` is a bit ambigous, "format AccountAddress as Hidden" makes it sound like it will be the empty string, or a `"****"` string. So maybe, `Middle` or `GetMiddle` or `ExtractMiddle` or `GetOmitted` or `ExtractOmitted` etc. or "Center" instead of "Middle". Hmm or "Dropped" "Cut" "Ignored" are alternatives to "Omitted". `GetCutBody`? hmm

CyonAlexRDX (changes requested):
🎉 great first PR! Lets try finding a more precis term than `Hidden` IMO, and maybe a Swift test?

src/profile/v100/address/address_format.rs
CyonAlexRDX: https://www.thesaurus.com/browse/omit (I love using thesaurus)

matiasbzurovski: Makes sense to use something different than `Hidden`. I like just `Middle` given the other ones (`Full`, `Raw` & `Default`) are just adjectives (and have no verbs as the other proposals).

CyonAlexRDX: yeah sure, just `Middle` works.

CyonAlexRDX (approved):
Good job!

PR #106, opened 2024-04-25
Implement android parsing for MnemonicWithPassphrase

micbakos-rdx:
* Serialization of MnemonicWithPassphrase was different on Android than iOS, and since Rust's implementation follows the iOS logic, this PR aims to hide the difference by providing its own serializer in the android layer.


CyonAlexRDX (approved):
Good tmp solution!

PR #107, opened 2024-04-25
RCM: Radix Connect Relay

GhenadieVP:
https://radixdlt.atlassian.net/browse/ABW-3209

Added the Radix Connect Relay service that implements the necessary [API](https://github.com/radixdlt/radix-connect-relay?tab=readme-ov-file#api-v1) + the logic to encrypt/decrypt the requests/responses.

Note: This service is not meant to be used by the Wallets directly, so it(and the related models) are not exposed to the Wallets.

Additionally, sketched out the RadixConnectMobile UniFFI API that will actually be used by the Wallets.


src/core/assert_json.rs
```diff
@@ -54,12 +54,12 @@ fn base_assert_equality_after_json_roundtrip<T>(
     let serialized = serde_json::to_value(model).unwrap();
     let deserialized: T = serde_json::from_value(json.clone()).unwrap();
     if expect_eq {
-        assert_eq!(&deserialized, model, "Expected `model: T` and `T` deserialized from `json_string`, to be equal, but they were not.");
+        pretty_assertions::assert_eq!(&deserialized, model, "Expected `model: T` and `T` deserialized from `json_string`, to be equal, but they were not.");
```
GhenadieVP, line 57: For better assertion messages.

src/gateway_api/client/gateway_client.rs
```diff
@@ -2,20 +2,14 @@ use crate::prelude::*;
...
 ///
 /// [docs]: https://radix-babylon-gateway-api.redoc.ly/
 #[derive(uniffi::Object)]
 pub struct GatewayClient {
-    /// An object implementing the `NetworkAntenna` traits, which iOS/Android
-    /// clients pass into the constructor of this GatewayClient, so that it can
-    /// execute network requests.
-    pub network_antenna: Arc<dyn NetworkAntenna>,
+    /// The HTTP client that actually executes the network requests.
+    pub http_client: HttpClient,
```
GhenadieVP, line 12: HttpClient now is the actual wrapper for NetworkAntenna.

src/gateway_api/client/gateway_client.rs
```diff
@@ -52,55 +46,6 @@ impl GatewayClient {
     }
 }
 
-/// A mocked network antenna, useful for testing.
-#[derive(Debug)]
-struct MockAntenna {
```
GhenadieVP, line 57: Moved to NetworkAntenna

src/gateway_api/client/gateway_client.rs
```diff
@@ -120,91 +65,6 @@ mod tests {
     #[allow(clippy::upper_case_acronyms)]
     type SUT = GatewayClient;
 
-    #[actix_rt::test]
```
GhenadieVP, line 123: Moved to HttpClient

src/gateway_api/client/gateway_client_dispatch_request.rs
```diff
@@ -17,7 +17,23 @@ impl GatewayClient {
         U: for<'a> Deserialize<'a>,
         F: Fn(U) -> Result<V, CommonError>,
     {
-        self.dispatch_network_request(path, NetworkMethod::Post, request, map)
+        // Append relative path to base url
+        let path = path.as_ref();
```
GhenadieVP, line 21: Moved here from previous `dispatch` function

src/http_client/client.rs
```diff
@@ -0,0 +1,162 @@
...
+        let response = self
+            .network_antenna
+            .execute_network_request(request)
+            .await?;
+
+        // Check for valid status code
+        if !(200..=299).contains(&response.status_code) {
+            return Err(CommonError::NetworkResponseBadCode);
+        }
+
```
GhenadieVP, line 33: Note: previously of the body was empty, it did throw an error, but it is completely valid to get back an empty response for some requests. The validation moves to `model_from_response` where if a model is expected as response the decoding will fail for empty response.

src/radix_connect/mobile/deep_link_parsing/connect_request.rs
```diff
@@ -1,31 +1,31 @@
+use super::*;
 use crate::prelude::*;
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, uniffi::Enum)]
-pub enum MobileConnectRequest {
-    Link(LinkRequest),
-    DappInteraction(DappRequest),
+pub enum RadixConnectMobileConnectRequest {
```
GhenadieVP, line 5: Renamed to be more specific, as this is meant to be used by the Wallets.

src/radix_connect/mobile/deep_link_parsing/connect_request.rs
```diff
@@ -1,31 +1,31 @@
...
-impl FromStr for MobileConnectRequest {
+impl FromStr for RadixConnectMobileConnectRequest {
     type Err = CommonError;
 
     fn from_str(s: &str) -> Result<Self> {
         parse_mobile_connect_request(s)
     }
 }
 
-#[uniffi::export]
```
GhenadieVP, line 17: Moved to RadixConnectMobile

src/radix_connect/mobile/deep_link_parsing/mod.rs
GhenadieVP, line 1: Added a deep_link_parsing module to encapsulate the related files.

src/radix_connect/mobile/relay_service/service.rs
```diff
@@ -0,0 +1,420 @@
...
+    fn radix_connect_relay_request(request: Request) -> Result<Self> {
+        NetworkRequest::new_post(Url::from_str(SERVICE_PATH).unwrap())
+            .with_serializing_body(request)
+    }
+}
+
+impl Service {
+    async fn get_wallet_interaction_requests(
+        &self,
+        session: Session,
```
GhenadieVP, line 40: The session will be injected from the outside in the component using the relay service.

src/radix_connect/mobile/relay_service/models/request/request.rs
```diff
@@ -0,0 +1,306 @@
+use super::request_method::Method;
+use crate::prelude::*;
+use crate::radix_connect::mobile::session::session_id::SessionID;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct Request {
+    pub method: Method,
+    pub session_id: SessionID,
+    pub data: Option<BagOfBytes>,
```
CyonAlexRDX, line 10: do we plan to UniFFI export this? If not, prefer using `Vec<u8>` - we very much hope to remove `BagOfBytes` some days. The intended usage is to use `BagOfBytes` in all UniFFI APIs (functions) and only in UniFFI exported **types**.

src/radix_connect/mobile/relay_service/models/request/request.rs
```diff
@@ -0,0 +1,306 @@
...
+}
+
+impl Request {
+    pub fn new(
+        method: Method,
+        session_id: impl Into<SessionID>,
```
CyonAlexRDX, line 16: nice usage of `impl` in constructors 👍

src/radix_connect/mobile/relay_service/models/request/request.rs
```diff
@@ -0,0 +1,306 @@
...
+        Self {
+            method,
+            session_id: session_id.into(),
+            data: data.into(),
+        }
+    }
+}
+
+impl Request {
+    pub fn new_send_request(
```
CyonAlexRDX, line 28: I think we should rename this just `send_request`, that is the naming convention I use elsewhere, see e.g. `NonFungibleLocalId` (ops it should have been named `LocalID` my bad!!):
https://github.com/radixdlt/sargon/blob/develop/src/profile/v100/address/non_fungible_local_id.rs#L27-L45

src/radix_connect/mobile/relay_service/models/request/request.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    #[test]
+    fn send_request_json_roundtrip() {
+        let session_id = SessionID::sample();
+        let data = BagOfBytes::sample();
+        let request = SUT::new_send_request(session_id.clone(), data.clone());
+
+        let expected_json = format!(
+            r#"
+            {{
```
CyonAlexRDX, line 183: nice usage of format! and `{{` :)

src/radix_connect/mobile/relay_service/models/request/request_method.rs
```diff
@@ -0,0 +1,43 @@
+use crate::prelude::*;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub enum Method {
+    SendRequest,
```
CyonAlexRDX, line 6: please add at least a oneline documentation to all types and to all variants :)

src/radix_connect/mobile/relay_service/models/response/session_handshake_request.rs
```diff
@@ -0,0 +1,61 @@
+use crate::prelude::*;
+
+#[serde_as]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct SessionHandshakeRequest {
+    #[serde_as(as = "DisplayFromStr")]
+    pub public_key: PublicKey,
```
CyonAlexRDX, line 8: 👍

src/radix_connect/mobile/relay_service/models/request/request.rs
CyonAlexRDX: ok I changed my mind, lets don't do this. This actually defeats the purpose right? We are no longer asserting against something "recorded", so we should not use format, rather put explicit values. Otherwise we are not actually asserting that an expected to pass JSON value will successfully deserialize right?!

src/radix_connect/mobile/relay_service/models/response/session_handshake_request.rs
```diff
@@ -0,0 +1,61 @@
...
+        assert_ne!(SUT::sample(), SUT::sample_other());
+    }
+
+    #[test]
+    fn json_roundtrip() {
+        let original = SUT::sample();
+        let json = format!(
+            r#"
+            {{
+                "publicKey":"{}"
```
CyonAlexRDX, line 53: extra important here to NOT use `format! because if we change our impl of `Display` of `PublicKey`, this test will just work, but we might have broken Sargon!

src/radix_connect/mobile/relay_service/models/response/wallet_interaction_requests.rs
```diff
@@ -0,0 +1,57 @@
+use crate::prelude::*;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+struct WalletInteractionRequests(Vec<BagOfBytes>);
```
CyonAlexRDX, line 4: Os the plural correct? that this is many requests?

If we plan on UniFFI exporting `WalletInteractionRequests` then we might wanna use the `secret_magic` dance here, otherwise `WalletInteractionRequests` will become `public typealias WalletInteractionRequests = Data` in Swift, and we can accidantally use `Data()` in Swift, which is wrong. 

If we dont plan on UniFFI exporting it, we ought to use `Vec<u8>` as per earlier comment.

src/radix_connect/mobile/session/session.rs
```diff
@@ -0,0 +1,62 @@
+use super::session_id::SessionID;
+use super::session_origin::SessionOrigin;
+use crate::prelude::*;
+
+#[derive(Debug, PartialEq)]
+pub struct Session {
+    pub id: SessionID,
+    pub origin: SessionOrigin,
+    pub encryption_key: Exactly32Bytes,
```
CyonAlexRDX, line 9: Ah, ok you can for now change to `Entropy32Bytes` (might wanna rename it later) which implements `Zeroize` and then you can add `Zeroize` to the derive.

CyonAlexRDX (changes requested):
Nice! Some few missing lines of code coverage, but most importantly I think we ought not not use dynamic JSON expected values, I think that defeats the purpose.

src/radix_connect/mobile/relay_service/models/request/request.rs
GhenadieVP: no, this is not meant to be expose through UniFFI

src/radix_connect/mobile/relay_service/models/request/request.rs
GhenadieVP: I just wanted to group together the constructors with the `new` prefix

src/radix_connect/mobile/relay_service/models/request/request_method.rs
GhenadieVP: Will link the docs of RadixConnectRelay

src/radix_connect/mobile/relay_service/models/response/wallet_interaction_requests.rs
GhenadieVP: Yes, the RadixConnectRelay returns a list of requests for a given session id, no only one.

And no, this is not meant to be exposed through UniFFI.

src/radix_connect/mobile/relay_service/models/request/request.rs
CyonAlexRDX: understood, but I follow a pattern which RET and Scrypto also follows, so I think best not use `new`. 

(Juxtaposition: however, for UniFFI exported global functions which are "constructors" we *always* use `new_name_of_type`)

Sajjon: @GhenadieVP oh I realized I was in the wrong! Lets DO USE `new_` because this is a `struct`! Not an enum! :) and lets use the name of the variant for enums only... sorry! :D

src/radix_connect/mobile/session/session.rs
GhenadieVP: dumb question I guess - Is `Entropy32Bytes` applicable here also? I mean this would be later a key derived from DH key agreement, isn't Entropy32Bytes about having a key derived from mnemonic?

CyonAlexRDX (approved):
LGTM 🎉

PR #108, opened 2024-04-26
Batch 12: Add analyze profile file contents

CyonAlexRDX:
Will be used by Wallets to first analyze file contents of a file selected by user to import. If it is case/variant `Encrypted` wallets should prompt users for decryption password.


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "sargon"
-version = "0.6.44"
+version = "0.6.51"
```
CyonAlexRDX, line 3: @GhenadieVP @jakub-rdx @matiasbzurovski @micbakos-rdx lets try to bump the version to the lastest you see in [Releases](https://github.com/radixdlt/sargon/releases) **`+1`**, and if anyone else merges their Sargon PR before you, please bump again.

We dont have time to fix this to be done automatically by CD.... it is very complex working with CD in Sargon and we have something which works now (took a lot of job by me and Michael), lets try to improve it later.

HOW IS THIS USED?
It is used by:

```rust
#[uniffi::export]
pub fn build_information() -> SargonBuildInformation {
    SargonBuildInformation::get()
}
```

which wallets can use to display the Sargon version (this version, not the GH release tag....) in DEBUG builds under app settings:
IMG_FA304BA3C81A-1

micbakos-rdx (approved):

PR #111, opened 2024-04-29
Expose deep link parsing in kotlin

jakub-rdx:
- expose deep link parsing in kotlin
- add unit tests
- convert uniffi exported `Url` to `java.net.URI` so that we can handle `RadixMobileConnectRequest.Link` origin property with arbitrary `protocol`, like `radix://xyz`

### DIFF
--- a/jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/MobileConnectRequest.kt
+++ b/jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/MobileConnectRequest.kt
@@ -1,9 +1,8 @@
 package com.radixdlt.sargon.extensions
 
-import com.radixdlt.sargon.MobileConnectRequest
+import com.radixdlt.sargon.RadixConnectMobileConnectRequest
 import com.radixdlt.sargon.newMobileConnectRequest
 
-@Throws(SargonException::class)
-fun MobileConnectRequest.Companion.parseFrom(url: String) = runCatching {
+fun RadixConnectMobileConnectRequest.Companion.parseFrom(url: String) = runCatching {
     newMobileConnectRequest(url = url)
 }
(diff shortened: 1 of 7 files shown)


jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/MobileConnectRequest.kt
```diff
@@ -0,0 +1,9 @@
+package com.radixdlt.sargon.extensions
+
+import com.radixdlt.sargon.MobileConnectRequest
+import com.radixdlt.sargon.newMobileConnectRequest
+
+@Throws(SargonException::class)
+fun MobileConnectRequest.Companion.parseFrom(url: String) = runCatching {
```
CyonAlexRDX, line 7: should probably not only accept `String` should probably have a method taking `Uri` too, update Rust Sargon accordingly if applicable?

jakub-rdx: why? deep link comes in to android wallet as `android.net.Uri` so even different type then `java.net.URI`

Sajjon: Ok.

We can also do a UniFFI newtype, that translates into `android.net.Uri` and use it here. That way you need not pass in a `String`.

But maybe not worth it.

jakub-rdx: using android sdk classes in business logic mean that we need android SDK to run tests, we prefer to avoid it to have them easier and faster to run. If you really want I can pass `Uri` using `java.net.Uri`, converting deep link to string first and then to `URI`

CyonAlexRDX (approved):
LGTM . Maybe await approval from @micbakos-rdx ?

PR #113, opened 2024-04-29
Fix AddressFormat::Middle formatting on NonFungibleGlobalId

matiasbzurovski:
## Changelog
Fixes the formatting made on `NonFungibleGlobalid` for type `Middle`.


src/profile/v100/address/non_fungible_global_id.rs
```diff
@@ -305,11 +313,53 @@ mod tests {
     }
 
     #[test]
-    fn formatted_middle() {
+    fn formatted_default_vs_middle() {
```
matiasbzurovski, line 316: This is the best test to understand the expected behavior on each case

apple/Tests/TestCases/Prelude/NonFungibleGlobalIDTests.swift
```diff
@@ -41,14 +41,25 @@ final class NonFungibleGlobalIDTests: IdentifiableByStringProtocolTest<NonFungib
...
         let sut = try SUT(
             nonFungibleResourceAddress: .sample,
             localID: .ruid(
                 value: .init(hex: "deadbeef12345678babecafe87654321fadedeaf01234567ecadabba76543210")
             )
         )
-        XCTAssertNoDifference(sut.formatted(.default), "reso...c9wlxa:dead...3210")
         XCTAssertNoDifference(sut.formatted(.raw), "resource_rdx1nfyg2f68jw7hfdlg5hzvd8ylsa7e0kjl68t5t62v3ttamtejc9wlxa:{deadbeef12345678-babecafe87654321-fadedeaf01234567-ecadabba76543210}")
+		XCTAssertNoDifference(sut.formatted(.default), "reso...c9wlxa:dead...3210")
+		XCTAssertNoDifference(sut.formatted(.middle), "urce_rdx1nfyg2f68jw7hfdlg5hzvd8ylsa7e0kjl68t5t62v3ttamtej:beef12345678-babecafe87654321-fadedeaf01234567-ecadabba7654")
```
matiasbzurovski, line 53: Looks like this file was using spaces instead of tabs for idents. Continuing with the discussion that @GhenadieVP brought on some other place, I think it would be better to go with third party tool [SwiftFormat](https://github.com/nicklockwood/SwiftFormat) rather than Apple's [swift-format](https://github.com/apple/swift-format).

I played a little bit with the latter and personally think the first one is on a better shape. Also, as we are already using SwiftFormat on the iOS wallet, it would be easier to keep the same configuration/standards on both projects.

CyonAlexRDX (approved):
LGTM

PR #116, opened 2024-04-30
WIP [RCM] - Linking and dapp request handling.

GhenadieVP:


src/core/types/keys/diffie_hellman/private_key.rs
```diff
@@ -0,0 +1,195 @@
+use crate::prelude::*;
+use crypto::keys::x25519::SecretKey as X25519PrivateKey;
+use hkdf::Hkdf;
+use k256::sha2::Sha256;
+
+#[derive(derive_more::Debug)]
+#[debug("{}", self.to_hex())]
+pub struct DiffieHellmanPrivateKey(X25519PrivateKey);
```
GhenadieVP, line 8: @CyonAlexRDX Here I have added the DH Private/Public key pair, please advice if something should be different.

src/core/types/keys/diffie_hellman/private_key.rs
```diff
@@ -0,0 +1,195 @@
...
+impl DiffieHellmanPrivateKey {
+    pub fn generate() -> Result<Self> {
+        Exactly32Bytes::generate().to_vec().try_into()
+    }
+
+    pub fn public_key(&self) -> DiffieHellmanPublicKey {
+        self.0.public_key().into()
+    }
+
+    pub fn hkdf_key_agreement(
```
GhenadieVP, line 19: @CyonAlexRDX HKDF key agreement, seems to be the proper way, but let me know.

src/radix_connect/mobile/client.rs
```diff
@@ -1,49 +1,174 @@
...
+            .relay_service
+            .get_session_handshake_request(request.session_id)
+            .await?;
+
+        // 2. Generate a new Diffie-Hellman key pair
+        let wallet_private_key = DiffieHellmanPrivateKey::generate()?;
+
+        // 3. Compute the shared secret
+        // random salt
+        let salt = hex_decode("000102030405060708090a0b0c0d0e0f").unwrap();
```
GhenadieVP, line 57: @CyonAlexRDX What values are the most appropriate and how should we store this? I am not sure these should be in plain sight, but maybe it is ok.

Cargo.toml
```diff
@@ -132,6 +132,7 @@ iota-crypto = { git = "https://github.com/iotaledger/crypto.rs", rev = "47460d64
     "slip10",
     "ed25519",
     "secp256k1",
+    "x25519",
```
CyonAlexRDX, line 135: Maybe add "zeroize"` too (they other flags might have activated it already, but does not hurt)

src/core/types/keys/diffie_hellman/private_key.rs
```diff
@@ -0,0 +1,195 @@
...
+        &self,
+        other: &DiffieHellmanPublicKey,
+        salt: &[u8],
```
CyonAlexRDX, line 22: can be:
`salt: impl AsRef<[u8]>` and use with `as_ref()` in body below, that way we can pass `BagOfBytes`, `Vec`, or `ExactlyNBytes` types as salt 

and do same for `into` ?

src/core/types/keys/diffie_hellman/private_key.rs
```diff
@@ -0,0 +1,195 @@
...
+            }
+        })?;
+
+        Ok(Exactly32Bytes::from(&symmetric_key))
+    }
+}
+
+impl PartialEq for DiffieHellmanPrivateKey {
+    fn eq(&self, other: &Self) -> bool {
+        self.to_bytes() == other.to_bytes()
```
CyonAlexRDX, line 41: safer to do `self.public_key() == other.public_key()`

src/core/types/keys/diffie_hellman/private_key.rs
CyonAlexRDX: maybe time for us to do a `pub type SymmetricKey = Exactly32Bytes` typealias? and use it here and in VersionedEncryption (and its impl, i.e. search for `encryption_key` and `decryption_key`), I think it is more clear.

src/core/types/keys/diffie_hellman/private_key.rs
```diff
@@ -0,0 +1,195 @@
...
+        &self,
+        other: &DiffieHellmanPublicKey,
+        salt: &[u8],
+        info: &[u8],
+    ) -> Result<Exactly32Bytes> {
+        let shared_secret = self.0.diffie_hellman(&other.0).to_bytes();
+        let mut symmetric_key = [0u8; 32]; // 32-byte buffer for the symmetric key
+
+        let hkdf = Hkdf::<Sha256>::new(Some(salt), &shared_secret);
+        hkdf.expand(info, &mut symmetric_key).map_err(|err| {
```
CyonAlexRDX, line 29: right, so we already have this... in `pb_hkdf_sha256` which uses password as input key material (`ikm`), and here we use `shared_secret` as `ikm`. Maybe worth generalizing. And I have already verified that it is safe to unwrap, so so this method need not be returning `Result` and you can remove the `HkdfExpandFailed` (you can verify my verification if you want :) but for our usage it is not possible to crash).

CyonAlexRDX: so I propose you break out the `Hkdf` into its own module, and let it have this API:

```
pub fn hkdf_sha256(ikm: impl AsRef<[u8]>, salt: impl Into<Option<&[u8]>>, info: impl Into<Option<&[u8]>> -> Exactly32Bytes
```

and reuse it in `PbHkdfSha256` and here.

src/core/types/keys/diffie_hellman/private_key.rs
CyonAlexRDX: Perhaps `KeyAgreementPrivateKey` is better than `DiffieHellman` ? Like CryptoKit uses.

src/radix_connect/mobile/client.rs
```diff
@@ -1,49 +1,174 @@
...
-//         _response: WalletToDappInteractionResponse,
-//     ) -> Result<Url> {
-//         todo!()
-//     }
-// }
+#[derive(uniffi::Object)]
+pub struct RadixConnectMobile {
+    /// The Radix Connect Relay service to be used to communicate with dApps.
+    relay_service: RelayService,
+    /// The secure storage to be used to store session data.
```
CyonAlexRDX, line 14: I prefer an empty line between fields

src/radix_connect/mobile/client.rs
```diff
@@ -1,49 +1,174 @@
...
+        Self {
+            relay_service: RelayService::new_with_network_antenna(
+                network_antenna,
+            ),
+            secure_storage: WalletClientStorage::new(secure_storage),
+            new_sessions: RwLock::new(HashMap::new()),
+        }
+    }
+
+    #[uniffi::method]
```
CyonAlexRDX, line 41: The `#[uniffi::method]` is not needed, only used if we wanna rename or set default values.

src/radix_connect/mobile/client.rs
CyonAlexRDX: The `salt` and the `info` must be known by both parties (wallet and web-counter-part) or something both can derive/calculate from shared knowledge.

Should this method, `handle_linking_request` save the `wallet_private_key` into `secure_storage`? or is it ephemeral? if it should be saved (that call is missing) then we should try to find a suitable salt/info. 

If it is ephemeral, salt and info is not as important.

So talk with @xstelea if there is any shared knowledge both parties have? Origin? Url? Some headers? Other metadata? Some stable onchain data?

src/radix_connect/mobile/client.rs
```diff
@@ -1,49 +1,174 @@
...
+            encryption_key,
+        );
+
+        {
+            self.new_sessions
+                .try_write()
+                .map(|mut new_sessions| {
+                    new_sessions.insert(request.session_id, session);
+                })
+                .unwrap();
```
CyonAlexRDX, line 78: an `expect` is preferable here `expect("Should be able to write to new_session since we handle one request at a time")` or similar

src/radix_connect/mobile/client.rs
```diff
@@ -1,49 +1,174 @@
...
+        }
+
+        // 5. TODO: use the actual dapp callback path
+        Ok(Url::from_str("https://example.com").unwrap())
```
CyonAlexRDX, line 82: I know this is gonna change, but reminder about `parse_url`: https://github.com/radixdlt/sargon/blob/develop/src/core/utils/string_utils.rs#L33

so just:

```rust
// 5. TODO: use the actual dapp callback path
parse_url("https://example.com")
```

src/radix_connect/mobile/client.rs
```diff
@@ -1,49 +1,174 @@
...
+    }
+
+    #[uniffi::method]
+    pub async fn handle_dapp_interaction_request(
+        &self,
+        dapp_request: RadixConnectMobileDappRequest,
+    ) -> Result<RadixConnectMobileSessionRequest> {
+        let session;
+        {
+            let new_sessions = self.new_sessions.try_read().unwrap();
```
CyonAlexRDX, line 92: prefer an `expect` here, the read counterpart of https://github.com/radixdlt/sargon/pull/116/files#r1584698094

src/radix_connect/mobile/client.rs
```diff
@@ -1,49 +1,174 @@
...
+
+            session = match new_sessions.get(&dapp_request.session_id) {
+                Some(session) => {
+                    // Save the session to the secure storage
+                    self.secure_storage.save_session(session.clone())?;
+                    let session_id = session.id;
+
+                    self.new_sessions
+                        .try_write()
```
CyonAlexRDX, line 101: ah third place you access the `RwLock` (two writes and one read), consider using this pattern I use for `Wallet`'s `RwLock` of `Profile:

https://github.com/radixdlt/sargon/blob/develop/src/wallet/wallet.rs#L214-L222

and 
https://github.com/radixdlt/sargon/blob/develop/src/wallet/wallet.rs#L240-L253

src/radix_connect/mobile/client.rs
```diff
@@ -1,49 +1,174 @@
...
+                })
+                .unwrap();
+        }
+
+        // 5. TODO: use the actual dapp callback path
+        Ok(Url::from_str("https://example.com").unwrap())
+    }
+
+    #[uniffi::method]
+    pub async fn handle_dapp_interaction_request(
```
CyonAlexRDX, line 86: needs a short doc

src/radix_connect/mobile/client.rs
```diff
@@ -1,49 +1,174 @@
...
+            ),
+            secure_storage: WalletClientStorage::new(secure_storage),
+            new_sessions: RwLock::new(HashMap::new()),
+        }
+    }
+
+    #[uniffi::method]
+    pub async fn handle_linking_request(
+        &self,
+        request: RadixConnectMobileLinkRequest,
```
CyonAlexRDX, line 44: needs a short doc

src/radix_connect/mobile/client.rs
```diff
@@ -1,49 +1,174 @@
...
+    #[uniffi::method]
+    pub async fn handle_dapp_interaction_request(
+        &self,
+        dapp_request: RadixConnectMobileDappRequest,
+    ) -> Result<RadixConnectMobileSessionRequest> {
+        let session;
+        {
+            let new_sessions = self.new_sessions.try_read().unwrap();
+
+            session = match new_sessions.get(&dapp_request.session_id) {
```
CyonAlexRDX, line 94: maybe some one line doc relating to the documentation of `new_sessions` here ? and maybe `Some(session) =>` can be `Some(in_flight_session) =>` ? and rename `new_sessions` to `in_flight_session`?  or similar.

src/radix_connect/mobile/client.rs
```diff
@@ -1,49 +1,174 @@
...
+#[derive(uniffi::Object)]
+pub struct RadixConnectMobile {
+    /// The Radix Connect Relay service to be used to communicate with dApps.
+    relay_service: RelayService,
+    /// The secure storage to be used to store session data.
+    secure_storage: WalletClientStorage,
+    /// The new sessions that have been created and are waiting to be validated on dApp side.
+    /// Once the session is validated, it will be moved to the secure storage.
+    /// Validation consists in verifying the origin of the session.
+    new_sessions: RwLock<HashMap<SessionID, Session>>,
```
CyonAlexRDX, line 19: I think a `pub type InFlightSession = Session;` and -> `in_flight_sessions: RwLock<HashMap<SessionID, InFlightSessions>>` 

or maybe "In Flight" is not best... maybe `PendingLinkingCompletedSession` and `pending_linking_completed_sessionss` etc?

CyonAlexRDX: see related comment https://github.com/radixdlt/sargon/pull/116/files#r1584715414

src/radix_connect/mobile/client.rs
CyonAlexRDX: see related comment: https://github.com/radixdlt/sargon/pull/116/files#r1584717789

src/radix_connect/mobile/client.rs
```diff
@@ -1,49 +1,174 @@
...
+        }
+
+        self.relay_service
+            .get_wallet_interaction_requests(session)
+            .await?
+            .into_iter()
+            .find(|intraction| {
+                intraction.interaction_id == dapp_request.interaction_id
+            })
+            .map(|interaction| RadixConnectMobileSessionRequest {
```
CyonAlexRDX, line 122: minor: I prefer to handle the bad case first, i.e. invert the last `map` and the `ok_or`, since that way the last statements **type** matches the return type of the fn, which I like.

CyonAlexRDX: prefer `RadixConnectMobileSessionRequest::new` inside the map. kind of like we in Swift tend to have one single init all other calls, it is advantageous for debugging (and logging) to have one shared entry )

src/radix_connect/mobile/session/session.rs
```diff
@@ -2,7 +2,7 @@ use super::session_id::SessionID;
 use super::session_origin::SessionOrigin;
 use crate::prelude::*;
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
```
CyonAlexRDX, line 5: I think `Session` can be `Copy` too, and then run `cargo clippy --fix --allow-staged`

CyonAlexRDX: Oh never mind, the `SessionOrigin` is not copy... and cannot due to `Url` not being copy...

src/radix_connect/mobile/session_dapp_request.rs
```diff
@@ -0,0 +1,54 @@
+use crate::prelude::*;
+
+#[derive(Debug, PartialEq, uniffi::Record)]
+pub struct RadixConnectMobileSessionRequest {
```
CyonAlexRDX, line 4: some few lines of doc would be nice!

src/radix_connect/mobile/session_wallet_response.rs
```diff
@@ -0,0 +1,54 @@
+use crate::prelude::*;
+
+#[derive(Debug, PartialEq, uniffi::Record)]
+pub struct RadixConnectMobileWalletResponse {
+    pub session_id: SessionID,
+    pub response: WalletToDappInteractionResponse,
```
CyonAlexRDX, line 6: some few lines of doc would be nice

PR #117, opened 2024-04-30
Guard for wrong curve discriminator

micbakos-rdx:
* Guard sargon against the case were the curve discriminator is defined as `curve25519` but the `compressedData` is in fact `secp256k1`. 

This is a bug introduced in Android version 1.5.0 which may result in profile's wrongly reporting the curve.


CyonAlexRDX (approved):
LGTM

PR #119, opened 2024-05-01
Profile integration stage 1 cyon batch 16

CyonAlexRDX:


apple/Sources/Sargon/Extensions/Methods/Profile/ProfileFileContents+Wrap+Functions.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+//  Created by Alexander Cyon on 2024-05-01.
+//
+
+import Foundation
+import SargonUniFFI
+
+extension ProfileFileContents {
+	public func hash(into hasher: inout Hasher) {
+		hasher.combine(profileFileContentsHashValue(contents: self))
```
Sajjon, line 13: `ProfileFileContents ` is an enum with the associated value of the `PlaintextProfile` being `RefProfile` - a reference type. Thus UniFFI cannot make `ProfileFileContents` conform to `Hashable` / `Equatable`. We have to build it ourselves. I've implemented in `Eq` and `Hash` in Rust and UniFFI exported it...

apple/Sources/Sargon/Extensions/SampleValues/Profile+Supporting+Types/References/RefBytes+SampleValues.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+import SargonUniFFI
+
+#if DEBUG
+extension RefBytes {
+	public static var sample: Self { newRefBytesSample() as! Self }
```
Sajjon, line 6: cannot use `static let sample: Self = newRefBytesSample()` due to covariance.

apple/Sources/Sargon/Extensions/Swiftified/Profile+Supporting+Types/ProfileFileContents+Swiftified.swift
```diff
@@ -0,0 +1,12 @@
...
+//  File.swift
+//  
+//
+//  Created by Alexander Cyon on 2024-05-01.
+//
+
+import Foundation
+import SargonUniFFI
+
+extension ProfileFileContents: @unchecked Sendable {}
```
Sajjon, line 11: Must be marked `@unchecked Sendable` since contains reference type.

apple/Sources/Sargon/Extensions/Swiftified/Profile+Supporting+Types/References/RefBytes+Swiftified.swift
```diff
@@ -0,0 +1,22 @@
...
+//  
+//
+//  Created by Alexander Cyon on 2024-05-01.
+//
+
+import Foundation
+import SargonUniFFI
+
+extension BagOfBytes: SargonModel {
+	public static let sample: Self = newBagOfBytesSampleAced()
```
Sajjon, line 12: Can be moved elsewhere, but meh, actually just wanted it because I put constraint `Inner: SargonModel`. Remember that `BagOfBytes` **is** `Data`, but put `extension BagOfBytes` so it "echoes" with `newBagOfBytes` below.

apple/Sources/Sargon/Extensions/Swiftified/Profile+Supporting+Types/References/RefProfile+Swiftified.swift
```diff
@@ -0,0 +1,23 @@
...
+//  Created by Alexander Cyon on 2024-05-01.
+//
+
+import Foundation
+import SargonUniFFI
+
+public protocol SargonReferenceType: SargonModel {
+	associatedtype Inner: SargonModel
+	func take() throws -> Inner
+	static func from(inner: Inner) -> Self
```
Sajjon, line 14: cannot use `init(inner: Inner)` since `RefBytes` and `RefProfile` is not `final` an those initializers are `convenience`....

apple/Tests/TestCases/Profile/ProfileTests.swift
```diff
@@ -83,4 +80,121 @@ final class ProfileTests: Test<Profile> {
...
+	// M2 Max: Average 0.6
+	// OMITTED - too slow (6 seconds)
+	func json_encoding_slow_performance() throws {
+		let (sut, _) = vector
+		measure {
+			let _ = profileToJsonBytes(profile: sut)
+		}
+	}
+	
+	// M2 Max: Average 0.003 (200x as fast)
```
Sajjon, line 109: ‼️ 200x speedup vs slow.

Compared to the previous impl of `Encodable` it is a `2000x` speed up. Yes TWO THOUSAND times faster.

apple/Tests/Utils/RefTest.swift
```diff
@@ -0,0 +1,23 @@
...
+import Foundation
+import Sargon
+import SargonUniFFI
+import XCTest
+
+class RefTest<SUT_: SargonReferenceType>: Test<SUT_> {
+	func test_take_twice_throws() {
+		func doTest(_ sut: SUT) {
+			XCTAssertNoThrow(try sut.take())
+			XCTAssertThrowsError(try sut.take())
```
Sajjon, line 11: `take()` **consumes** `inner` **without making any expensive `clone()`**, and will throw if call a second time.

src/profile/encrypted/encryption/aes_gcm_256.rs
```diff
@@ -23,7 +23,7 @@ impl AesGcm256 {
 
 impl AesGcm256 {
     fn seal(
-        plaintext: Vec<u8>,
+        plaintext: impl AsRef<[u8]>,
```
Sajjon, line 26: we should try to always use `impl AsRef<[u8]>` instead of `BagOfBytes` / `Vec<u8>` or `&[u8]` so that we can pass any :) can also pass `ExactlyNBytes` types etc.

src/wallet/wallet_accounts.rs
```diff
@@ -405,6 +405,45 @@ mod tests {
         assert_eq!(loaded, private);
     }
 
+    #[test]
+    #[ignore]
```
Sajjon, line 409: ignored, will not be run, you can use this if you need to generate big profiles.

src/radix_connect/mobile/relay_service/service.rs
```diff
@@ -52,22 +52,14 @@ impl Service {
...
-                    new_dapp_to_wallet_interaction_unvalidated_from_json_bytes(
-                        bytes.clone().into(),
-                    )
-                })
-                .collect();
-
-        deserialized_wallet_interactions_result
+        decrypted_wallet_interactions
+            .iter()
+            .map(DappToWalletInteractionUnvalidated::new_from_json_bytes)
```
Sajjon, line 61: a bit shorter

src/profile/v100/profile.rs
```diff
@@ -46,39 +46,15 @@ pub struct Profile {
...
-/// which we were able to successfully JSON deserialize from the bytes,
-/// or EncryptedProfile for which wallets will continue prompting the
-/// user for an encryption password and then call JSON deserialize
-/// of `EncryptedProfileSnapshot` using [`Profile::new_from_encryption_bytes`](Profile::new_from_encryption_bytes)
-/// or if we failed to parse as Profile and `EncryptedProfileSnapshot`
-/// then `NotProfile` is used, indicating that the bytes is not at all
-/// a Profile.
-#[derive(Clone, PartialEq, Eq, Debug, Hash, uniffi::Enum)]
-#[allow(clippy::large_enum_variant)]
-pub enum ProfileFileContents {
```
Sajjon, line 59: moved into its own file

src/profile/references/ref_macro.rs
```diff
@@ -0,0 +1,173 @@
...
+        $inner: ty
+    ) => {
+        paste! {
+            use crate::prelude::*;
+            use std::sync::{Arc, RwLock};
+
+            #[derive(Debug, uniffi::Object)]
+            #[uniffi::export(Debug, Eq, Hash)]
+            pub(crate) struct $struct_name {
+                inner: RwLock<Option<$inner>>,
```
Sajjon, line 17: This is an EXTREMELY useful trick! We are inside a type which is `uniffi::Object` exported which we use with `Arc`. However, in Rust, an `Arc`:ed type **cannot be mutated**, a method which would otherwise be marked `&mut self` cannot be used. However, we can use **interior mutability**, such as `Cell` / `RefCell`, unfortunately neither of these can be used with UniFFI IIRC. But `RwLock` works with UniFFI. So how do we mutate the inner value without mutating the outer Object? Well we can "take" the `Some($inner)` from the `RwLock` leaving it with `None`, which is OK since it is interior mutability!

Sajjon: See `take` method below!

src/profile/references/ref_macro.rs
```diff
@@ -0,0 +1,173 @@
...
+                #[uniffi::constructor]
+                pub(crate) fn new(inner: $inner) -> Arc<Self> {
+                    Arc::new(Self::from(inner))
+                }
+
+                pub(crate) fn take(self: Arc<Self>) -> Result<$inner> {
+                    self.inner
+                        .try_write()
+                        .expect("Should only acquire write lock once.")
+                        .take()
```
Sajjon, line 81: the `take` call on the RwLockWriteGuard here replaces `Some` with `None` and returns the value. This consuming this reference type and returning the inner value. 

This function cannot be called twice in Rust, thanks to `self` being consumed (not passed by reference) - so we are protected in the land of Rust. And it returns `Result` so is throwing in the FFI land (and WILL throw if called twice in FFI land).

apple/Sources/Sargon/Extensions/SampleValues/Profile+Supporting+Types/References/RefBytes+SampleValues.swift
kugel3: I suppose you could do
```swift
public static let sample: Self = newRefBytesSample() as! Self
```

Sajjon: I thought so to, alas, no (feels like a bug actually):

Screenshot 2024-05-02 at 09 42 59

apple/Sources/Sargon/Extensions/Methods/Profile/Profile+Wrap+Functions.swift
```diff
@@ -1,32 +1,56 @@
...
+			reference: RefBytes(inner: Data(bytes)),
 			decryptionPassword: decryptionPassword
-		)
+		).take()
 	}
-	
+
 	public func profileSnapshot() -> Data {
-        profileToJsonBytes(profile: self)
+		try! profileSnapshotRef().take()
```
GhenadieVP, line 40: Intentional crash?

apple/Sources/Sargon/Extensions/Methods/Profile/Profile+Wrap+Functions.swift
```diff
@@ -1,32 +1,56 @@
...
+		profileToJsonBytesFastByRef(reference: RefProfile(inner: self))
 	}
-	
-	public func encrypt(password: String) -> Data {
-		profileEncryptWithPassword(profile: self, encryptionPassword: password)
+
+	public func encrypt(
+		password: String
+	) -> Data {
+		try! profileEncryptWithPasswordFastByRef(
```
GhenadieVP, line 51: Intentional crash?

apple/Sources/Sargon/Extensions/Swiftified/Profile+Supporting+Types/References/RefProfile+Swiftified.swift
GhenadieVP: Can't we make `RefBytes` and `RefProfile` final? is there a reason for these to not be?

apple/Tests/TestCases/Profile/ProfileTests.swift
GhenadieVP: 200x or 2000x :))?

kugel3:
> Furthermore Profile is no longer Codable in Swift, instead use Profile:init:jsonBytesReference: RefBytes to JSON deserialize bytes into Profile and use profileSnapshotRef() on Profile to get RefBytes containing JSON serialized profile.

Could this not be encapsulated in the wrapper?

apple/Sources/Sargon/Extensions/Swiftified/Profile+Supporting+Types/References/RefProfile+Swiftified.swift
Sajjon: not possible by UniFFI.... we would need to add more logic to our custom Post Processing step. I lack the Regexp skills :P

apple/Sources/Sargon/Extensions/SampleValues/Profile+Supporting+Types/References/RefBytes+SampleValues.swift
kugel3: Ah, but what if you just replace `Self` with `RefBytes`?

apple/Tests/TestCases/Profile/ProfileTests.swift
Sajjon: 200x Rust fast vs Rust slow  - 2000x vs `Swift.Encodable` solution, i.e. the `Swift.Encodable` wrapper made the Rust call take 10x the time.

apple/Sources/Sargon/Extensions/Swiftified/Profile+Supporting+Types/References/RefProfile+Swiftified.swift
GhenadieVP: well, we did this previously, "by hand" for RET types...
And really there is no reason for Sargon objects to be non-final, as those are not meant to be subclassed, though it is interesting what the behaviour is if we do so 😄

apple/Sources/Sargon/Extensions/SampleValues/Profile+Supporting+Types/References/RefBytes+SampleValues.swift
Sajjon: Then it does not fulfill the protocol requirement :D 

```swift
protocol SargonModel {
    static var sample: Self { get }
    static var sampleOther: Self { get }
}
```
oh non final classes... such crap.

apple/Sources/Sargon/Extensions/Swiftified/Profile+Supporting+Types/References/RefProfile+Swiftified.swift
Sajjon: Nothing should be done by hand. I have created the context where it should happen, the post process script I invented :) I just think we ought to do it with a regexp. I can take a look later.

apple/Sources/Sargon/Extensions/Methods/Profile/Profile+Wrap+Functions.swift
Sajjon: not possible to crash.

apple/Sources/Sargon/Extensions/Swiftified/Profile/Profile+Swiftified.swift
```diff
@@ -1,7 +1,19 @@
 import SargonUniFFI
 
 extension Profile: SargonModel {}
-extension Profile: SargonObjectCodable {}
+
+@available(*, unavailable, message: "Profile should not use Swift `Encodable` (Codable), since it is too slow.")
+extension Profile: Encodable {
+	@available(*, unavailable, message: "Profile should not use Swift `Encodable` (Codable), since it is too slow.")
+	public func encode(to encoder: any Encoder) throws { fatalError("Unreachable") }
```
Sajjon, line 8: CodeCov will complain about these lines, and they are.... completely untestable.

GhenadieVP (approved):

PR #120, opened 2024-05-01
Extensions for `EntropyNBytes` 

micbakos-rdx:
* Extensions for `EntropyNBytes` added
* Android's [`SecureRandom`](https://developer.android.com/reference/kotlin/java/security/SecureRandom) is used as an internal function to produce entropy. Each entropy class exposes a `random()` factory.
* Two methods exposed for notarizing signed intent hash and calculating public key from said entropy.


src/core/types/secret_bytes.rs
```diff
@@ -53,6 +53,12 @@ macro_rules! decl_secret_bytes {
                     .map(|secret_magic| $struct_name { secret_magic })
             }
 
+           impl $struct_name {
+                pub fn to_bytes(&self) -> &[u8] {
+                    &self.secret_magic.0.as_slice()
```
CyonAlexRDX, line 58: please change `[< $struct_name:snake _to_bytes >](bytes: &$struct_name) ` below to use this method

jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/Entropy32Bytes.kt
```diff
@@ -0,0 +1,34 @@
...
+fun Entropy32Bytes.Companion.init(bytes: BagOfBytes): Entropy32Bytes =
+    newEntropy32BytesFromBytes(bytes = bytes)
+
+val Entropy32Bytes.Companion.LENGTH: Int
+    get() = 32
+
+fun Entropy32Bytes.asGeneral() = Bip39Entropy.EntropyOf32Bytes(this)
+
+@Throws(SargonException::class)
+fun Entropy32Bytes.notarize(signedIntentHash: SignedIntentHash): NotarySignature =
```
CyonAlexRDX, line 26: Personally I would not create `notarize` as a method on `Entropy32Bytes`, I find it a bit strange. I think I would create my own `data class`: `NotaryPrivateKey` which has `notarize` and the public key func returning `Ed25519PublicKey`.

CyonAlexRDX (approved):
LGTM, left some minor remarks.

jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/Entropy32Bytes.kt
micbakos-rdx: You are right was thinking about it too.

jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/NotaryPrivateKey.kt
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    override fun hashCode(): Int {
+        return entropy32Bytes.hashCode()
+    }
+
+    @KoverIgnore
+    override fun toString(): String {
+        return "NotaryPrivateKey(entropy32Bytes=$entropy32Bytes)"
+    }
```
micbakos-rdx, line 47: IDE auto-generated methods when we create just a class instead of data class. The reason it is just a class is because we need to hide the constructor to the end user and instead provide a factory that is similar to how the PrivateKey api was.

PR #121, opened 2024-05-02
update CI / CD to use main branch instead of develop

CyonAlexRDX:
I've also changed the default branch to be `main` (and changes the branch protection rules to use `main`)

### DIFF
--- a/apple/Tests/TestCases/Prelude/DependencyInformationTests.swift
+++ b/apple/Tests/TestCases/Prelude/DependencyInformationTests.swift
@@ -6,6 +6,6 @@ import XCTest
 
 final class DependencyInformationTests: Test<DependencyInformation> {
 	func test_description() {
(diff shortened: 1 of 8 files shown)


apple/Tests/TestCases/Prelude/DependencyInformationTests.swift
```diff
@@ -6,6 +6,6 @@ import XCTest
 
 final class DependencyInformationTests: Test<DependencyInformation> {
 	func test_description() {
-		XCTAssertNoDifference(SUT.sample.description, "develop")
+		XCTAssertNoDifference(SUT.sample.description, "main")
```
Sajjon, line 9: intentional (but not needed :P ) :D I wanted to limit number of hits on search for `develop`...

GhenadieVP (approved):

PR #123, opened 2024-05-02
CAP-36 New P2P links

danvleju-rdx:
This PR implements CAP-36 requirements by
- removing `p2pLinks` from the profile as they are moved to secure storage
- adding new `LinkConnectionQrData`, `P2PLink`, `RadixConnectPurpose`, and `InvalidRadixConnectPurpose` models to reflect the QR the wallet users scan to connect a link and the information required to establish the connection
- adding the `check_if_profile_json_contains_legacy_p2p_links` function which checks if a profile JSON contains `p2pLinks` without decoding the whole profile to be able to tell the wallet users to re-connect their links
- making `Ed25519Signature`, `Ed25519PublicKey`, `DisplayName`, `RadixConnectPassword`, `P2PLink`, `RadixConnectPurpose`, `LinkConnectionQrData` models `Codable`, to be able to encode/decode when storing/retrieving from/to secure storage


apple/Tests/TestCases/Profile/ProfileTests.swift
```diff
@@ -5,109 +5,119 @@ import SargonUniFFI
 import XCTest
 
 final class ProfileTests: Test<Profile> {
-	
-	func test_description_and_debug() {
-		XCTAssertGreaterThan(SUT.sample.debugDescription, SUT.sample.description)
-	}
+    
+    func test_description_and_debug() {
+        XCTAssertGreaterThan(SUT.sample.debugDescription, SUT.sample.description)
```
CyonAlexRDX, line 10: not your code, mine, but I think I meant to have `XCTAssertGreaterThan` be using `debugDescription.count` and `description.count`. Because it makes very little sense to compare the strings.... maybe just change to `XCTNotEqual`?

apple/Tests/TestCases/Profile/ProfileTests.swift
```diff
@@ -5,109 +5,119 @@ import SargonUniFFI
...
+    }
+    
+    func test_profile_description_equals() throws {
+        XCTAssertNoDifference(SUT.sample.description, SUT.sample.description)
```
CyonAlexRDX, line 14: I think this is already tested in `Test` super class?

apple/Tests/TestCases/Profile/ProfileTests.swift
CyonAlexRDX: But I think this is already tested in `Test` super class?

CyonAlexRDX: @danvleju-rdx can you revert the formatting change, and just add your method? we should also merge @matiasbzurovski s Swift format PR...

fixtures/vector/only_plaintext_profile_snapshot_version_100.json
Sajjon, line 1: revert this please, this is a test that ensures that Radix Wallet 1.0.0 profiles are compatible with new ones. Better to adjust the test in `tests/vector` to check that JSON decoding works an that JSON encoding works, but skip asserting that the encoded JSON equals the one in this file.

Also keeping this as is, will allow you to write tests using `check_if_profile_json_contains_legacy_p2p_links` with **this** profile and asserting that `check_if_profile_json_contains_legacy_p2p_links` returns `true` right? Which you can also use not only in Rust, but also in Kotlin and Swift Sargon.

fixtures/vector/multi_profile_snapshots_test_version_100_patch_after_app_version_120.json
Sajjon, line 1: this must be reverted, else encryption will fail. see related comment: https://github.com/radixdlt/sargon/pull/123/files#r1587700061

src/radix_connect/p2p_links/radix_connect_purpose.rs
```diff
@@ -0,0 +1,136 @@
...
+    Copy,
+    Debug,
+    PartialEq,
+    Eq,
+    Hash,
+    PartialOrd,
+    Ord,
+    uniffi::Enum,
+)]
+pub enum RadixConnectPurpose {
```
GhenadieVP, line 16:
```
#[serde(rename_all = "lowercase")]
```
better instead of:
```
#[serde(rename = "general")]
```

src/radix_connect/p2p_links/radix_connect_purpose.rs
```diff
@@ -0,0 +1,136 @@
...
+        Self::new_from_json_string(s).map_err(|_| {
+            CommonError::InvalidRadixConnectPurpose {
+                bad_value: s.to_owned(),
+            }
+        })
+    }
+}
+
+impl RadixConnectPurpose {
+    pub fn from_str_default_value(s: &str) -> Self {
```
GhenadieVP, line 36: Maybe worth having:
```
impl Default for RadixConnectPurpose {
    fn default() -> Self {
        RadixConnectPurpose::Unknown
    }
}
```

and then I think you can do:
```
RadixConnectPurpose::from_str(s).unwrap_or_default()
```

src/radix_connect/p2p_links/radix_connect_purpose.rs
```diff
@@ -0,0 +1,136 @@
...
+
+#[uniffi::export]
+pub fn new_radix_connect_purpose_from_string(
+    string: String,
+) -> RadixConnectPurpose {
+    RadixConnectPurpose::from_str_default_value(&string)
+}
+
+#[uniffi::export]
+pub fn radix_connect_purpose_to_string(kind: RadixConnectPurpose) -> String {
```
GhenadieVP, line 63: Why is this API needed? I mean where does the Wallet need the String representation of the purpose?

sergiupuhalschi-rdx: You're right, it's not needed in the Wallet.

fixtures/vector/only_plaintext_profile_snapshot_version_100.json
sergiupuhalschi-rdx: Reverted the change and added the unit test you suggested in Rust, but I'm not sure about having it also in Kotlin and Swift Sargon. At least in Kotlin we would have to copy the JSON file to the test resources folder and load the file's contents (which slows down the tests workflow), not sure it's worth it.

src/core/types/signatures/ed25519_signature.rs
```diff
@@ -1,9 +1,13 @@
 use crate::prelude::*;
 
+json_data_convertible!(Ed25519Signature);
+
 /// Represents an ED25519 signature.
 #[derive(
     Clone,
     Copy,
+    Serialize,
```
Sajjon, line 9: I think this is incorrect, remove `Serialize` and `Deserialize` from `derive` and use only `json_data_convertible!(Ed25519Signature);`

apple/Sources/Sargon/Extensions/Methods/Crypto/Keys/Ed25519PublicKey+Wrap+Functions.swift
```diff
@@ -12,6 +12,14 @@ extension Ed25519PublicKey {
         // Rust: `new_ed25519_public_key_from_bytes`
         self = try newEd25519PublicKeyFromBytes(bytes: Data(bytes))
     }
+    
+    public init(jsonData: some DataProtocol) throws {
+        self = try newEd25519PublicKeyFromBytes(bytes: Data(jsonData))
```
Sajjon, line 17: @danvleju-rdx this is semantically incorrect and will break if we were to change the JSON representation of the keys. Instead you should do:

```rust
json_data_convertible!(Ed25519PublicKey);
```

in Rust, and then use the newly exported UniFFI methods which will be:
`newEd25519PublicKeyFromJsonBytes` (N.B. the `Json`.)
and 
`ed25519PublicKeyToJsonBytes`

apple/Sources/Sargon/Extensions/Methods/Crypto/Signatures/Ed25519Signature+Wrap+Functions.swift
```diff
@@ -9,6 +9,14 @@ extension Ed25519Signature {
 	public init(exactly exactly64Bytes: Exactly64Bytes) {
 		self = newEd25519SignatureFromExactly64Bytes(bytes: exactly64Bytes)
 	}
+    
+    public init(jsonData: some DataProtocol) throws {
+        self = try newEd25519SignatureFromBytes(bytes: Data(jsonData))
```
Sajjon, line 14: this is correct :)

src/profile/v100/entity/display_name.rs
```diff
@@ -1,5 +1,7 @@
 use crate::prelude::*;
 
+json_data_convertible!(DisplayName);
```
Sajjon, line 3: hmm this should be `json_string_convertible`

src/profile/v100/profile.rs
```diff
@@ -190,6 +190,21 @@ impl Profile {
...
 
+impl Profile {
+    pub fn check_if_profile_json_contains_legacy_p2p_links(
+        json: impl AsRef<[u8]>,
+    ) -> bool {
+        let json = json.as_ref();
+        let result =
+            serde_json::from_slice::<ProtoProfileMaybeWithLegacyP2PLinks>(json);
+
+        match result {
```
Sajjon, line 201: this can be written as:

```rust
impl Profile {
    pub fn check_if_profile_json_contains_legacy_p2p_links(
        json: impl AsRef<[u8]>,
    ) -> bool {
        let json = json.as_ref();
        serde_json::from_slice::<ProtoProfileMaybeWithLegacyP2PLinks>(json)
            .map_or_else(|_| false, |s| s.app_preferences.p2p_links.is_empty())
    }
}
```
which is a bit more clear IMO (more functional, which Sargon uses...)

src/core/types/signatures/ed25519_signature.rs
danvleju-rdx: I'm getting this error if I remove them:
> json_data_convertible!(Ed25519Signature);
the trait `for<'a> interaction_id::_::_serde::Deserialize<'a>` is not implemented for `ed25519_signature::Ed25519Signature`

src/profile/v100/entity/display_name.rs
danvleju-rdx: @CyonAlexRDX  Nice! I should use `json_string_convertible` on `radix_connect_purpose.rs` too, right?

src/profile/v100/profile.rs
sergiupuhalschi-rdx: Looks better now. I only had to add a negation here: `s.app_preferences.p2p_links.is_empty()`. Thanks.

src/core/types/signatures/ed25519_signature.rs
Sajjon: @danvleju-rdx you are absolutely right, sorry! I got confused by all my macros I've written :D this one uniffi exportes it only, thus requiring it to be marked serde... but I think you should use:

```
    SerializeDisplay,
    DeserializeFromStr,
```

instead, relying on `std::fmt::Display` and `FromStr` (via our `derive_more::Display` and `derive_more::FromStr`)

src/profile/v100/entity/display_name.rs
Sajjon: Yes it makes sense for that type too

src/profile/v100/profile.rs
Sajjon: oh right, NOT empty :)

apple/Tests/TestCases/RadixConnect/P2PLinks/RadixConnectPasswordTests.swift
```diff
@@ -0,0 +1,19 @@
...
+		let raw = "\"deaddeaddeaddeaddeaddeaddeaddeaddeaddeaddeaddeaddeaddeaddeaddead\"".data(using: .utf8)!
+		
+		// test decoding
+		let sut = try JSONDecoder().decode(SUT.self, from: raw)
+		XCTAssertEqual(sut, SUT.sample)
+		
+		// test encoding
+		let encoded = try JSONEncoder().encode(sut)
+		try XCTAssertEqual(JSONDecoder().decode(SUT.self, from: encoded), sut)
+	}
```
Sajjon, line 18: Apart from those JSON tests which tests against known/expected values, you can also use the:

Perhaps add this function in to bottom of `Test.swift` below `func doTestCodableRoundtrip` 

```swift
	func eachSampleCodableRoundtripTest(
		_ test: (SUT) throws -> Void
	) rethrows {
		try eachSample { try doTestCodableRoundtrip($0) }
	}
```

and you can then add a tiny test to this file `RadixConnectPasswordTests.swift` (and all other test files doing JSON test) ? which is:

```swift
/// Cyon: We might be able remove this function once we have converted to `swift-testing` which has much more 
/// powerful discovery than XCTest, and maybe `eachSampleCodableRoundtripTest` will be picked up as
/// a test directly.
func testJSONRoundtripAllSamples() throws {
    try eachSampleCodableRoundtripTest()
}
```

In general I forgot about the `eachSample` function which we wanna try use as much as possible (ping @matiasbzurovski and @GhenadieVP ).

If you find tests like this:

```swift
	func test_phrase_roundtrip() throws {
		func doTest(_ sut: SUT) throws {
			let string = sut.description
			try XCTAssertEqual(SUT(phrase: string), sut)
			try XCTAssertEqual(SUT(phrase: string, language: .english), sut)
			try XCTAssertEqual(SUT(phrase: string, language: nil), sut) // ok to skip language
		}
		try SUT.sampleValues.forEach(doTest)
	}
```

That SHOULD be converted to:

```swift
	func test_phrase_roundtrip() throws {
		try eachSample { sut in
			let string = sut.description
			try XCTAssertEqual(SUT(phrase: string), sut)
			try XCTAssertEqual(SUT(phrase: string, language: .english), sut)
			try XCTAssertEqual(SUT(phrase: string, language: nil), sut) // ok to skip language
		}
	}
```

And if you are working with a type which is `Codable` (`SargonStringCodable` or `SargonObjectCodable`) we should add add a:

```swift
/// Cyon: We might be able remove this function once we have converted to `swift-testing` which has much more 
/// powerful discovery than XCTest, and maybe `eachSampleCodableRoundtripTest` will be picked up as
/// a test directly.
func testJSONRoundtripAllSamples() throws {
    try eachSampleCodableRoundtripTest()
}
```

apple/Tests/TestCases/Profile/ProfileTests.swift
```diff
@@ -110,4 +110,33 @@ final class ProfileTests: Test<Profile> {
...
 	}()
+
+    func test_check_if_profile_json_contains_legacy_p2p_links_when_p2p_links_are_not_present() {
+        func doTest(_ sut: SUT) {
+            XCTAssertEqual(
+                SUT.checkIfProfileJsonContainsLegacyP2PLinks(contents: sut.jsonData()),
+                false
+            )
+        }
+        SUT.sampleValues.forEach(doTest)
```
Sajjon, line 121: See https://github.com/radixdlt/sargon/pull/123/files#r1596434869, prefer using `eachSample { sut in }` 

I was my mistake having forgot I wrote that.... we should remove all inner `func doTest`...

apple/Tests/TestCases/Profile/ProfileTests.swift
```diff
@@ -110,4 +110,33 @@ final class ProfileTests: Test<Profile> {
...
+    }
+
+	func test_check_if_profile_json_contains_legacy_p2p_links_when_p2p_links_are_present() {
+		XCTAssertEqual(
```
Sajjon, line 125: personally I prefer `let json = """` and then a much shorter `XCTAssertEqual`, I think it is much easier to read.

tests/vectors/main.rs
```diff
@@ -10,13 +10,11 @@ mod profile_snapshot_tests {
...
-                env!("FIXTURES_VECTOR"),
-                "only_plaintext_profile_snapshot_version_100.json"
-            )))
-            .expect("V100 Profile to deserialize");
-        assert_json_value_eq_after_roundtrip(&profile, json)
+        fixture_and_json::<Profile>(include_str!(concat!(
+            env!("FIXTURES_VECTOR"),
+            "only_plaintext_profile_snapshot_version_100.json"
+        )))
+        .expect("V100 Profile to deserialize");
```
Sajjon, line 17: maybe leave a comment as to why we can no longer roundtrip test, mentioning CAP36 ?

src/profile/v100/profile.rs
```diff
@@ -525,6 +535,77 @@ mod tests {
         );
     }
 
+    #[test]
+    fn check_if_profile_json_contains_legacy_p2p_links_when_p2p_links_are_present(
+    ) {
+        let value = r#"
```
Sajjon, line 541: maybe rename `let value` -> `let json` ?

src/profile/v100/profile.rs
```diff
@@ -525,6 +535,77 @@ mod tests {
...
+    fn check_if_profile_json_contains_legacy_p2p_links_when_empty_json() {
+        let value = r#"
+        {
+            
+          }
+        "#;
+        assert_eq!(
+            false,
+            SUT::check_if_profile_json_contains_legacy_p2p_links(
+                value.as_bytes()
```
Sajjon, line 571: I think this can be `SUT::check_if_profile_json_contains_legacy_p2p_links(BagOfBytes::new())`

src/profile/v100/profile.rs
```diff
@@ -525,6 +535,77 @@ mod tests {
...
+                {
+                  "connectionPassword": "babebabebabebabebabebabebabebabebabebabebabebabebabebabebabebabe",
+                  "displayName": "Brave on PC"
+                }
+              ]
+            }
+          }
+        "#;
+        assert_eq!(
+            true,
```
Sajjon, line 554: please flip the args, I have been using RHS as "expected value" throughout the whole of Rust Sargon.

Sajjon: for this and all your other `assert` ... :)

src/profile/v100/profile_uniffi_fn.rs
```diff
@@ -148,4 +155,42 @@ mod uniffi_tests {
             sut
         );
     }
+
+    #[test]
+    fn check_if_profile_json_contains_legacy_p2p_links_when_p2p_links_are_present(
+    ) {
+        let value = r#"
```
Sajjon, line 162: `let json`

src/profile/v100/profile_uniffi_fn.rs
```diff
@@ -148,4 +155,42 @@ mod uniffi_tests {
...
+    #[test]
+    fn check_if_profile_json_contains_legacy_p2p_links_when_empty_json() {
+        let value = r#"
+        {
+            
+          }
+        "#;
+        assert_eq!(
+            false,
+            check_if_profile_json_contains_legacy_p2p_links(BagOfBytes::from(
```
Sajjon, line 191: `BagOfBytes::new()` ?

src/profile/v100/proto_profile_maybe_with_legacy_p2p_links.rs
```diff
@@ -0,0 +1,124 @@
...
+#[derive(Deserialize, Serialize, Eq, PartialEq, Debug)]
+pub struct ProtoAppPreferencesMaybeWithLegacyP2PLinks {
+    #[serde(rename = "p2pLinks")]
+    pub p2p_links: Vec<ProtoDummyLinkMaybeWithLegacyP2PLinks>,
+}
+
+#[derive(Deserialize, Serialize, Eq, PartialEq, Debug)]
+pub struct ProtoDummyLinkMaybeWithLegacyP2PLinks {
+    #[serde(rename = "displayName")]
+    pub display_name: String,
```
Sajjon, line 18: ah, if you actually make use of `DisplayName` type here, you can use `DisplayName::sample()` in your:

```
HasSampleValues for ProtoDummyLinkMaybeWithLegacyP2PLinks
```
which will be shorter :)

src/profile/v100/proto_profile_maybe_with_legacy_p2p_links.rs
```diff
@@ -0,0 +1,124 @@
...
+        Self {
+            display_name: String::from("Sample Display Name Other"),
+        }
+    }
+}
+
+impl HasSampleValues for ProtoAppPreferencesMaybeWithLegacyP2PLinks {
+    fn sample() -> Self {
+        Self {
+            p2p_links: Vec::from([
```
Sajjon, line 38: prefer `vec![]`

src/profile/v100/proto_profile_maybe_with_legacy_p2p_links.rs
```diff
@@ -0,0 +1,124 @@
...
+            ]),
+        }
+    }
+}
+
+impl HasSampleValues for ProtoProfileMaybeWithLegacyP2PLinks {
+    fn sample() -> Self {
+        Self {
+            app_preferences: ProtoAppPreferencesMaybeWithLegacyP2PLinks::sample(
+            ),
```
Sajjon, line 57: strange newline in middle of parenthesis, strange that cargo fmt did not complain? or maybe IT produced this :P

sergiupuhalschi-rdx: cargo fmt produced it :(

src/radix_connect/p2p_links/link_connection_qr_data.rs
```diff
@@ -0,0 +1,173 @@
...
+            purpose,
+            password,
+            public_key,
+            signature,
+        }
+    }
+}
+
+#[uniffi::export]
+pub fn new_link_connection_qr_data_sample() -> LinkConnectionQRData {
```
Sajjon, line 49: ok so please split the UniFFI related functions and tests into a separate file, called `the_other_file_uniffi_fn` like we do for the rest of the files. See e.g. `sargon_build_information_uniffi_fn`

src/radix_connect/p2p_links/link_connection_qr_data.rs
```diff
@@ -0,0 +1,173 @@
...
+        assert_eq!(
+            LinkConnectionQRData::sample().password.hash().to_string(),
+            "9059d2ac799749e2f9f18541015197051ff9f803741d566744fe34ea004a5908"
+        );
+    }
+}
+
+#[cfg(test)]
+mod uniffi_tests {
+    use crate::prelude::*;
```
Sajjon, line 160: can be `use super::*:`

src/radix_connect/p2p_links/link_connection_qr_data.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    #[test]
+    fn sample_values() {
+        assert_eq!(
+            new_link_connection_qr_data_sample(),
+            LinkConnectionQRData::sample()
```
Sajjon, line 166: Prefer `SUT` and maybe you can use hash, like `sargon_build_information_uniffi_fn`:

```rust
    #[test]
    fn hash_of_samples() {
        assert_eq!(
            HashSet::<SUT>::from_iter([
                new_sargon_build_information_sample(),
                new_sargon_build_information_sample_other(),
                // duplicates should get removed
                new_sargon_build_information_sample(),
                new_sargon_build_information_sample_other(),
            ])
            .len(),
            2
        );
    }
```

src/radix_connect/p2p_links/p2p_link.rs
```diff
@@ -25,6 +27,10 @@ pub struct P2PLink {
     /// for the `ID`.
     pub connection_password: RadixConnectPassword,
 
+    pub connection_purpose: RadixConnectPurpose,
```
Sajjon, line 30: missing doc, at least a oneliner per field :)

Sajjon: e.g. WHO sets the purpose, was this set by DAPP when Sargon / Host wallet app received it. What is the .... purpose of the purpose :D i.e. how does wallet use it.

src/radix_connect/p2p_links/p2p_link.rs
```diff
@@ -25,6 +27,10 @@ pub struct P2PLink {
...
+
+    pub public_key: Ed25519PublicKey,
```
Sajjon, line 32: doc... whos public key is this? Is this the public key of the Keyexchanged key? etc

src/radix_connect/p2p_links/p2p_links.rs
```diff
@@ -33,7 +33,6 @@ mod tests {
         assert_ne!(P2PLinks::sample(), P2PLinks::sample_other());
     }
 
-    #[test]
```
Sajjon, line 36: intentional disabling of test? if so prefer `[ignore]`, maybe an accident? if it is broken, lets try to fix it.

src/radix_connect/p2p_links/radix_connect_purpose.rs
```diff
@@ -0,0 +1,151 @@
+use crate::prelude::*;
+
+json_string_convertible!(RadixConnectPurpose);
+
+#[derive(
```
Sajjon, line 5: missing doc

src/radix_connect/p2p_links/p2p_link.rs
```diff
@@ -1,5 +1,7 @@
 use crate::prelude::*;
 
+json_data_convertible!(P2PLink);
+
 /// A client the user have connected P2P with, typically a
 /// WebRTC connections with a DApp, but might be Android or iPhone
 /// client as well.
```
Sajjon, line 7: is this doc still accurate? maybe it can be updated a bit?

Sajjon: links to any Confluence Page with CAP36 would be good. See how I've done it e.g. here with links:

https://github.com/radixdlt/sargon/blob/39ad077a83b50b93b322a00319398d3a47835d80/src/profile/v100/address/account_address.rs#L4-L26

src/radix_connect/p2p_links/link_connection_qr_data.rs
```diff
@@ -0,0 +1,173 @@
...
+            RadixConnectPassword::sample_other(),
+            Ed25519PublicKey::sample_other(),
+            Exactly64Bytes::sample_other(),
+        )
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::prelude::*;
```
Sajjon, line 87: prefer `use super::*;`

Sajjon: prefer `type SUT =. ..`

src/profile/v100/proto_profile_maybe_with_legacy_p2p_links.rs
```diff
@@ -0,0 +1,124 @@
+use crate::prelude::*;
+
+#[derive(Deserialize, Serialize, Eq, PartialEq, Debug)]
```
Sajjon, line 3: one liner doc for all types would be good, something along the lines.

```
/// This proto-Profile model has one single purpose - facilitate the query:
/// "does this profile contain legacy links", in a type-safe manner, using 
/// the fact that the `p2pLinks` were only "3 levels" deep into Profile, 
/// we can thus try deserialize JSON into this proto-Profile and check if
/// the proto-p2plinks collection is empty or not.
```

CyonAlexRDX (changes requested):
Almost there !

fixtures/vector/only_plaintext_profile_snapshot_version_100.json
CyonAlexRDX: @sergiupuhalschi-rdx both Kotlin Sargon and Swift Sargon already access test vectors directly, without the need to copy, see:
https://github.com/radixdlt/sargon/blob/39ad077a83b50b93b322a00319398d3a47835d80/jvm/sargon-android/src/test/java/com/radixdlt/sargon/TransactionManifestTest.kt#L615

and in Swift:
gon/blob/main/apple/Tests/Utils/Test.swift#L19

src/profile/v100/proto_profile_maybe_with_legacy_p2p_links.rs
CyonAlexRDX: damn :D do a fmt ignore then :)

CyonAlexRDX: https://stackoverflow.com/a/67289474

apple/Tests/TestCases/RadixConnect/P2PLinks/RadixConnectPasswordTests.swift
danvleju-rdx: I've done changes for all files in [this](https://github.com/radixdlt/sargon/pull/123/commits/28d2b1842e7f02d778e1c03e466210ec89cbb8a0) commit

fixtures/vector/only_plaintext_profile_snapshot_version_100.json
sergiupuhalschi-rdx: Great, thank you!

apple/Tests/TestCases/RadixConnect/P2PLinks/RadixConnectPasswordTests.swift
CyonAlexRDX: awesome!

CyonAlexRDX (approved):
Awesome job @danvleju-rdx and @sergiupuhalschi-rdx congrats on a great PR!

CyonAlexRDX (approved):
Awesome job @danvleju-rdx and @sergiupuhalschi-rdx congrats on a great PR!

PR #126, opened 2024-05-07
Add SwiftFormat config and pre-commit

matiasbzurovski:
## Changelog
Adds the configuration for Swift formatting and the pre-commit hook.

## Notes
Format changes will be done in a separate PR, leaving this one with only the configuration changes.

## Rationale
The project already had an almost empty `.swift-format` file, that would have allowed us to use Apple's [swift-format](https://github.com/apple/swift-format), instead of the [SwiftFormat](https://github.com/nicklockwood/SwiftFormat) developed by [nicklockwood](https://github.com/nicklockwood/). 

Even though I mostly prefer going with native tools, the Apple one doesn't allow as many rules as the selected one. Also, given we are already using it for the iOS Wallet project, it makes sense to use the same tool to easily replicate the same coding standards and styles in both projects.


CyonAlexRDX:
@matiasbzurovski @GhenadieVP lets close this for now and re-introduce it once we have merge https://github.com/radixdlt/sargon/pull/176

PR #127, opened 2024-05-08
Collection types -> Array | Faster since no UniFFI roundtrip needed.

CyonAlexRDX:
# Context
I made some assumptions about UniFFi which turned out to be false - specifically: I assumed roundtrip of collections would be extremely fast, e.g. roundtrip of 1000 accounts. Why we hid the elements of `Accounts` (and all sibling types such as `Personas`, `FactorSources` etc) and had to use a UniFFI exported function to read them out. This incurs a 3N cost, where N is the number of accounts passed over the UniFFI boundary - which is somewhat expensive! This PR changes the `O(3N)` cost to `O(1)` ie constant time. Since the collection types are now plain Array/List in Swift / Kotlin. 

# Changes
We loose some of the non-empty check of `FactorSources` (and `SupportedCurves` in the Rust side, since the underlying data type does not enforce it. However, we validate FactorSources to not be empty in the constructor of Profile; and when the type itself (`OrderedMap<FactorSource>`) is JSON serialised, and deserialized and when it is passed to Rust, from FFI (from Swift side / Kotlin side). The same applies for SupportedCurves.

Since `OrderedMap<T>` gets translated into `Array<T>` in Swift (and similar in Kotlin) we loose both the non empty check on `FactorSources` and we lose the uniqueness check for **all** `OrderedMap` since they are just arrays... in PR https://github.com/radixdlt/babylon-wallet-ios/pull/1108 I've ""fixed"" this by converting to IdentifiedArrayOf eagerly, e.g. the clients (eg `AccountsClient`) they return `IdentifiedArrayOf<Account>` instead of `Array<Account>`. 

# Rust 
* Remove dependency on my own extern crate `IdentifiedVec`, replaced by...
* Add new type `OrderedMap`, which is backed by Rust community de facto standard for ordered unique collections: `IndexMap` - where element in `OrderedMap` is required to impl ...
* Add new trait `Identifiable` (was previously declared by extern crate `IdentifiedVec`) 
* Written custom UniFFI conversion for `OrderedMap` where we ensure Kotlin and Swift **will CRASH if those collections contains duplicates** (when "sent to Rust side over UniFFI" - which we do when we "save profile" in wallets (JSON encoding happens rust sude...)). Also **will CRASH** if `FactorSource`s collection (Swift: `Array<FactorSource>` and corresponiding in Kotlin)  is **empty**, same goes for `Slip10Curve`s collection (`supportedCurves` in Factor Source common).
* Removed our wrapper around `IdentifiedVec` called `IdentifeidVecVia`
* Removed our macro to defined the identefied elements collections: `decl_can_be_empty_identified_array_of` and its sibling `never_empty`
* Add macro `decl_ordered_map` which **just** declares a typealias for the `OrderedMap<T>` and UniFFI exports sample and sample_other values (also with unit tests)
* Changed `persona_data_identified_collection` types to use `OrderedMap` instead of `IdentifiedVecVia`

# Swift
Removed the protocols `BaseIdentifiedCollection`, `CanBeEmtpyCollection` and `NeverEmptyCollection` and translated tests...


src/profile/logic/profile_next_derivation.rs
```diff
@@ -114,7 +113,7 @@ impl Profile {
         header.content_hint = networks.content_hint();
         Self::with(
             header,
-            FactorSources::from_iter([FactorSource::sample_ledger()]).unwrap(),
+            FactorSources::from_iter([FactorSource::sample_ledger()]), //.unwrap(),
```
CyonAlexRDX, line 116: FIXME: remove the dead code here and elsewhere...

src/profile/v100/factors/factor_source_crypto_parameters.rs
```diff
@@ -115,10 +114,9 @@ impl HasSampleValues for SupportedCurves {
             SLIP10Curve::Curve25519,
             SLIP10Curve::Secp256k1,
         ])
-        .unwrap()
+        // .unwrap()
```
CyonAlexRDX, line 117: FIXME: dead code

src/profile/v100/factors/factor_sources/factor_sources.rs
```diff
@@ -67,16 +66,16 @@ mod tests {
...
-        let fs = FactorSource::sample();
-        let sut = SUT::from_iter([fs.clone()]).unwrap();
-        assert_eq!(sut.len(), 1);
-        assert_eq!(
-            new_factor_sources_removed_by_id(&fs.id(), &sut),
-            Err(CommonError::FactorSourcesMustNotBeEmpty)
-        );
-    }
+    // #[test]
+    // fn remove_returns_err_if_empty() {
```
CyonAlexRDX, line 70: FIXME: remove dead code

src/profile/v100/factors/factor_sources/factor_sources.rs
```diff
@@ -50,34 +51,47 @@ mod tests {
...
+        assert!(!SUT::sample().is_empty());
+    }
+
     #[test]
     fn inequality() {
         assert_ne!(SUT::sample(), SUT::sample_other());
     }
 
+    #[test]
+    fn manual_perform_uniffi_if_factor_sources_empty() {
```
matiasbzurovski, line 68: You mean something like `manual_error_in_uniffi_if_factor_sources_empty` ?

matiasbzurovski (approved):

micbakos-rdx (approved):

sergiupuhalschi-rdx (approved):

PR #129, opened 2024-05-08
Implement kotlin `IdentifiableArray` impl

micbakos-rdx:
* Simple `IdentifiableArray` implementation 
* Collections like `Accounts`, `Personas` delegate to this array.
* Url is now being represented with okHttp's [`HttpUrl`](https://square.github.io/okhttp/3.x/okhttp/okhttp3/HttpUrl.html) in Kotlin, which seems to have the same behaviour on url equality. Also okhttp is already a dependency for our network requests so no more dependency was necessary.


CyonAlexRDX (approved):
Nice LGTM !

jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/ProfileEntity.kt
```diff
@@ -3,7 +3,7 @@ package com.radixdlt.sargon.extensions
 import com.radixdlt.sargon.Account
 import com.radixdlt.sargon.AccountAddress
 import com.radixdlt.sargon.AddressOfAccountOrPersona
-import com.radixdlt.sargon.EntityFlags
+//import com.radixdlt.sargon.EntityFlags
```
sergiupuhalschi-rdx, line 6: Seems like this line is no longer needed

sergiupuhalschi-rdx (approved):

PR #132, opened 2024-05-09
Fix equality and hashcode in identifiable collections

micbakos-rdx:
* `equals`, `hashCode` and `toString` need to be exposed in interface so the actual class can use those methods when delegating to `array` instead of the default implementations on `Any`
* Individual tests for each collection
* Added `operator fun contains` so we can use `in` operator in project
* Inner `array` should become property so equality could work with delegation. Not sure if it was possible without making `array` property of the class.
* There is a false positive in code coverage regarding `javaClass` check, so I had to ignore the whole method, even though it is tested.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2529,7 +2529,7 @@ dependencies = [
 
 [[package]]
 name = "sargon"
-version = "0.7.9"
+version = "0.7.10"
 dependencies = [
  "actix-rt",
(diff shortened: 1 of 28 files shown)


CyonAlexRDX (approved):
LGTM!

sergiupuhalschi-rdx (approved):

PR #134, opened 2024-05-10
Decimal192 string parsing and representing in text fields

micbakos-rdx:
* Android keyboards may allow more characters than decimals to be typed regardless the current locale
* We need to act upon the text input and edit the string accordingly.
* Only digits and decimal separator is allowed.
* Any text is stripped from non-allowed characters with a regex and passes the result to sargon along with the locale's decimal separator.
* The same applies when the converted decimal192 is converted back to string for text inputs. Again only digits and decimal separator are allowed.
* Space is represented as 0. (In case user pastes `" "`)


jakub-rdx (approved):

CyonAlexRDX (approved):
LGTM

PR #136, opened 2024-05-13
Add P2PLink and LinkConnectionQrData extensions

sergiupuhalschi-rdx:


jvm/sargon-android/src/main/java/com/radixdlt/sargon/extensions/P2pLink.kt
danvleju-rdx, line 1: This one should be deleted I think

danvleju-rdx (approved):

CyonAlexRDX (approved):
LGTM

PR #137, opened 2024-05-13
Add 'regenerate' argument to build and test gradle tasks

sergiupuhalschi-rdx:
This is an attempt to improve consecutive builds time by adding `regenerate` boolean argument to build gradle tasks. If you don't pass any arg to the gradle task, it will follow the same steps as before (generating uniffi bindings which takes long and re-run even if there are no changes in rust), but if passing `regenerate=false` to gradle task, it skips this step entirely and rely upon previously generated bindings:

`./jvm/gradlew -p jvm/sargon-android testDebugUnitTest -Pregenerate=false`

### DIFF
--- a/jvm/sargon-android/build.gradle.kts
+++ b/jvm/sargon-android/build.gradle.kts
@@ -157,6 +157,12 @@ android.libraryVariants.all {
         "generate${buildTypeUpper}UniFFIBindings",
         Exec::class
     ) {
+        val rebuild = properties["rebuild"] ?: true
+        
+        onlyIf {
+            rebuild == true || !File("${buildDir}/generated/src/${buildType}/java").exists()
+        }
+
         group = BasePlugin.BUILD_GROUP
 
         workingDir = rootDir.parentFile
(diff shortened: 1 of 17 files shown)


CyonAlexRDX (approved):
Nice, so additive change, leaving same behaviour as before

CyonAlexRDX (commented):
Bump version - always :)

jvm/sargon-android/build.gradle.kts
```diff
@@ -157,6 +157,12 @@ android.libraryVariants.all {
         "generate${buildTypeUpper}UniFFIBindings",
         Exec::class
     ) {
+        val rebuild = properties["rebuild"] ?: true
```
micbakos-rdx, line 160: If we go with this way, I would rename it to `regenerate` and not `rebuild` since this task just generates code and does not build anything.

But regardless this is a param for us to use when we know that the build didn't produce any changes. As a gradle user this is not the correct way to do it. 

Do we really need this for now? If so yes, we can "rely" on it for now but such mechanism is not reliable. We can make gradle do this automatically with `upToDateWhen`.

sergiupuhalschi-rdx: I renamed the param to `regenerate`. We should spend more time on this to have a proper solution in place. This quick solution on the other hand will help us shorten the time spent on consecutive builds.

sergiupuhalschi-rdx:
Closed in favor of [#158](https://github.com/radixdlt/sargon/compare/feature/gradle-skip-regenerating-bindings?expand=1)

PR #140, opened 2024-05-13
Make TextInputDecimal decimal property nullable

jakub-rdx:
## Description
- wrap `decimal` property of `TextInputDecimal` in runCatching {} to catch decimal parse error
- add test examples for this particular case


CyonAlexRDX (approved):
LGTM

micbakos-rdx (approved):
Nice

PR #141, opened 2024-05-15
Update the Scrypto Dependency for Bottlenose

0xOmarA:
No change in functionality, just updating the version of Scrypto and RET for bottlenose.


CyonAlexRDX (approved):
Thank you! 🤩

PR #142, opened 2024-05-16
[CAP-36] Add encrypted profile legacy P2P links check

danvleju-rdx:
Adds a new function that checks if an encrypted profile contains legacy P2P links.

### DIFF
--- a/apple/Tests/TestCases/Profile/ProfileTests.swift
+++ b/apple/Tests/TestCases/Profile/ProfileTests.swift
@@ -113,33 +113,29 @@ final class ProfileTests: Test<Profile> {
 
     func test_check_if_profile_json_contains_legacy_p2p_links_when_p2p_links_are_not_present() {
 		eachSample { sut in
-			XCTAssertEqual(
-				SUT.checkIfProfileJsonContainsLegacyP2PLinks(contents: sut.jsonData()),
-				false
+			XCTAssertFalse(
+				SUT.checkIfProfileJsonContainsLegacyP2PLinks(contents: sut.jsonData())
 			)
 		}
     }
 
 	func test_check_if_profile_json_contains_legacy_p2p_links_when_p2p_links_are_present() throws {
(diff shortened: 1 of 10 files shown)


apple/Tests/TestCases/Profile/ProfileTests.swift
```diff
@@ -127,4 +127,19 @@ final class ProfileTests: Test<Profile> {
 			true
 		)
 	}
+
+	func test_check_if_encrypted_profile_json_contains_legacy_p2p_links_when_empty_json() {
+		XCTAssertEqual(
```
Sajjon, line 132: nit: I prefer `XCTAssert` only when using bool (that is expected to be true)

src/profile/v100/profile.rs
```diff
@@ -135,7 +135,7 @@ impl Profile {
 			error!("Failed to deserialize JSON as EncryptedProfileSnapshot, error: {:?}", e);
             CommonError::FailedToDeserializeJSONToValue {
                 json_byte_count: json.len() as u64,
-                type_name: "EncryptedProfileSnapshot".to_owned(),
+                type_name: type_name::<EncryptedProfileSnapshot>(),
```
Sajjon, line 138: 👍

src/profile/v100/profile_uniffi_fn.rs
```diff
@@ -186,4 +197,32 @@ mod uniffi_tests {
...
         );
     }
+
+    #[test]
+    fn check_if_encrypted_profile_json_contains_legacy_p2p_links_when_p2p_links_are_present(
+    ) {
+        let json =
+            serde_json::to_vec(&EncryptedProfileSnapshot::sample()).unwrap();
+        let password = "babylon";
+        assert_eq!(
```
Sajjon, line 207: if you run `cargo clippy --fix --allow-staged` you will notice it will change this to `assert!(..)` only

CyonAlexRDX (approved):
LGTM

PR #144, opened 2024-05-23
Use String for JSON coding of Profile (10x speedup)

CyonAlexRDX:
# JSON String for Profile
I forgot that we perform 2's complement on **every single byte, byte by byte** in BagOfBytes, which is very costly for large data - such as Profile. Therefore, we SHOULD NOT use JSON Data for Profile API's - but rather JSON String. 

We get an immediate 10x speedup in Swift by doing so.

# Performance gain - 10x speedup
The `vector` is [the huge profile with 1000 accounts](https://github.com/radixdlt/sargon/blob/main/fixtures/vector/huge_profile_1000_accounts.json).

```swift
// Macbook Pro M2: 0.64 seconds
func test_performance_json_encoding_data() throws {
	let (sut, _) = vector
	measure {
		let _ = sut.jsonData()
	}
}

// Macbook Pro M2: 0.06 (10x speedup vs BagOfBytes)
func test_performance_json_encoding_string() throws {
	let (sut, _) = vector
	measure {
		let _ = sut.jsonString(prettyPrint: false)
	}
}

// Macbook Pro M2: 0.26 seconds
func test_performance_json_decoding_data() throws {
	let (_, jsonData) = vector
	measure {
		let _ = try! SUT(jsonData: jsonData)
	}
}

// Macbook Pro M2: 0.1 (2.5x speedup vs BagOfBytes)
func test_performance_json_decoding_string() throws {
	let (_, _jsonData) = vector
	let jsonString = String(data: _jsonData, encoding: .utf8)!
	measure {
		let _ = try! SUT(jsonString: jsonString)
	}
}
```

# Base64 - not worth it
I also tried Base64, it does in fact give a 11x speedup instead of 10x speed up for encoding, i.e. a 10% speedup vs Json String - however, for decoding it was 50% slower than using Json String. 

So I figured, maybe Base64 performance increase will be much larger for a 1MB attached blob. So I temporarily added a `blob: Vec<u8>` (avoided `BagOfBytes` for max performance) to Profile root and then generated test vectors (base64 string) for a sample Profile with such a blob of 1MB, and measured performance of that. Then YES using Bas64 is faster for both encoding **and decoding**, but decoding is only 10% faster, so still not worth it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2586,7 +2586,7 @@ dependencies = [
 
 [[package]]
 name = "sargon"
-version = "0.7.18"
+version = "0.7.20"
 dependencies = [
  "actix-rt",
  "aes-gcm",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sargon"
-version = "0.7.18"
+version = "0.7.20"
 edition = "2021"
 build = "build.rs"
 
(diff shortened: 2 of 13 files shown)


danvleju-rdx (approved):
LGTM! Cool test strings :D

sergiupuhalschi-rdx (approved):
👏

PR #145, opened 2024-05-23
Fix android mnemonic  serialization

jakub-rdx:
It looks like there was a typo in current sargon jvm for backwards compatible android mnemonic.


sergiupuhalschi-rdx (approved):
🏅

CyonAlexRDX (approved):
LGTM! Well spotted, glad we caught this 😅😅😅😅

PR #147, opened 2024-05-24
Throw err for invalid manifest instead of crash in summary

CyonAlexRDX:
Bump `RET` to a version where the non sensical manifest does not crash when calling `summary` on it.

```text
CALL_METHOD
  Address("account_tdx_2_1c90qdw5e3t3tjyd8axt3zg9zezhhhymt2mr8y4l0k2285mfwhczhdt")
  "withdraw"
  Address("account_tdx_2_12y0errvzu4caktegxc5v0ug93u5yat9d90k4zdqkcy6n55pt7wlgmq")
  Decimal("0.01")
;
TAKE_FROM_WORKTOP
  Address("resource_tdx_2_1thqru9whuem5sjltshg4q7vj3e22xfh27u6s9xvwecz4fallqjny90")
  Decimal("0.01")
  Bucket("burn")
;
BURN_RESOURCE
  Bucket("burn")
;
```

The above nonsensical manifest crashed when we called `summary` on it, but not now after bumping RET.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2163,7 +2163,7 @@ dependencies = [
 [[package]]
 name = "radix-engine-toolkit"
 version = "2.1.0-dev1"
-source = "git+https://github.com/radixdlt/radix-engine-toolkit?rev=429dd3de58c0860ddbb8a83e9b57e65a1c0765d1#429dd3de58c0860ddbb8a83e9b57e65a1c0765d1"
+source = "git+https://github.com/radixdlt/radix-engine-toolkit?rev=c699783b487e9c1eb78847f9093d918d3dbd5e39#c699783b487e9c1eb78847f9093d918d3dbd5e39"
 dependencies = [
  "bech32",
  "cargo_toml 0.15.3 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


src/wrapped_radix_engine_toolkit/low_level/transaction_manifest/instructions/instructions.rs
```diff
@@ -89,6 +89,17 @@ impl Instructions {
         )
         .map_err(|e| extract_error_from_error(e, network_id))
         .and_then(|manifest| {
+            // This catches and fixes "broken invariants"
+            std::panic::catch_unwind(|| {
```
CyonAlexRDX, line 93: @iamyulong @0xOmarA   is this the correct thing to do? Or is this example manifest "valid" but `summary` is broken? hmm I would appreciate any input. We are about to do a release soon, so maybe we wont have time to bump RET / Scrypto dependencies before our next release. cc @GhenadieVP

CyonAlexRDX: reverted addition of `catch_unwind` since newer version of RET does not crash,

0xOmarA (approved):

PR #148, opened 2024-05-27
Add percentage to TransactionGuarantee 

matiasbzurovski:
## Rationale
The `TransactionGuarantee` holds the `amount` that we guarantee as minimum for a transaction. Such value is calculated from the `total` * `percentage`, where the latter comes from user default preferences or a specific value selected for this transaction. 

After a `TransactionGuarantee` has been saved, user could want to edit the percentage selected. Up until now, this value was shown by doing the reverse calculation (`percentage` = `amount` / `total`). However, this would lead to different values than the original user input on certain cases. 

For example, let's say we have a transaction of `20 CANDY`, an asset that has 0 divisibility. If the user sets a percentage of `92%`, the `amount` will be set to `18` (since 18.4 isn't possible for this asset). However, if the user opens the view to edit this guarantee, it would indicate that the percentage is 90% (`18/20 = 0.9`).

The solution then is to keep the user's preference value, so that they get a better UX and doesn't look as we are modifying their selection.

## Related issues
[ABW-3195](https://radixdlt.atlassian.net/browse/ABW-3195)

[ABW-3195]: https://radixdlt.atlassian.net/browse/ABW-3195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2586,7 +2586,7 @@ dependencies = [
 
 [[package]]
 name = "sargon"
-version = "1.0.2"
+version = "1.0.3"
 dependencies = [
  "actix-rt",
  "aes-gcm",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sargon"
-version = "1.0.2"
+version = "1.0.3"
 edition = "2021"
 build = "build.rs"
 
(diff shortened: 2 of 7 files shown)


CyonAlexRDX (approved):
LGTM

PR #149, opened 2024-05-27
fix third party deposit manifest

CyonAlexRDX:
Fix bug present in Production, user is unable to update Third Party Deposit settings by removing `depositor`. 

This fixes that.


jakub-rdx (approved):

jakub-rdx:
@CyonAlexRDX sneaky one

PR #152, opened 2024-05-31
just bump to 1.0.6

CyonAlexRDX:
something got messed up with sargon release versions. main branch has 1.0.4 version in toml, but last release tag is 1.0.5 and last merge to main did not result in an iOS release....


micbakos-rdx (approved):
That was my exact problem the previous time

PR #158, opened 2024-06-05
Add 'regenerate' argument to build and test gradle tasks 

sergiupuhalschi-rdx:
This is an attempt to improve consecutive builds time by adding `regenerate` boolean argument to build gradle tasks. If you don't pass any arg to the gradle task, it will follow the same steps as before (generating uniffi bindings which takes long and re-run even if there are no changes in rust), but if passing `regenerate=false` to gradle task, it skips this step entirely and rely upon previously generated bindings:

`./jvm/gradlew -p jvm/sargon-android testDebugUnitTest -Pregenerate=false`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2586,7 +2586,7 @@ dependencies = [
 
 [[package]]
 name = "sargon"
-version = "1.0.8"
+version = "1.0.9"
 dependencies = [
  "actix-rt",
  "aes-gcm",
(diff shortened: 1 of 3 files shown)


micbakos-rdx:
Related discussion in this PR https://github.com/radixdlt/sargon/pull/137

micbakos-rdx (approved):

PR #159, opened 2024-06-05
[ABW-3208] Callback_path fetching

jakub-rdx:
> [!NOTE]
> Target branch is not `main`

## Description
- add client for fetching well known file together with models
- add callback_path resolving, which is for now just temporary string


src/radix_connect/well_known_client/client.rs
```diff
@@ -0,0 +1,141 @@
...
+    async fn test_get_well_known_file_failure_response() {
+        let client = WellKnownClient::new_always_failing();
+
+        let req =
+            client.get_well_known_file(Url::from_str(TEST_ORIGIN).unwrap());
+        let result = timeout(MAX, req).await.unwrap();
+        assert!(result.is_err());
+    }
+
+    // #[actix_rt::test]
```
CyonAlexRDX, line 123: remove dead code?

src/radix_connect/well_known_client/client.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    fn new_succeeding_http_client(response: &DappDefinitions) -> Self {
+        Self::new_with_network_antenna(Arc::new(MockAntenna::with_response(
+            response,
+        )))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
CyonAlexRDX, line 56: please annotate tests with `ARRANGE` , `ACT`, `ASSERT`

src/radix_connect/well_known_client/client.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    #[actix_rt::test]
+    async fn test_get_well_known_file_correct_request_made() {
+        let mock_antenna_with_spy =
+            MockAntenna::with_spy(200, vec![], |request| {
+                let request = NetworkRequest::get_well_known_request(
+                    Url::from_str(TEST_ORIGIN).unwrap(),
+                );
+
+                let expected_request = NetworkRequest {
```
CyonAlexRDX, line 80: prefer using a (or if lacking, writing a) constructor for `NetworkRequest`.

CyonAlexRDX: and if you create expected `NetworkRequest` you can use `assert_eq!(request, expected)`;

src/radix_connect/well_known_client/client.rs
```diff
@@ -0,0 +1,141 @@
...
+
+impl NetworkRequest {
+    fn get_well_known_request(url: Url) -> Self {
+        let well_known_url = url.join(SUFFIX_WELL_KNOWN_FILE).unwrap();
+        NetworkRequest::new_get(well_known_url)
+    }
+}
+
+impl WellKnownClient {
+    pub async fn get_well_known_file(
```
CyonAlexRDX, line 29: missing documentation

src/radix_connect/well_known_client/client.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    pub fn new_with_network_antenna(
+        network_antenna: Arc<dyn NetworkAntenna>,
+    ) -> Self {
+        Self::new(HttpClient::new(network_antenna))
+    }
+}
+
+impl NetworkRequest {
+    fn get_well_known_request(url: Url) -> Self {
```
CyonAlexRDX, line 22: lacking documentation. the `get...request` makes it sound like this is a getter, but this is a constructor! I would drop the `_request, it is being called on `NetworkRequest, so seems superflous.

src/radix_connect/well_known_client/client.rs
```diff
@@ -0,0 +1,141 @@
+use crate::prelude::*;
+
+const SUFFIX_WELL_KNOWN_FILE: &str = ".well-known/dapp.json";
```
CyonAlexRDX, line 3: add a trivial test:

`assert_eq!(SUFFIX_WELL_KNOWN_FILE, ".well-known/dapp.json")` which will capture if we accidentally change this.

Such tests ARE valuable, take 10 seconds to write but will forever protect us from accidentally changing this value!

src/radix_connect/mobile/session/session.rs
```diff
@@ -38,6 +41,7 @@ impl HasSampleValues for Session {
             SessionID::sample_other(),
             SessionOrigin::sample_other(),
             Exactly32Bytes::sample_other(),
+            String::from("callback_path_other"),
```
CyonAlexRDX, line 44: prefer: `"callback_path_other".to_owned()`

src/radix_connect/mobile/session/session.rs
```diff
@@ -30,6 +32,7 @@ impl HasSampleValues for Session {
             SessionID::sample(),
             SessionOrigin::sample(),
             Exactly32Bytes::sample(),
+            String::from("callback_path"),
```
CyonAlexRDX, line 35: prefer: `"callback_path".to_owned()`

src/radix_connect/mobile/session/session.rs
```diff
@@ -7,19 +7,21 @@ pub struct Session {
     pub id: SessionID,
     pub origin: SessionOrigin,
     pub encryption_key: Exactly32Bytes,
-    // TBA: possibly will contain also the callback path
+    pub callback_path: String, // TBA: possibly will contain also the callback path
```
CyonAlexRDX, line 10: hmm what is this? a URL? Why it this a `String`? We should be **very vary about `String`**, we should always try to use the most restrictive and desecriptive datatype for our data, which allows us developer to protect against bad values :)

src/radix_connect/mobile/client.rs
```diff
@@ -52,6 +58,22 @@ impl RadixConnectMobile {
...
+        if !dev_mode && dapp_definitions.callback_path.is_none() {
+            return Err(
+                CommonError::RadixConnectMobileDappCallbackPathNotFound {
+                    origin: request.origin,
+                },
+            );
+        }
+        let callback_path = dapp_definitions
+            .callback_path
+            .unwrap_or(String::from("default_callback_path"));
```
CyonAlexRDX, line 75: This is a strong argument against using `String`! `.unwrap_or(String::from("default_callback_path"));` is clearly placeholder values. If you create a `newtype` for this:

```rust
#[derive(Clone, PartialEq, Eq, derive_more::FromStr, derive_more::Display, DeserializeFromStr, SerializeDisplay)]
pub struct RCMCallbackPath(pub(crate) String);
```

(if needed you to make it UniFFI compatible take a look at `BIP39Passphrase`).

And then you impl `Default` for it:

```rust
impl Default for RCMCallbackPath {
    fn default() -> Self {
        err!("Using PLACEHOLDER Default RCM CallbackPath");
        Self("default_callback_path".to_owned())
    }
}
```

and then you can use:

```rust
 let callback_path = dapp_definitions
            .callback_path
            . unwrap_or_default();
```

CyonAlexRDX: and probably a missing `//FIXME replace this placeholder default value`....

src/radix_connect/mobile/session/session.rs
CyonAlexRDX: I like and try to follow Object Calisthenics by William Durand, it has been a source of inspiration for me for 15 years, great article and great "heuristics":

https://williamdurand.fr/2013/06/03/object-calisthenics/#3-wrap-all-primitives-and-strings

This is rule 3: *Wrap all primitives and strings*

src/radix_connect/mobile/client.rs
jakub-rdx: so `callback_path` in both `Session` struct and `DappDefinition` which is purely model of json response should use `RCMCallbackPath`?

src/radix_connect/mobile/client.rs
```diff
@@ -52,6 +58,22 @@ impl RadixConnectMobile {
...
+        if !dev_mode && dapp_definitions.callback_path.is_none() {
+            return Err(
+                CommonError::RadixConnectMobileDappCallbackPathNotFound {
+                    origin: request.origin,
+                },
+            );
+        }
+        let callback_path = dapp_definitions
+            .callback_path
+            .map_or(RCMCallbackPath::default(), RCMCallbackPath::new);
```
Sajjon, line 75: I prefer:

```rust
 let callback_path = dapp_definitions
            .callback_path
            .map(RCMCallbackPath::new)
            .unwrap_or_default();
```

src/radix_connect/mobile/session/callback_path.rs
```diff
@@ -0,0 +1,54 @@
+use crate::prelude::*;
+
+#[derive(
```
Sajjon, line 3: Missing documentation

src/radix_connect/mobile/session/callback_path.rs
```diff
@@ -0,0 +1,54 @@
...
+    #[test]
+    fn equality() {
+        assert_eq!(SUT::sample(), SUT::sample());
+        assert_eq!(SUT::sample_other(), SUT::sample_other());
+    }
+
+    #[test]
+    fn inequality() {
+        assert_ne!(SUT::sample(), SUT::sample_other());
+    }
```
Sajjon, line 53: add JSON tests

src/radix_connect/well_known_client/client.rs
```diff
@@ -0,0 +1,116 @@
...
+        // ACT
+        let request = NetworkRequest::get_well_known(url.clone());
+        // ASSERT
+        assert_eq!(request.url, url.join(SUFFIX_WELL_KNOWN_FILE).unwrap());
+        assert_eq!(request.method, NetworkMethod::Get);
+    }
+
+    #[actix_rt::test]
+    async fn test_get_well_known_file_correct_request_made() {
+        // ARRANGE
```
Sajjon, line 79: nice! just we a few `ARRANGE`, `ACT`, `ASSERT` this tests became much more clear! especially good in this case when the `ASSERT` is spelled in lines above the `ACT`, it helps me a lot, and others too I'm sure!

src/radix_connect/well_known_client/models/dapp_definitions.rs
```diff
@@ -0,0 +1,55 @@
+use crate::prelude::*;
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct DappDefinitions {
+    #[serde(rename = "dApps")]
+    pub dapp_definitions: Vec<DappDefinition>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub callback_path: Option<String>,
```
Sajjon, line 9: Should be `RCMCallbackPath` type? (as in  `Option<RCMCallbackPath>`)?

jakub-rdx: ok, how I make serde be able to serialize/deserialize RCMCallbackPath. Do I need additional setup?

src/radix_connect/well_known_client/models/dapp_definitions.rs
```diff
@@ -0,0 +1,55 @@
+use crate::prelude::*;
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct DappDefinitions {
```
Sajjon, line 5: `DappDefinitions` lacks documentation!

It also lacks a constructor! Write a constructor which uses `impl ItoIterator<Item = DappDefinition>` and `impl Into<Option<RCMCallbackPath>>`

src/radix_connect/mobile/session/callback_path.rs
jakub-rdx: JSON tests are performed in the dapp_definitions for a whole model, is there a need for them here too?

src/radix_connect/well_known_client/models/dapp_definitions.rs
```diff
@@ -0,0 +1,55 @@
+use crate::prelude::*;
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
```
Sajjon, line 3: Lacks `Clone`, **always** derive `Clone` and probably we want `PartialEq, Eq` too!

src/radix_connect/well_known_client/models/dapp_definition.rs
```diff
@@ -0,0 +1,40 @@
+use crate::prelude::*;
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
```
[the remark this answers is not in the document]
Sajjon: lacks documentation

src/radix_connect/well_known_client/models/dapp_definition.rs
```diff
@@ -0,0 +1,40 @@
...
+pub struct DappDefinition {
+    #[serde(rename = "dAppDefinitionAddress")]
+    pub dapp_definition_address: AccountAddress,
+}
+
+impl HasSampleValues for DappDefinition {
```
Sajjon, line 10: lacks constructor, please add one, and use the constructor in samples!

src/radix_connect/well_known_client/client.rs
```diff
@@ -0,0 +1,116 @@
+use crate::prelude::*;
+
+const SUFFIX_WELL_KNOWN_FILE: &str = ".well-known/dapp.json";
+
+pub struct WellKnownClient {
```
Sajjon, line 5: lacks documentation!

src/http_client/network_antenna/network_request.rs
```diff
@@ -19,6 +19,15 @@ impl NetworkRequest {
         }
     }
 
+    pub fn new_get(url: Url) -> Self {
```
Sajjon, line 22: this type would benefit from a constructor, and change all other to delegate to that.

```rust
impl NetworkRequest {
    pub fn new(url: impl Into<Url>, method: impl Into<NetworkMethod>, headers: impl Into<HashMap<String, String>>, body: impl Into<BagOfBytes>) -> Self {
        Self {
           ....
        }
    }
}
```

And then:
`new_get` becomes:
```rust
pub fn new_get(url: Url) -> Self {
        Self::new(url, NetworkMethod::Get, [], [])
}
```
(also rewrite `new_post`!!)
🎉

jakub-rdx: this sample seem to not work. I'm not sure how I would convert from [] to BagOfBytes

src/radix_connect/mobile/client.rs
```diff
@@ -70,9 +70,7 @@ impl RadixConnectMobile {
                 },
             );
         }
-        let callback_path = dapp_definitions
-            .callback_path
-            .map_or(RCMCallbackPath::default(), RCMCallbackPath::new);
+        let callback_path = dapp_definitions.callback_path.unwrap_or_default();
```
Sajjon, line 73: the code and comment in the end of this function:

```rust
       // 5. TODO: use the actual dapp callback path
        Ok(Url::from_str("https://example.com").unwrap())
```

hmm... should we not do:

```rust
Ok(callback_path)
```
???

Or we we have two different callback paths?

src/radix_connect/mobile/session/callback_path.rs
Sajjon: takes 1 min to add them, so yes please do, it asserts that the Serde behaviour of this newtype is indeed "transparent", that wrapping the str in this newtype does not add one layer.

If the test fails you need to add `#[serde(transparent)]`, see `BIP39Passphrase`

src/radix_connect/mobile/client.rs
jakub-rdx: this is code Ghenadie wrote, I just added parsing callback_path and storing it. But yes, we will need to combine `origin` and `callback_path` together as final result of this method

src/radix_connect/well_known_client/models/dapp_definition.rs
```diff
@@ -1,23 +1,29 @@
 use crate::prelude::*;
 
-#[derive(Debug, Serialize, Deserialize, PartialEq)]
+/// Wrapper around the address of a dapp.
```
Sajjon, line 3: Is this just a wrapper around a dapp definition ADDRESS? If it is, then we should delete this, we already have:

```rust
pub type DappDefinitionAddress = AccountAddress;
```

I did not notice at first, since this type was just named `DappDefinition`. If this is not that, but an actual structure, with more than one field, it should remain. But at this point it is not. So I vote, delete this and use the already existing `DappDefinitionAddress` type (alias)!

src/radix_connect/well_known_client/models/dapp_definitions.rs
Sajjon: Is this question still relevant? 

`RCMCallbackPath` has the same shape as `BIP39Passphrase`. Just add `Serialize, Deserialize` in the derive... and use `#[serde(transparent)]`. look at `BIP39Passphrase`.

src/radix_connect/mobile/client.rs
GhenadieVP: @jakub-rdx you can use the callback_path now to construct the redirect URL now that we have it.

src/radix_connect/mobile/session/callback_path.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl From<String> for RCMCallbackPath {
+    fn from(value: String) -> Self {
+        Self(value)
+    }
+}
+
+impl Default for RCMCallbackPath {
+    fn default() -> Self {
+        Self::new("default_callback_path")
```
GhenadieVP, line 30: I think it should be `connect`

src/radix_connect/well_known_client/client.rs
```diff
@@ -0,0 +1,117 @@
+use crate::prelude::*;
+
+const SUFFIX_WELL_KNOWN_FILE: &str = ".well-known/dapp.json";
```
GhenadieVP, line 3: Isn't the path `.well-known/radix.json`?

src/radix_connect/well_known_client/models/dapp_definitions.rs
```diff
@@ -0,0 +1,67 @@
+use crate::prelude::*;
+
+/// Struct that represents content of well known file
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct DappDefinitions {
```
GhenadieVP, line 6: Maybe `DappWellKnownData/Info`? It should represent the well known that could contain more info in the future.

CyonAlexRDX (approved):
Good job @jakub-rdx ! LGTM!

GhenadieVP (approved):

PR #160, opened 2024-06-06
Truncate Decimal192 fractional part when it has more than 18 decimals

sergiupuhalschi-rdx:
Update `Decimal192` parsing to truncate the fractional part to 18 decimals when parsing into the `ScryptoDecimal192` fails with the specific `ParseDecimalError::MoreThanEighteenDecimalPlaces`.

### DIFF
--- a/src/core/types/decimal192.rs
+++ b/src/core/types/decimal192.rs
@@ -147,7 +147,7 @@ impl From<f32> for Decimal {
     ///
     /// assert!(Decimal::from(208050.17).to_string() == "208050.17");
     ///
-    /// assert!(Decimal::from(f32::MIN_POSITIVE) == "0");
+    /// assert!(Decimal::from(f32::MIN_POSITIVE).to_string() == "0");
     ///
     /// ```
     fn from(value: f32) -> Self {
--- a/src/core/types/decimal192.rs
+++ b/src/core/types/decimal192.rs
@@ -1248,6 +1248,17 @@ mod test_decimal {
     }
 
     #[test]
+    fn from_str_more_than_18_decimals_is_ok() {
+        assert_eq!(
+            "4.012345678901234567890"
+                .parse::<Decimal192>()
+                .unwrap()
+                .to_string(),
+            "4.012345678901234567"
+        ); // Over 18 decimals is OK (precision lost)
+    }
+
+    #[test]
     fn from_more_than_one_decimal_point_with_more_than_18_decimals_string() {
         assert_eq!(
             "4.0123456789012345678.123".parse::<Decimal192>(),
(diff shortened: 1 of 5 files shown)


Sajjon:
Hmm CodeCov says the execution path with `MoreThanEighteenDecimalPlaces` is never hit, is that correct?

sergiupuhalschi-rdx:
> Hmm CodeCov says the execution path with `MoreThanEighteenDecimalPlaces` is never hit, is that correct?

It should have been hit by the test from `test_parse_formatted_decimal`, but I've added an explicit test for this case anyway.

sergiupuhalschi-rdx:
> Hmm CodeCov says the execution path with `MoreThanEighteenDecimalPlaces` is never hit, is that correct?

I'm not sure why it's not marked as hit by CodeCov, that execution path is being called in 2 tests now. Unfortunately, I need to leave now and I can't address this until Monday :(

src/core/types/decimal192.rs
```diff
@@ -124,19 +149,11 @@ impl TryFrom<f32> for Decimal {
     ///
     /// assert!(Decimal::try_from(208050.17).is_ok());
     ///
-    /// assert_eq!(
-    ///     Decimal::try_from(f32::MIN_POSITIVE),
-    ///     Err(CommonError::DecimalOverflow { bad_value: f32::MIN_POSITIVE.to_string() })
-    /// );
+    /// assert!(Decimal::try_from(f32::MIN_POSITIVE).is_ok());
     /// ```
     fn try_from(value: f32) -> Result<Self, Self::Error> {
```
Sajjon, line 154: is it now in fact possible to get it to throw from an `f32`? it seems not? Maybe change this `TryFrom<f32>` into a `From<f32>` (return `Decimal192` instead of `Result<>`)? If you ensure to try `MAX_POSITIVE` and `MIN_POSITIVE`, `MAX_NEGATIVE` and `MIN_NEGATIVE` and none of those for 4 scenarios throw, then this cannot throw and it is safe to force unwrap

and if so, also change the `pub fn new_decimal_from_f32(value: f32) -> Result<Decimal192>` to not return `Result`.

What about `f64`? Same thing applies?

CyonAlexRDX:
> I'm not sure why it's not marked as hit by CodeCov, that execution path is being called in 2 tests now. Unfortunately, I need to leave now and I can't address this until Monday :(

You are right, it gets hit, tarpaulin is a bit quirky sometimes (still better than many other coverage tools!)! You see it actually counts the "body" of the variant hit! Where you call `Self::from_str_by_truncating`.

So you can mark this PR with the label for this: "Coverage Ok" or what I named it...

giannis-rdx (approved):

src/core/types/decimal192.rs
sergiupuhalschi-rdx: Good point! This applies to f32, but not to f64 as the parsing into `Scrypto192Decimal` fails with `Overflow` error.

PR #161, opened 2024-06-10
Change WalletInteractionId from Uuid to String

danvleju-rdx:
Change WalletInteractionId from Uuid to String


CyonAlexRDX (changes requested):
If we do this, we should change to NanoID, not String. Never String.

e.g. https://docs.rs/nanoid/latest/nanoid/

And have to make it UniFFI compatible. And tests that. And test UniFFI in Swift and Kotlin.

PR #163, opened 2024-06-11
Implement RadixConnectSessionStorage for kotlin

micbakos-rdx:
In order to implement such "secure" storage I had to introduce to sargon some logic that existed in Android wallet.

* Some `Result` extensions were needed. Unit tested and made public to replace the ones we have in android wallet.
* `KeySpec` reintroduced. Temporarily became `internal` in order to not introduce a duplicate type to the existing one, but rather lie the foundation for all kinds of secure storages to be implemented when SargonOS will be integrated. * * The only class that is public for construction from wallet host will be `RadixConnectSessionStorage(context: Context)`. The host should instantiate this as a singleton using Hilt.
* Unit tests are difficult when you combine code with methods that exist on Android. Tried to unit test everything. `EncryptionHelper ` can encrypt with either a `KeySpec` directly, or an `encryptionKey: ByteArray`. Only the latter is able to be used in unit tests since it is backed by [Kotlin's `Base64`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io.encoding/-base64/) implementation instead of [Android's Base64](https://developer.android.com/reference/android/util/Base64). Look at the comment bellow for this one.
* `KeySpec` is ignored from coverage. It is possible to test such code, but this will require an emulator and the test must be in a form of integration test. After discussion with @GhenadieVP we concluded to start simple and decide later what to do for code in sargon jvm that requires an emulator in order to be tested.
* As discussed in previous standup, the `KeySpec.RadixConnect` is configured to not require biometrics. (Implemented similar to `KeySpec.Profile`)

> [!NOTE]
> Target branch is not `main`

### DIFF
--- a/jvm/gradle/libs.versions.toml
+++ b/jvm/gradle/libs.versions.toml
@@ -16,6 +16,8 @@ compose-bom = "2024.02.00"
 kover = "0.7.6"
 datastore = "1.1.1"
 coroutines = "1.8.0"
+androidx-test = "1.5.0"
+androidx-test-junit = "1.1.5"
 
 [libraries]
 androidx-activity-compose = { module = "androidx.activity:activity-compose", version.ref = "activity-compose" }
@@ -34,6 +36,9 @@ junit-params = { module = "org.junit.jupiter:junit-jupiter-params", version.ref 
 coroutines-android = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-android", version.ref = "coroutines" }
 coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-test", version.ref = "coroutines" }
 mockk = { module = "io.mockk:mockk", version.ref = "mockk" }
+androidx-test-runner = { module = "androidx.test:runner", version.ref = "androidx-test" }
+androidx-test-rules = { module = "androidx.test:rules", version.ref = "androidx-test" }
+androidx-test-junit-ktx = { module = "androidx.test.ext:junit-ktx", version.ref = "androidx-test-junit" }
 
 [plugins]
 android-application = { id = "com.android.application", version.ref = "agp" }
--- a/jvm/gradle/libs.versions.toml
+++ b/jvm/gradle/libs.versions.toml
@@ -14,6 +14,8 @@ material = "1.11.0"
 cargo-ndk = "0.3.4"
 compose-bom = "2024.02.00"
 kover = "0.7.6"
+datastore = "1.1.1"
+coroutines = "1.8.0"
 
 [libraries]
 androidx-activity-compose = { module = "androidx.activity:activity-compose", version.ref = "activity-compose" }
@@ -25,9 +27,12 @@ androidx-compose-ui = { module = "androidx.compose.ui:ui" }
 androidx-compose-ui-graphics = { module = "androidx.compose.ui:ui-graphics" }
 androidx-compose-ui-preview = { module = "androidx.compose.ui:ui-tooling-preview" }
 androidx-compose-material3 = { module = "androidx.compose.material3:material3" }
+androidx-datastore-preferences = { module = "androidx.datastore:datastore-preferences", version.ref = "datastore" }
 material = { module = "com.google.android.material:material", version.ref = "material" }
 junit = { module = "org.junit.jupiter:junit-jupiter-engine", version.ref = "junit" }
 junit-params = { module = "org.junit.jupiter:junit-jupiter-params", version.ref = "junit" }
+coroutines-android = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-android", version.ref = "coroutines" }
+coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-test", version.ref = "coroutines" }
 mockk = { module = "io.mockk:mockk", version.ref = "mockk" }
 
 [plugins]
--- a/jvm/sargon-android/src/main/java/com/radixdlt/sargon/os/storage/EncryptionHelper.kt
+++ b/jvm/sargon-android/src/main/java/com/radixdlt/sargon/os/storage/EncryptionHelper.kt
@@ -1,0 +1,146 @@
+@file:OptIn(ExperimentalEncodingApi::class)
+
+package com.radixdlt.sargon.os.storage
+
+import com.radixdlt.sargon.extensions.then
+import java.nio.ByteBuffer
+import java.nio.charset.StandardCharsets
+import java.security.SecureRandom
+import java.security.spec.AlgorithmParameterSpec
+import javax.crypto.Cipher
+import javax.crypto.SecretKey
+import javax.crypto.spec.GCMParameterSpec
+import javax.crypto.spec.SecretKeySpec
+import kotlin.io.encoding.Base64
+import kotlin.io.encoding.ExperimentalEncodingApi
+
+internal object EncryptionHelper {
+    fun encrypt(
+        input: ByteArray,
+        secretKey: SecretKey
+    ): Result<ByteArray> {
+        return runCatching {
+            val cipher = Cipher.getInstance(AES_GCM_NOPADDING)
+            val ivBytes = ByteArray(GCM_IV_LENGTH)
+            SecureRandom().nextBytes(ivBytes)
+            val parameterSpec = GCMParameterSpec(AUTH_TAG_LENGTH, ivBytes)
+            cipher.init(Cipher.ENCRYPT_MODE, secretKey, parameterSpec)
+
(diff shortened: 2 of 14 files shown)


jvm/gradle/libs.versions.toml
```diff
@@ -14,6 +14,8 @@ material = "1.11.0"
 cargo-ndk = "0.3.4"
 compose-bom = "2024.02.00"
 kover = "0.7.6"
+datastore = "1.1.1"
+coroutines = "1.8.0"
```
giannis-rdx, line 18: maybe we should go to 1.9.0 and also update our wallet with 2.0.0 kotlin?

i know it's not the most important thing at the moment, so it's not a requirement from my side.

jvm/sargon-android/src/main/java/com/radixdlt/sargon/os/storage/EncryptionHelper.kt
```diff
@@ -0,0 +1,146 @@
...
+import java.security.SecureRandom
+import java.security.spec.AlgorithmParameterSpec
+import javax.crypto.Cipher
+import javax.crypto.SecretKey
+import javax.crypto.spec.GCMParameterSpec
+import javax.crypto.spec.SecretKeySpec
+import kotlin.io.encoding.Base64
+import kotlin.io.encoding.ExperimentalEncodingApi
+
+internal object EncryptionHelper {
```
CyonAlexRDX, line 17: No need for this, at all, right?! We already have AES in Sargon! We should not maintain multiple encryption "implementations".

CyonAlexRDX (changes requested):
I think we should not impl any encryption in Kotlin at all. Just like we should not have any encryption in Swift at all, we already have AES in Rust Sargon: https://github.com/radixdlt/sargon/blob/bakos/rcm_linking/src/profile/encrypted/encryption/aes_gcm_256.rs#L24

so write an ultra ultra "thin" wrapper which is UniFFI exporting that!

jvm/sargon-android/src/main/java/com/radixdlt/sargon/os/storage/EncryptionHelper.kt
```diff
@@ -0,0 +1,146 @@
...
+            source = EncryptionHelper.encrypt(
+                input = toByteArray(),
+                secretKey = secretKey
+            ).getOrThrow()
+        ) as T
+
+        is ByteArray -> EncryptionHelper.encrypt(input = this, secretKey = secretKey)
+            .getOrThrow() as T
+
+        else -> throw UnsupportedOperationException(
```
giannis-rdx, line 77: minor but wouldn't be more a `IllegalArgumentException` here?

giannis-rdx (approved):
👌🏽

jvm/sargon-android/src/main/java/com/radixdlt/sargon/os/storage/EncryptionHelper.kt
micbakos-rdx: I don't think we can do that. In order to encrypt with Keystore, we need to use Android APIs. Take a closer look at the encrypt method. Indeed the cipher we use is `AES/GCM/NoPadding`, but please take a look at the code:

```
val cipher = Cipher.getInstance(AES_GCM_NOPADDING)
val ivBytes = ByteArray(GCM_IV_LENGTH)
SecureRandom().nextBytes(ivBytes)
val parameterSpec = GCMParameterSpec(AUTH_TAG_LENGTH, ivBytes)
cipher.init(Cipher.ENCRYPT_MODE, secretKey, parameterSpec)

val ciphertext: ByteArray = cipher.doFinal(input)

val byteBuffer = ByteBuffer.allocate(ivBytes.size + ciphertext.size)
byteBuffer.put(ivBytes)
byteBuffer.put(ciphertext)

byteBuffer.array()
```

There is no other way to bypass this API if we want to use Android Keystore.

CyonAlexRDX: please elaborate...

CyonAlexRDX: which part is not possible to bypass?

CyonAlexRDX: The signature is:

```kotlin
 fun encrypt(
        input: ByteArray,
        secretKey: SecretKey
    ): Result<ByteArray> {
```

Which is the same signature as Sargon has:

```rust
    fn encrypt(
        &self,
        plaintext: impl AsRef<[u8]>,
        encryption_key: &mut EncryptionKey,
    ) -> Vec<u8> {
```

sergiupuhalschi-rdx (approved):

GhenadieVP (approved):
LGTM! Nice 💪.

micbakos-rdx:
@CyonAlexRDX @GhenadieVP 

A summary on how SecretKey behaves:

We have 2 kinds of `SecretKey` generation currently on Android. 

1. The one is used in WebSocket for WebRTC to encrypt information with an already provided `encryptionKey: ByteArray`
2. The other is used by creating specific Keys which are stored in Android Keystore. For these keys we provide an alias, so we can refer to them and some properties like
    * It is for encryption, or decryption, or both
    * If it is strong box backed.
    * The kind of alogrithm
    * The padding
    * If they require authentication by the user and what kinds of authentication (biometrics, passcode etc...)
----

Now `javax.crypto.SecretKey: java.security.Key`. This `Key` interface has some methods that would seem useful:
* `getAlgorithm(): String`
* `getFormat(): String`
* `getEncoded(): ByteArray`

The last one seems to be what we need right?

When we create a secret key with method (1), after calling `getEncoded()` we can retrieve the key.
But when we do the same with method (2), `getEncoded()` returns null. Also `getFormat()` returns null too. The underlying class is `class android.security.keystore2.AndroidKeyStoreSecretKey` which is not possible for us to cast since this is an Android internal class. 
The following:
```
secretKey as android.security.keystore2.AndroidKeyStoreKey

```
does not compile with `Unresolved reference: keystore2`

Last but not least I would advise looking at the [Keystore](https://developer.android.com/privacy-and-security/keystore#:~:text=The%20Android%20Keystore%20system%20lets,key%20material%20remaining%20non%2Dexportable.) documention. It specifically explains in the first paragraph:

```
The Android Keystore system lets you store cryptographic keys in a container
to make them more difficult to extract from the device. Once keys are in the
keystore, you can use them for cryptographic operations, with the key material 
remaining non-exportable. Also, the keystore system lets you restrict when and 
how keys can be used, such as requiring user authentication for key use or 
restricting keys to use only in certain cryptographic modes. 
See the [Security Features](https://developer.android.com/privacy-and-security/keystore#SecurityFeatures) section for more information.
```
There is also a specific section regarding [Extraction Prevention](https://developer.android.com/privacy-and-security/keystore#ExtractionPrevention).

jvm/sargon-android/src/main/java/com/radixdlt/sargon/os/storage/EncryptionHelper.kt
micbakos-rdx: Right will change that

CyonAlexRDX:
> @CyonAlexRDX @GhenadieVP
> 
> A summary on how SecretKey behaves:

Thank you for that excellent summary, that was what I wanted to see :) good to merge :) 

Unfortunate that Android does not provide the `EncryptionHelper` behaviour as is, very strange! But it is what it is. Good that `EncryptionHelper` is not public, so we dont use this AES impl for anything else than this very specific scenario. For all other situations, if we find the need, lets UniFFI export the AES impl in Sargon.

micbakos-rdx:
> Thank you for that excellent summary, that was what I wanted to see :) good to merge :)
> 
> Unfortunate that Android does not provide the `EncryptionHelper` behaviour as is, very strange! But it is what it is. Good that `EncryptionHelper` is not public, so we dont use this AES impl for anything else than this very specific scenario. For all other situations, if we find the need, lets UniFFI export the AES impl in Sargon.

Yes, the goal was to only expose the driver now (`RadixConnectSessionStorage`), and later the bios only which will be kept as a singleton by `Hilt` (dependency injection library).

micbakos-rdx:
Implemented some instrumentation tests (tests that require the android runtime) regarding `KeySpec.RadixConnect` https://github.com/radixdlt/sargon/pull/163/commits/b9f54df7f827c077d56b4da66a660bd1e8d69285 

The command to run those in a connected device/emulator is
`./gradlew sargon-android:connectedDebugAndroidTest`

Currently I will not update the README about it since there is no support in CI for running such tests.

PR #164, opened 2024-06-12
Expose kotlin serde for RadixConnectMobileLinkRequest

micbakos-rdx:
* Create kotlin serde for RadixConnectMobileLinkRequest. On Android we need to serialize objects to pass them between different screens.
* Make `RadixConnectSessionStorage` internal as the only starting point for radix connect is `RadixConnectMobile` which got the appropriate initialiser.


CyonAlexRDX (approved):
LGTM

sergiupuhalschi-rdx (approved):

PR #169, opened 2024-06-19
Pass serde message information to `FailedToDeserializeJSONToValue`

micbakos-rdx:
`FailedToDeserializeJSONToValue` was missing meaningful information that hosts could log, in order to find out what the problem is when deserialising a value.

2 methods can be applied on each `Result<*, serde_json::Error>` that can extract all needed information about deserialisation for both strings and byte arrays.

The aim is to use the `message` in hosts to extract logs.

### DIFF
--- a/src/core/error/common_error.rs
+++ b/src/core/error/common_error.rs
@@ -255,6 +255,7 @@ pub enum CommonError {
     FailedToDeserializeJSONToValue {
         json_byte_count: u64,
         type_name: String,
+        message: String,
     } = 10070,
 
     #[error("Failed To create ProfileID (UUID) from string: {bad_value}")]
--- a/src/core/error/common_error.rs
+++ b/src/core/error/common_error.rs
@@ -251,11 +251,11 @@ pub enum CommonError {
     #[error("Failed Serialize value to JSON.")]
     FailedToSerializeToJSON = 10069,
 
-    #[error("Failed deserialize JSON with #{json_byte_count} bytes to value of type {type_name} with error: \"{message}\"")]
+    #[error("Failed deserialize JSON with #{json_byte_count} bytes to value of type {type_name} with error: \"{serde_message}\"")]
     FailedToDeserializeJSONToValue {
         json_byte_count: u64,
         type_name: String,
-        message: String,
+        serde_message: String,
     } = 10070,
 
     #[error("Failed To create ProfileID (UUID) from string: {bad_value}")]
--- a/src/core/error/common_error.rs
+++ b/src/core/error/common_error.rs
@@ -251,7 +251,7 @@ pub enum CommonError {
     #[error("Failed Serialize value to JSON.")]
     FailedToSerializeToJSON = 10069,
 
-    #[error("Failed deserialize JSON with #{json_byte_count} bytes to value of type {type_name}")]
+    #[error("Failed deserialize JSON with #{json_byte_count} bytes to value of type {type_name} with error: \"{message}\"")]
     FailedToDeserializeJSONToValue {
         json_byte_count: u64,
         type_name: String,
--- a/src/core/error/common_error_map.rs
+++ b/src/core/error/common_error_map.rs
@@ -1,0 +1,56 @@
+use de::StdError;
+
+use crate::prelude::*;
+
+pub trait MapToFailedToDeserializeJSONToValue<R> {
+    fn map_failed_to_deserialize_string<T>(
+        self,
+        input: impl AsRef<str>,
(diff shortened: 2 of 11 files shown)


src/core/error/common_error.rs
```diff
@@ -255,6 +255,7 @@ pub enum CommonError {
     FailedToDeserializeJSONToValue {
```
GhenadieVP, line 255: Would be nice to also add the message in the `#[error` on top

micbakos-rdx: Does this affect the size of the enum? I was afraid to add it.

CyonAlexRDX: yes it does, but we can add it for now, and remove it once UniFFI explodes :D

src/core/error/common_error_map.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    fn map_failed_to_deserialize_bytes<T>(
+        self,
+        input: &[u8],
+    ) -> Result<R, CommonError> {
+        self.map_err(|e| {
+            error!(
+                "Failed to deserialize JSON to {}, from (UTF-8):\n{:?}\nError: {}", 
+                type_name::<T>(),
+                String::from_utf8(input.to_vec()),
```
CyonAlexRDX, line 46: we should probably not print the *whole* `input`, it can blow up, imagine for a 1MB Transaction (many blobs) it will blow up :) maybe do a .prefix(500) or otherwise reasonable limit?

src/profile/v100/json_data_convertible.rs
```diff
@@ -101,16 +93,31 @@ macro_rules! json_data_convertible {
...
+                            assert_eq!(
+                                {{
+                                    const STRINGIFIED: &'static str = stringify!($type);
+                                    STRINGIFIED
+                                }}.to_owned(),
+                                type_name,
+                            );
+                        }
+                        _ => {
+                            error!("Expected CommonError::FailedToDeserializeJSONToValue but other error occurred")
```
CyonAlexRDX, line 118: should probably `panic!` here instead of print an error. Most prints are filtered out, and when we `panic!` in a unit test, that is a failure of the unit test, which we want

src/profile/v100/json_data_convertible.rs
```diff
@@ -181,16 +188,31 @@ macro_rules! json_string_convertible {
...
```
CyonAlexRDX, line 213:
```suggestion
                            panic!("Expected CommonError::FailedToDeserializeJSONToValue but other error occurred")
```

src/core/error/common_error_map.rs
```diff
@@ -0,0 +1,56 @@
...
+            error!(
+                "Failed to deserialize JSON to {}, from:\n{:?}\nError: {}",
+                type_name::<T>(),
+                input.as_ref(),
+                e
+            );
+            CommonError::FailedToDeserializeJSONToValue {
+                json_byte_count: input.as_ref().len() as u64,
+                type_name: type_name::<T>(),
+                message: format!("{}", e),
```
CyonAlexRDX, line 33: if you do `format!("{:?}", e)` instead, do we get any more meaningful info? i.e. does `"expected value at line 1 column 1"` change into something even more helpful?

src/core/error/common_error_map.rs
micbakos-rdx: Is there such method on &str?

src/profile/v100/json_data_convertible.rs
micbakos-rdx: Yes that is what I was looking for. Confused with kotlin's error that does the same thing as panic here.

src/core/error/common_error_map.rs
micbakos-rdx: Not particularly. For this mentioned example we get `Error(\"expected value\", line: 1, column: 1)`

src/core/error/common_error_map.rs
CyonAlexRDX: very many different ways of doing it, you will see hundreds of suggestions when googled. Not sure if `.chars().take(n)` works if `n` > `len()`, check! if it works you can use that. (and then `.collect::<String>()`

src/core/error/common_error_map.rs
CyonAlexRDX: mm not so helpful :P

GhenadieVP (approved):

sergiupuhalschi-rdx (approved):

PR #170, opened 2024-06-19
[ABW-3433] Fix Manifest Crash

kugel3:
Jira ticket: [ABW-3433](https://radixdlt.atlassian.net/browse/ABW-3433)

- Removes the check `guarantees.len() > self.instructions().len()`
- Propagates panics to Swift as thrown errors

[ABW-3433]: https://radixdlt.atlassian.net/browse/ABW-3433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core/error/common_error.rs
+++ b/src/core/error/common_error.rs
@@ -493,7 +493,9 @@ pub enum CommonError {
     #[error("Invalid RadixConnectPurpose, bad value: {bad_value}")]
     InvalidRadixConnectPurpose { bad_value: String } = 10137,
 
-    #[error("Index of transaction Guarantee is out of bounds: {index} >= #{count}")]
+    #[error(
+        "Index of transaction Guarantee is out of bounds: {index} >= #{count}"
+    )]
     TXGuaranteeIndexOutOfBounds { index: u64, count: u64 } = 10138,
 }
 
--- a/src/core/error/common_error.rs
+++ b/src/core/error/common_error.rs
@@ -494,7 +494,7 @@ pub enum CommonError {
     InvalidRadixConnectPurpose { bad_value: String } = 10137,
 
     #[error(
-        "Index of transaction Guarantee is out of bounds: {index} >= #{count}"
+        "Transaction Guarantee's 'instruction_index' is out of bounds, the provided manifest contains #{count}, but an 'instruction_index' of {index} was specified."
     )]
     TXGuaranteeIndexOutOfBounds { index: u64, count: u64 } = 10138,
 }
--- a/src/core/error/common_error.rs
+++ b/src/core/error/common_error.rs
@@ -493,8 +493,8 @@ pub enum CommonError {
     #[error("Invalid RadixConnectPurpose, bad value: {bad_value}")]
     InvalidRadixConnectPurpose { bad_value: String } = 10137,
 
-    #[error("Index out of bounds: {index} >= #{count}")]
-    IndexOutOfBounds { index: u64, count: u64 } = 10138,
+    #[error("Index of transaction Guarantee is out of bounds: {index} >= #{count}")]
+    TXGuaranteeIndexOutOfBounds { index: u64, count: u64 } = 10138,
 }
 
 #[uniffi::export]
--- a/src/core/error/common_error.rs
+++ b/src/core/error/common_error.rs
@@ -492,6 +492,9 @@ pub enum CommonError {
 
     #[error("Invalid RadixConnectPurpose, bad value: {bad_value}")]
     InvalidRadixConnectPurpose { bad_value: String } = 10137,
+
+    #[error("Index out of bounds: {index} >= #{count}")]
+    IndexOutOfBounds { index: u64, count: u64 } = 10138,
 }
 
 #[uniffi::export]
--- a/src/wrapped_radix_engine_toolkit/high_level/manifest_building/modify_manifest.rs
+++ b/src/wrapped_radix_engine_toolkit/high_level/manifest_building/modify_manifest.rs
@@ -66,13 +66,13 @@ impl TransactionManifest {
     ///
     /// Also panics if the number of TransactionGuarantee's is larger than the number
     /// of instructions of `manifest` (does not make any sense).
-    pub(crate) fn modify_add_guarantees<I>(self, guarantees: I) -> Self
+    pub(crate) fn modify_add_guarantees<I>(self, guarantees: I) -> Result<TransactionManifest>
     where
         I: IntoIterator<Item = TransactionGuarantee>,
     {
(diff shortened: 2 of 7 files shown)


src/wrapped_radix_engine_toolkit/high_level/manifest_building/modify_manifest.rs
```diff
@@ -471,30 +455,8 @@ CALL_METHOD
...
-            ]),
-            manifest
-        );
-    }
-
-    #[test]
-    #[should_panic(
-        expected = "Does not make sense to add more guarantees than there are instructions."
-    )]
-    fn test_modify_manifest_add_guarantees_panics_if_more_guarantees_than_instructions(
```
GhenadieVP, line 483: Maybe we should keep this kind of test, but the assertion would be that it does not fail if there are lots of guarantees?

kugel3: I thought about it, but it seemed a bit strange to test that we _don't_ arbitrarily limit the number of guarantees... But I can of course add it.

GhenadieVP: I agree, it is sort of a paranoia check, but still better to have

src/wrapped_radix_engine_toolkit/high_level/manifest_building/modify_manifest.rs
```diff
@@ -66,21 +66,20 @@ impl TransactionManifest {
     ///
     /// Also panics if the number of TransactionGuarantee's is larger than the number
     /// of instructions of `manifest` (does not make any sense).
-    pub(crate) fn modify_add_guarantees<I>(self, guarantees: I) -> Self
+    pub(crate) fn modify_add_guarantees<I>(
+        self,
+        guarantees: I,
+    ) -> Result<TransactionManifest>
```
matiasbzurovski, line 72: From what we've discussed on Slack, the only reason we add this `Result<>` is so that UniFFi converts it into a throwable method when translated into Swift. If that's the case, one could be tempted to remove the `Result<>` from this function as it is never used. Therefore, to avoid this issue coming back in the future, I'd recommend:
- adding a comment explaining why we need to use `Result<>`
- update line 88 so that instead of `panic!`, we return some `Err` and then catch such error on the host side. Would that work?

matiasbzurovski (commented):
Missing bump of version on `Cargo.toml`.

src/wrapped_radix_engine_toolkit/high_level/manifest_building/modify_manifest.rs
kugel3: Returning an error should work, though then we'd need to add that error to CommonError. If we did that, we wouldn't have to add a comment though, since the Result type would make sense.

matiasbzurovski: right, I meant one of the two options

kugel3: Got it. The second one seems to make sense either way. It is very strange that Uniffi converts the panic into a secret Err.

GhenadieVP: Agree, we should replace the specific panic! in this function with a CommonError.

kugel3: Added IndexOutOfBounds to CommonError

src/wrapped_radix_engine_toolkit/high_level/ret_api.rs
```diff
@@ -123,14 +123,11 @@ pub fn modify_manifest_lock_fee(
 /// # Panics
 /// Panics if any of the TransactionGuarantee's `instruction_index` is out of
```
Sajjon, line 124: Please update doc here as well, saying we panic if we are out of bounds, no longer true since replaced with an error, right?

kugel3: Correct. I know I removed this comment, maybe that got lost in rebasing, will remove again

src/core/error/common_error.rs
```diff
@@ -492,6 +492,9 @@ pub enum CommonError {
 
     #[error("Invalid RadixConnectPurpose, bad value: {bad_value}")]
     InvalidRadixConnectPurpose { bad_value: String } = 10137,
+
+    #[error("Index out of bounds: {index} >= #{count}")]
+    IndexOutOfBounds { index: u64, count: u64 } = 10138,
```
Sajjon, line 497: Hmm, I think this error is not too helpful for us debugging, we should use a more specific one:
`TXGuaranteeIndexOutOfBounds { ... }` (only used for this scenario).

At least until we have come up with some genius "error system", because now we don't have any "context" and when left as `IndexOutOfBounds` it makes sense to use it elsewhere for possible OOB errors, and then it will make debugging harder, when we are given an Error Code by user. See how detailed I've been for other errors:
```rust
    InvalidEd25519PrivateKeyFromBytes { bad_value: BagOfBytes } = 10001,

    #[error("Failed to create Ed25519 Private key from String {bad_value}.")]
    InvalidEd25519PrivateKeyFromString { bad_value: String } = 10002,

    #[error(
        "Failed to create Secp256k1 Private key from bytes {bad_value:?}."
    )]
    InvalidSecp256k1PrivateKeyFromBytes { bad_value: BagOfBytes } = 10003,

    #[error(
        "Failed to create Secp256k1 Private key from String {bad_value:?}."
    )]
    InvalidSecp256k1PrivateKeyFromString { bad_value: String } = 10004,
```

These four could have been 1, `InvalidPrivateKey`, but now we get much more details, we know which curve and from which input.

CyonAlexRDX (changes requested):
Nice! I just think we should change the error to be specifically about TXGuarantees before merging

src/core/error/common_error.rs
kugel3: Sure, I really don't have an opinion on the name of the error, it has been changed now

CyonAlexRDX (approved):
LGTM! Congrats on your first Rust PR 🥳

GhenadieVP (approved):

matiasbzurovski (approved):
Need to bump version again (https://github.com/radixdlt/sargon/pull/171 was merged in the middle), otherwise looks great

danvleju-rdx (approved):

PR #171, opened 2024-06-20
Truncate display names with longer than 30 chars

micbakos-rdx:
* User reported with profile that had display names with longer than 30 chars. This was resulting the deserialisation of the profile to return an error.
* The display names are now truncated.
* Question for @CyonAlexRDX: The `CommonError::InvalidDisplayNameTooLong` in now unused. I purposely did not delete it. Is it safe to remove it and fix the error codes of the rest of the following errors?
* Question to all. I used the [`truncate()`](https://doc.rust-lang.org/std/string/struct.String.html#method.truncate) function. This might panic if "new_len does not lie on a [char](https://doc.rust-lang.org/std/primitive.char.html) boundary." Can someone help with suggestions on how this might fail? Or even suggest a better function?

### DIFF
--- a/src/profile/v100/entity/display_name.rs
+++ b/src/profile/v100/entity/display_name.rs
@@ -44,13 +44,14 @@ impl DisplayName {
     pub const MAX_LEN: usize = 30;
 
     pub fn new(value: impl AsRef<str>) -> Result<Self> {
-        let mut value = value.as_ref().trim().to_string();
+        let value = value.as_ref().trim().to_string();
         if value.is_empty() {
             return Err(CommonError::InvalidDisplayNameEmpty);
         }
-        value.truncate(Self::MAX_LEN);
 
-        Ok(Self { value })
+        Ok(Self {
+            value: value.chars().take(Self::MAX_LEN).collect(),
+        })
     }
 }
 
--- a/src/profile/v100/entity/display_name.rs
+++ b/src/profile/v100/entity/display_name.rs
@@ -13,6 +13,15 @@ use crate::prelude::*;
 /// assert_eq!("Satoshi".parse::<SUT>().unwrap().to_string(), "Satoshi");
 /// ```
 ///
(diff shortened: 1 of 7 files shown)


src/profile/v100/entity/display_name.rs
```diff
@@ -35,16 +44,11 @@ impl DisplayName {
...
         if value.is_empty() {
             return Err(CommonError::InvalidDisplayNameEmpty);
         }
-        if value.len() > Self::MAX_LEN {
-            return Err(CommonError::InvalidDisplayNameTooLong {
-                expected: Self::MAX_LEN as u64,
-                found: value.len() as u64,
-            });
-        }
+        value.truncate(Self::MAX_LEN);
```
Sajjon, line 51: use the same you used in recent PR? `.chars().take(Self::MAX_LEN).collect()` instead? which seemed safe.?

micbakos-rdx:
> Yes ok to leave error variant for now. Too costly to shift all codes... you might rename it UnusedError or something? And make a note that we can use "that slot" if we want

Clever idea, will do that

jakub-rdx (approved):

src/profile/v100/entity/display_name.rs
```diff
@@ -39,14 +48,10 @@ impl DisplayName {
...
-        if value.len() > Self::MAX_LEN {
-            return Err(CommonError::InvalidDisplayNameTooLong {
-                expected: Self::MAX_LEN as u64,
-                found: value.len() as u64,
-            });
-        }
 
-        Ok(Self { value })
+        Ok(Self {
+            value: value.chars().take(Self::MAX_LEN).collect(),
```
CyonAlexRDX, line 53: Since we do this in two places, might as well write a func or method for it, put it in `string_utils` and add tests for it. 

You can perhaps extend the trait `StrExt`

CyonAlexRDX (approved):
LGTM !

PR #174, opened 2024-06-21
Fix Transaction Manifest builder

matiasbzurovski:
## Changelog
Fixes some issues introduced on https://github.com/radixdlt/sargon/pull/173, where the Transaction Manifest generated would be invalid due since some of the referred ids included in metadata weren't present on Stokenet.

[Here](https://stokenet-dashboard.radixdlt.com/resource/resource_tdx_2_1t4h3u5knhu308hfp5h6lwf7cq6pjl4wu0cvmjwjfy32hjfnzd745dj/metadata) is an example token built using these changes.

### DIFF
--- a/src/wrapped_radix_engine_toolkit/high_level/token_definition_metadata.rs
+++ b/src/wrapped_radix_engine_toolkit/high_level/token_definition_metadata.rs
@@ -19,6 +19,21 @@ impl From<TokenDefinitionMetadata>
     for ScryptoModuleConfig<ScryptoMetadataInit>
 {
     fn from(value: TokenDefinitionMetadata) -> Self {
+        let resource_address = ResourceAddress::sample_stokenet_nft_abandon();
+        let non_fungible_local_id = NonFungibleLocalId::Integer { value: 1 };
+        let non_fungible_local_id_other =
+            NonFungibleLocalId::Integer { value: 2 };
+        let non_fungible_resource_address =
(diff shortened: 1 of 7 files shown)


src/wrapped_radix_engine_toolkit/high_level/token_definition_metadata.rs
```diff
@@ -41,7 +56,7 @@ impl From<TokenDefinitionMetadata>
                 ScryptoMetadataValue::StringArray(value.tags),
             ),
             (
-                "extra_String".to_string(),
+                "extra_string".to_string(),
```
Sajjon, line 59: Any reason why those keys are string literals instead of variants on the MetadataKey (or at least another new enum for ExtraMetadataKeys)?

matiasbzurovski: I've named them `extra_XYZ` because they don't belong to the "predefined" metadata keys that we search for its values in the host apps (`name`, `symbol`, `icon_url`). They are named however the resource creator decided to name them.

We won't be searching for `extra_string` on the host app, just displaying the key/value pair. In this case it is `"extra_string": "foo_bar"`, but it could have been anything else (`"whatever": "lorem ipsum"`).

CyonAlexRDX (approved):
LGTM

PR #178, opened 2024-07-01
Fix new sessions handling

GhenadieVP:
This is a fix for the following bug:
- User performs the first connect to a dApp.
- User is presented with Origin verification screen.
- User goes back to the dApp and sends a another request. RDT will attach the same session_id.
- User is redirected to the Wallet.
- User dismisses the first request.
- When user tries to complete the second request and error will be thrown telling them that the session for the given session id was not found.

The above did happen because the session was saved to the storage only on success interaction, and it was immediately removed after user completed the interaction.

The fix for this is to have the per wallet_interaction_id session, so until a session is verified for a given session_id we can have multiple sessions in memory for different interactions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sargon"
-version = "1.0.19"
+version = "1.0.21"
 edition = "2021"
 build = "build.rs"
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2593,7 +2593,7 @@ dependencies = [
 
 [[package]]
 name = "sargon"
-version = "1.0.19"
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "sargon"
-version = "1.0.19"
+version = "1.0.21"
```
GhenadieVP, line 3: Bump to 21, as the update to 20 was not done.

sergiupuhalschi-rdx (approved):
LGTM

danvleju-rdx (approved):

micbakos-rdx (approved):

PR #180, opened 2024-07-02
[ABW-3465] Decode deep link and download metadata

matiasbzurovski:
## Changelog
Adds the `DeferredDeepLinkParser` that will be responsible for decoding the information sent by a deferred deep link. From such data, it will create the corresponding list of `HomeCards` that the host app need to display.

Also, it will extend the API models to allow parsing the metadata when fetching an entity details. Such metadata will be use to retrieve the `icon_url` of a `dapp_referrer`.


src/wallet/home_cards/manager.rs
```diff
@@ -0,0 +1,92 @@
...
+pub struct HomeCardsManager {
+    #[allow(dead_code)] // Remove this line after gateway_client is being used
+    parser: DeferredDeepLinkParser,
+    #[allow(dead_code)] // Remove this line after cards_storage is being used
+    cards_storage: Arc<dyn HomeCardsStorage>,
+    observer: Arc<dyn HomeCardsObserver>,
+    cards: RwLock<HomeCards>,
+}
+
+impl HomeCardsManager {
```
matiasbzurovski, line 28: This logic is being worked on by @sergiupuhalschi-rdx , as part of [ABW-3466](https://radixdlt.atlassian.net/browse/ABW-3466)

src/wallet/home_cards/deferred_deep_link/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+            .map_err(|e| {
+                println!("{}", e);
+                CommonError::DeferredDeepLinkInvalidValueFormat
+            })?;
+
+        let deep_link_value =
+            serde_json::from_slice::<OnboardingDeepLinkValue>(
+                decoded_value_json_bytes.as_ref(),
+            )
+            .map_err(|_| CommonError::DeferredDeepLinkDecodingFailed)?;
```
GhenadieVP, line 44: `?` could be removed and just return `deep_link_value`, which in this case could be renamed to `deep_link_value_result`

src/wallet/home_cards/manager.rs
```diff
@@ -0,0 +1,378 @@
...
+    /// Loads the home cards from storage.
+    async fn load_cards(&self) -> Result<HomeCards> {
+        let cards_bytes = self
+            .cards_storage
+            .load_cards()
+            .await?
+            .ok_or(CommonError::FailedAccessingHomeCards)?;
```
GhenadieVP, line 152: This error is more like there are no cards in the storage, no? Is it an error if there are no cards?

src/core/error/common_error.rs
```diff
@@ -544,6 +544,21 @@ pub enum CommonError {
 
     #[error("Radix Connect Mobile failed to create new in flight session")]
     RadixConnectMobileFailedToCreateNewSession = 10151,
+
+    #[error("Deferred Deep Link invalid deep link value format")]
+    DeferredDeepLinkInvalidValueFormat = 10152,
```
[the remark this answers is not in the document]
matiasbzurovski: This is the analogue of `RadixConnectMobileInvalidRequestFormat`. Maybe we can join both of them in a generic `Base64InvalidFormat`?

fixtures/models/gateway/state/response_entity_details__single_account_no_assets.json
```diff
@@ -25,83 +25,9 @@
 			},
 			"metadata":
 			{
-				"total_count": 2,
+				"total_count": 0,
```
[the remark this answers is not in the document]
matiasbzurovski: yes, although it would be clear on a specific test for it. Will revert change and add `response_entity_details__single_resource_no_metadata.json` and its corresponding test.

src/wallet/home_cards/manager.rs
sergiupuhalschi-rdx: Most likely yes, but it's for any error returned by the `load_cards` trait function. Should I rename it to `HomeCardsNotFound` to match the equivalent error mapped from `session_storage.load_session`? Please let me know if you think of a better name for this kind of error.

src/gateway_api/models/types/request/state/entity/opt_ins.rs
```diff
@@ -0,0 +1,25 @@
...
+    Clone,
+    Debug,
+    PartialEq,
+    Eq,
+    Serialize,
+    Deserialize, /* Deserialize so we can test roundtrip of JSON vectors */
+)]
+pub struct StateEntityDetailsOptIns {
+    #[serde(
+        rename = "explicit_metadata",
```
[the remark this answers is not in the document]
matiasbzurovski: not really, I thought you needed to specify the `rename` param in order to set `skip_serializing_if`, but seems it isn't necessary. Will remove here and in similar cases.

src/gateway_api/models/types/response/state/entity/details/metadata/collection.rs
```diff
@@ -0,0 +1,8 @@
+use crate::prelude::*;
+
+#[derive(
+    Deserialize, Serialize, Clone, PartialEq, Eq, Debug, uniffi::Record,
```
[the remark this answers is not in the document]
matiasbzurovski: Yes, since we are exposing `StateEntityDetailsResponse` (which has a `Vec<StateEntityDetailsResponseItem>`, and each item has an `EntityMetadataCollection`).

src/wallet/home_cards/deferred_deep_link/deferred_deep_link_special_dapp.rs
```diff
@@ -0,0 +1,30 @@
+use crate::prelude::*;
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
+#[serde(rename_all = "lowercase")]
+pub enum DeferredDeepLinkSpecialDapp {
```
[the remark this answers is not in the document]
matiasbzurovski: that we weren't very creative on Friday evening meeting to name it better 😅

src/core/error/common_error.rs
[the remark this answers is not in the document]
GhenadieVP: I guess my discomfort was using the `DeferredDeepLink` terminology, but after revisiting the code and what AppsFlyer does I am ok with it,

src/wallet/home_cards/deferred_deep_link/deferred_deep_link_special_dapp.rs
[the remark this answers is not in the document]
GhenadieVP: Then at least some docs would help :)

src/gateway_api/models/types/response/state/entity/details/metadata/collection.rs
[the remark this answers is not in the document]
GhenadieVP: .... ok, there is really no reason for `StateEntityDetailsResponse` to be UniFFI exposed, but since it is already, so be it for now.

src/wallet/home_cards/deferred_deep_link/parser.rs
```diff
@@ -0,0 +1,286 @@
...
+        self.transform_onboarding_deep_link_value(decoded).await
+    }
+}
+
+impl Parser {
+    fn decode(
+        &self,
+        encoded_value: impl AsRef<str>,
+    ) -> Result<OnboardingDeepLinkValue> {
+        let decoded_value_json_bytes = URL_SAFE_NO_PAD
```
GhenadieVP, line 33: For the errors thrown, let's forward the error message in the CommonError, it would greatly help to know what exactly failed if it will happen.

src/wallet/home_cards/deferred_deep_link/parser.rs
```diff
@@ -0,0 +1,286 @@
...
+            })?;
+
+        serde_json::from_slice::<OnboardingDeepLinkValue>(
+            decoded_value_json_bytes.as_ref(),
+        )
+        .map_err(|_| CommonError::DeferredDeepLinkDecodingFailed)
+    }
+}
+
+#[cfg(test)]
```
GhenadieVP, line 47: Better for tests to be at the bottom, I was confused finding non test code after test code :)

src/wallet/home_cards/deferred_deep_link/parser.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        if let Some(special_dapp) = value.special_dapp {
+            if special_dapp == DeferredDeepLinkSpecialDapp::RadQuest {
+                result.push(HomeCard::ContinueRadQuest);
+            } else {
+                result.push(HomeCard::StartRadQuest);
+            }
+        } else {
+            result.push(HomeCard::StartRadQuest);
+        }
```
GhenadieVP, line 101: should be possible to rewrite as such:
```rust
match value.special_dapp {
    Some(DeferredDeepLinkSpecialDapp::RadQuest) => result.push(HomeCard::ContinueRadQuest),
    _ => result.push(HomeCard::StartRadQuest),
}
```

src/wallet/home_cards/deferred_deep_link/parser.rs
```diff
@@ -0,0 +1,286 @@
...
+                    if let Some(icon_url) = metadata.get_icon_url() {
+                        result.push(HomeCard::Dapp {
+                            icon_url: Some(icon_url),
+                        });
+                    } else {
+                        result.push(HomeCard::Dapp { icon_url: None });
+                    }
+                }
+                Err(_) => result.push(HomeCard::Dapp { icon_url: None }),
+            }
```
GhenadieVP, line 115: You should be able to convert this, to something like this:
```rust
let icon_url = self.gateway_client.fetch_dapp_metadata(dapp_referrer).await
        .ok()
        .and_then(|metadata| metadata.get_icon_url());
    
result.push(HomeCard::Dapp { icon_url });
```
given that HomeCard::Dapp is always appened in the all execution paths.

src/wallet/home_cards/deferred_deep_link/parser.rs
```diff
@@ -0,0 +1,286 @@
...
+        let result = sut.decode(encoded_value).unwrap_err();
+        assert_eq!(result, CommonError::DeferredDeepLinkDecodingFailed);
+    }
+}
+
+impl Parser {
+    async fn transform_onboarding_deep_link_value(
+        &self,
+        value: OnboardingDeepLinkValue,
+    ) -> Result<HomeCards> {
```
GhenadieVP, line 90: After going through this function, I don't see any error that is being thrown, so this does not need to be a Result.

src/wallet/home_cards/deferred_deep_link/parser.rs
matiasbzurovski: Yeah, I can see why this is confusing, will move. My idea was following some idea that was mentioned on the course about placing the tests next to the function/feature tested. However, it definitely doesn't help when looking at a single file.

Unrelated, but one thing that bothers me a bit about our project is that every time I want to search for how something similar is done/used, I mostly get test search results. Then you need to manually filter among each results for the cases that are actually production code.
This could be resolved if tests were placed under a separate file with `_tests.rs` suffix. This way, you could easily filter them out on searches.

src/wallet/home_cards/deferred_deep_link/parser.rs
matiasbzurovski: nice!

src/wallet/home_cards/deferred_deep_link/parser.rs
matiasbzurovski: nice x2!

src/wallet/home_cards/deferred_deep_link/parser.rs
GhenadieVP: about tests placement - we will likely have separate files for tests once we will move to have a workspace with multiple crates, then each crate will have at top level two folders `src` and `tests`.

src/wallet/home_cards/manager.rs
GhenadieVP: `HomeCardsNotFound` sounds good.

src/wallet/home_cards/manager.rs
```diff
@@ -76,16 +83,33 @@ impl HomeCardsManager {
     }
 
     /// Handles a deferred deep link by parsing it and saving the generated `HomeCards` to `HomeCardsStorage`.
+    /// `HomeCard::ContinueRadQuest` if found in the link parsing result, replaces `HomeCard::StartRadQuest` if it's stored.
```
GhenadieVP, line 86: hmm, this does not seem to be possible - by definition this function should be called only after user did install the wallet, so there cannot be `StartRadQuest` stored. 
Rather, it is the other way around - in `wallet_created`, we should not add the `HomeCard::StartRadQuest` if `HomeCard::ContinueRadQuest` is already present.

sergiupuhalschi-rdx: You're right. What about keeping this logic in both methods to make sure there is no chance both `StartRadQuest` and `ContinueRadQuest` are present at the same time? This way Sargon won't care about the order of the method calls. Also, for the deep link to be resolved the AF SDK performs a network request, then we perform a network request for the dApp icon URL, so there is a tiny possibility the deep link won't be handled fast enough. Keeping this logic also would prevent any scenarios like this.

src/wallet/home_cards/manager.rs
```diff
@@ -149,17 +133,37 @@ impl HomeCardsManager {
...
+        println!("Cards to insert after: {:?}", write_guard.clone());
+
+        cards_to_insert.into_iter().for_each(|card| {
             if write_guard.try_insert_unique(card).is_ok() {
                 debug!("Home card inserted");
             } else {
                 debug!("Home card insert failed");
             }
-        })
+        });
```
GhenadieVP, line 164: What about this as alternative:
```rust
// Insert all cards into write_guard
cards.into_iter().for_each(|card| {
    if write_guard.try_insert_unique(card).is_ok() {
        debug!("Home card inserted");
    } else {
        debug!("Home card insert failed");
    }
});

// Check if ContinueRadQuest is present and remove StartRadQuest if it is
if write_guard.contains_id(&HomeCard::ContinueRadQuest) {
    write_guard.remove_id(&HomeCard::StartRadQuest);
}
```

sergiupuhalschi-rdx: This is very much simpler, thanks!

PR #182, opened 2024-07-08
Skip duplicate Gateways without crashing

matiasbzurovski:
## Changelog
- When deserializing `SavedGateways`, if JSON has elements with same Url we will just skip those that are repeated.
- Add function to check if `AppPreferences` have a `Gateway` with the given `Url`.

Both changes should unlock this user and prevent having issues like this in the future. [JIRA](https://radixdlt.atlassian.net/browse/ABW-3500)

### DIFF
--- a/src/profile/v100/app_preferences/gateways/saved_gateways.rs
+++ b/src/profile/v100/app_preferences/gateways/saved_gateways.rs
@@ -66,24 +66,34 @@ impl<'de> Deserialize<'de> for SavedGateways {
         struct Wrapper {
             #[serde(rename = "current")]
             url: Url,
-            saved: IdentifiedVecOf<Gateway>,
+            saved: Vec<Gateway>,
         }
         let wrapped = Wrapper::deserialize(deserializer)?;
         let current = wrapped
             .saved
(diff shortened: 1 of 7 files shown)


src/profile/v100/app_preferences/gateways/saved_gateways.rs
```diff
@@ -66,24 +66,34 @@ impl<'de> Deserialize<'de> for SavedGateways {
...
             .ok_or({
                 CommonError::InvalidGatewaysJSONCurrentNotFoundAmongstSaved
             })
             .map_err(de::Error::custom)?;
 
-        let mut other = wrapped.saved.clone();
+        let saved = wrapped.saved.clone();
+        let mut other = IdentifiedVecOf::<Gateway>::new();
+        for item in saved {
+            let _ = other.try_insert_unique(item.clone()).map_err(|e| {
```
GhenadieVP, line 84: The map_err seems superflous, rather I think you wanted to match on the result and log the error for debug purpose

GhenadieVP (approved):

PR #184, opened 2024-07-11
SargonOS v1 integration API

GhenadieVP:
Implements SargonOS V1:
- Jira + docs -> https://radixdlt.atlassian.net/browse/ABW-3591
- whimsical ref -> https://whimsical.com/sargon-os-integration-J8FN4ScciWkMAWqk9fSWFc@2bsEvpTYSt1Hj3YqMyWrCwCYuSAypqEEy24

This iOS Wallet PR - https://github.com/radixdlt/babylon-wallet-ios/pull/1234 - roughly showcases how the V1 integration would look like.

Note - this PR also contains the Android PR -> https://github.com/radixdlt/sargon/pull/196;
This [commit](https://github.com/radixdlt/sargon/pull/184/commits/498f627defb421a0f49c495ff68e10b19e77b2ed) shows the changes that were made to facilitate V1.


src/system/clients/client/profile_change_client/profile_change_client.rs
```diff
@@ -0,0 +1,18 @@
+use crate::prelude::*;
+
+#[derive(Debug)]
+pub struct ProfileChangeClient {
```
GhenadieVP, line 4: This is the v1 client that notifies the hosts that profile change, this will change as we get to the final version of SargonOS, where we would expose granular updates. But for the sake of easier migration - this is the first/intermediary step.

src/system/sargon_os/sargon_os_profile.rs
```diff
@@ -38,6 +38,24 @@ impl SargonOS {
 
 #[uniffi::export]
 impl SargonOS {
+    pub async fn set_profile(&self, profile: Profile) -> Result<()> {
```
GhenadieVP, line 41: This function will be called by hosts after they have mutated the Profile. Similar to the ProfileChangeClient, this is the intermediary step to facilitate the migration. In SargonOS V2 this function will not exist anymore, instead there will be specific APIs to mutate the profile directly in Sargon.

This intermediary step allows migrating Profile mutation to SargonOS piece by piece while still having the Wallet operational.

src/system/sargon_os/sargon_os_profile.rs
```diff
@@ -38,6 +38,24 @@ impl SargonOS {
...
+        if profile.id() != self.profile().id() {
+            return Err(
+                CommonError::TriedToUpdateProfileWithOneWithDifferentID,
+            );
+        }
+
+        self.update_profile_with(|mut p| {
+            *p = profile.clone();
```
GhenadieVP, line 49: Just replace the old profile.

src/system/sargon_os/sargon_os_profile.rs
```diff
@@ -269,17 +149,10 @@ impl SargonOS {
     /// # Emits
     /// Emits `Event::ProfileSaved` after having successfully written the JSON
     /// of the active profile to secure storage.
-    pub(crate) async fn maybe_validate_ownership_update_profile_with<F, R>(
-        &self,
-        validate_ownership: bool, // should only ever pass `false` from `claim`
```
GhenadieVP, line 274: Ownership is not validated on profile save, it is rather validated on the Wallet side when uploading the backup or after doing periodic cloud check

src/system/drivers/profile_change_driver/profile_change_driver.rs
```diff
@@ -0,0 +1,7 @@
+use crate::prelude::*;
+
+#[uniffi::export(with_foreign)]
+#[async_trait::async_trait]
+pub trait ProfileChangeDriver: Send + Sync + std::fmt::Debug {
+    async fn handle_profile_change(&self, changed_profile: Profile);
```
CyonAlexRDX, line 6: this is quite costly, sending the whole Profile. But it allows for simple implementation in hosts.

We should just be mindful that this might result in poor performance. 

An alternative would be, 6 drivers instead of one:
`FactorSourcesChangedDriver`
`AppPreferencesChangedDriver`
`GatewaysChangedDriver`
`AccountsChangedDriver`
`PersonasChangedDriver`
`AuthorizedDappsChangedDriver`

Which would be **drastically** smaller payloads sent. But perhaps not worth the complexity, especially since this is  all in flux and the Profile "API surface" will get smaller over time once more and more logic into Sargon.

Also, Profile payload being big is probably just a problem for Profiles with 10+ accounts which most user do not have.

src/system/drivers/profile_change_driver/support/test/rust_profile_change_driver.rs
```diff
@@ -0,0 +1,34 @@
+use crate::prelude::*;
```
CyonAlexRDX, line 1: can this be `#![cfg(test])` ? (or also used in integration test?)

src/system/sargon_os/sargon_os_profile.rs
```diff
@@ -725,11 +600,9 @@ mod tests {
         profile.networks.append(new_network.clone());
 
         // ACT
-        os.with_timeout(|x| {
-            x.DEPRECATED_save_ffi_changed_profile(profile.clone())
-        })
-        .await
-        .unwrap();
+        os.with_timeout(|x| x.set_profile(profile.clone()))
```
CyonAlexRDX, line 570: rename test to -> `test_set_profile` ?

src/system/sargon_os/sargon_os_profile.rs
```diff
@@ -742,11 +615,7 @@ mod tests {
         let os = SUT::fast_boot().await;
 
         // ACT
-        let res = os
-            .with_timeout(|x| {
-                x.DEPRECATED_save_ffi_changed_profile(Profile::sample())
-            })
-            .await;
+        let res = os.with_timeout(|x| x.set_profile(Profile::sample())).await;
```
CyonAlexRDX, line 585: rename test -> `test_set_profile_is_err_when_different_profile_id`

src/system/sargon_os/sargon_os_profile.rs
```diff
@@ -161,6 +127,10 @@ impl SargonOS {
         Ok(profile_id)
     }
 
+    pub fn profile(&self) -> Profile {
```
CyonAlexRDX, line 126: preserve comment:

```rust
    /// Returns the current profile in full. This is a COSTLY operation
    /// and hosts SHOULD NOT do it lightheartedly, prefer using more specific
    /// reading operations such as `os.current_network_id` or `os.accounts_for_display_on_current_network` etc, which are cheap operations compared
    /// to using this.
    ///
    /// In the future will will most likely deprecate this method.
```
? 

I think it important to keep it documented that this is in fact a very costly UniFFI call.

apple/Sources/Sargon/Drivers/ProfileChange/ProfileChange.swift
```diff
@@ -0,0 +1,52 @@
...
+extension ProfileChangeDriver where Self == ProfileChangeBus {
+
+    public static var shared: Self { Self.shared }
+}
+
+/// An `EventBusDriver` actor which handles incoming
+/// `EventNotifications` and forwards them to any
+/// subscriber of `notifications()`, being a multicasted
+/// async sequence.
+public final actor ProfileChangeBus {
```
CyonAlexRDX, line 16: `Bus` alone does not make sense IMO. `Event Bus` is a common term, but it is most commonly used when it is a singleton for many different kinds of events, which is not the case here.

Maybe `ProfileChangeNotifier` or `ProfileChangeEventPublisher` ?

Also update incorrect copy paste errors in documentation. And hmm, maybe we want a single shape for this? A generic? Seing as this type is almost identical to `actor EventBus` (which at least in commit 498f627defb421a0f49c495ff68e10b19e77b2ed was not removed ).

and then you can:

```swift
public final actor SargonEventPublisher<Element> { ... }
extension SargonEventPublisher:  ProfileChangeDriver  where Element ==  Profile {
     public func handleProfileChange(changedProfile: Profile) async {
        subject.send(changedProfile)
    }
}

```
and then you can have a single actor?

CyonAlexRDX (changes requested):
A single actor Swift side is proabably preferable:

https://github.com/radixdlt/sargon/pull/184/commits/498f627defb421a0f49c495ff68e10b19e77b2ed#r1711241760

src/system/drivers/profile_change_driver/profile_change_driver.rs
GhenadieVP: I don't think this applies here, the profile change is async. So what would happen is that the async operation of modifying a profile will take slightly longer, it should not impact the overall performance while using the app.

And as mentioned, **this is the first step**, going straight away to the final solution is too costly from the development perspective.

src/system/sargon_os/sargon_os_profile.rs
GhenadieVP: Sure, this will go away completely anyway with ProfileState that will be introduced by @micbakos-rdx.

apple/Sources/Sargon/Drivers/ProfileChange/ProfileChange.swift
GhenadieVP: I would rename it, will see about the generic suggestion.

Sajjon: if you try the generic approach, I think you will indeed see that they have a 95% identical shape.

CyonAlexRDX (approved):
LGTM

PR #186, opened 2024-07-15
Add function to clear home cards

sergiupuhalschi-rdx:
This PR adds a `uniffi` exported function to notify the `HomeCardsManager` to clear the home cards both in-memory and from storage. This is required when factory resetting the wallet and ensures that all the control over the home cards is inside Sargon. Thanks, @matiasbzurovski for the suggestion.

### DIFF
--- a/src/wallet/home_cards/manager.rs
+++ b/src/wallet/home_cards/manager.rs
@@ -110,6 +110,18 @@ impl HomeCardsManager {
             .await?;
         self.save_cards(updated_cards).await
     }
+
+    /// Clears the home cards from the `HomeCardsStorage`.
+    /// Notifies `HomeCardsObserver`.
+    #[uniffi::method]
+    pub async fn wallet_reset(&self) -> Result<()> {
+        let updated_cards = self
+            .update_cards(|write_guard| {
+                **write_guard = HomeCards::new();
+            })
+            .await?;
+        self.save_cards(updated_cards).await
+    }
 }
 
 impl HomeCardsManager {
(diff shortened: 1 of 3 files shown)


src/wallet/home_cards/manager.rs
```diff
@@ -110,6 +110,18 @@ impl HomeCardsManager {
             .await?;
         self.save_cards(updated_cards).await
     }
+
+    /// Clears the home cards from the `HomeCardsStorage`.
+    /// Notifies `HomeCardsObserver`.
+    #[uniffi::method]
+    pub async fn wallet_reset(&self) -> Result<()> {
```
matiasbzurovski, line 117: I would name it `wallet_resetted` to follow the past tense used for other actions (`wallet_created`, `link_received`, etc)

src/wallet/home_cards/manager.rs
```diff
@@ -474,4 +486,27 @@ mod tests {
...
+        let manager = SUT::new(
+            Arc::new(MockAntenna::new_always_failing()),
+            NetworkID::Stokenet,
+            Arc::new(MockHomeCardsStorage::new_with_stored_cards(
+                initial_cards.clone(),
+            )),
+            observer.clone(),
+        );
+
+        manager.bootstrap().await.unwrap();
```
matiasbzurovski, line 506: I think we should remove this `bootstrap()` call to ensure the modification takes place on `wallet_resetted`

matiasbzurovski (approved):
Thanks for taking care of it, small suggestions only

src/wallet/home_cards/manager.rs
sergiupuhalschi-rdx: I think `reset` is the correct past participle form. [Here](https://twominenglish.com/resetted-or-reset/#:~:text=Instead%2C%20it%20maintains%20the%20same,as%20its%20base%20form%3A%20reset.) is a reference.

src/wallet/home_cards/manager.rs
sergiupuhalschi-rdx: But then the test wouldn't be complete as the manager wouldn't have any cards in memory. The test will pass even if the wallet_reset doesn't do anything, but calling `bootstrap()` loads the mocked cards into the `HomeCardsManager` memory so that the test makes sense.

matiasbzurovski: oh, I thought the test has the `initial_cards` initally set

PR #188, opened 2024-07-17
Clear home cards after restore

sergiupuhalschi-rdx:
Currently, whenever the wallet is restored there should be no home cards. 
The use case is as follows: 
The user clicks on a valid deep link -> installs the app -> the deep link is resolved -> home cards are saved to storage -> the user restores the wallet -> at this point we should ensure no home cards are stored.

### DIFF
--- a/src/wallet/home_cards/manager.rs
+++ b/src/wallet/home_cards/manager.rs
@@ -99,6 +99,19 @@ impl HomeCardsManager {
         self.save_cards(updated_cards).await
     }
 
+    /// Marks the wallet restoration.
+    /// Ensures only the expected `HomeCards` remain in `HomeCardsStorage` - currently none.
+    /// Notifies `HomeCardsObserver`.
+    #[uniffi::method]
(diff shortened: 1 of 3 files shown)


src/wallet/home_cards/manager.rs
```diff
@@ -99,6 +99,19 @@ impl HomeCardsManager {
...
+    /// Ensures only the expected `HomeCards` remain in `HomeCardsStorage` - currently none.
+    /// Notifies `HomeCardsObserver`.
+    #[uniffi::method]
+    pub async fn wallet_restored(&self) -> Result<()> {
+        let updated_cards = self
+            .update_cards(|write_guard| {
+                **write_guard = HomeCards::new();
+            })
+            .await?;
+        self.save_cards(updated_cards).await
```
matiasbzurovski, line 112: Should we put this in a private method that `wallet_restored` & `wallet_reset` call?

matiasbzurovski (approved):

GhenadieVP (approved):

PR #189, opened 2024-07-22
Implement HostInfo struct

micbakos-rdx:
This PR aims to decouple DeviceInfo which is stored in profile with host information got from the device in realtime. HostInfoDriver aims to construct host info. Previously it used to return `DeviceInfo` which includes optional information about the newer data like host app version, host os version and vendor. Also DeviceInfo still uses the old `description: String` which is not helping when it comes to naming factor sources with device and model names. Since I wanted to keep compatibility with Profile's DeviceInfo, I chose to use a new struct that directly describes the state of the host.

1. HostInfo replaced DeviceInfo in sargon and wherever a host information is used to create new factor sources or relevant data types.
2. When in need to compare profile device infos we just compare with id and do not perform equals in the whole DeviceInfo model. (Will attach a comment related to this)
3. Changed the logic on what needs to be stored on secure storage. Now in order to resolve the host information, we need to:
    1. Load if any stored device id along with its creation date. This is the `HostId`. If none exists, os generates a new one and stores it in secure storage
    2. Resolve the rest attributes of `HostInfo` on the fly and attach the host id from (i).
4. `HostInfo` can be converted to `DeviceInfo` in order for the latter to be stored in profile, but I didn't want to expose the opposite. Since only the OS can/should query the most recent host information and should not rely on older data stored on profile.


examples/android/src/main/java/com/radixdlt/sargon/android/MainActivity.kt
```diff
@@ -57,69 +56,69 @@ class MainActivity : ComponentActivity() {
 
 @Composable
 fun WalletContent(modifier: Modifier = Modifier, storage: SecureStorageDriver) {
-    var walletState: Wallet? by remember { mutableStateOf(null) }
```
micbakos-rdx, line 60: Didn't have time to make example app compatible with sargon os. I can do that if needed.

src/system/clients/client/host_info_client/host_info_client.rs
```diff
@@ -21,6 +21,30 @@ impl HostInfoClient {
         )
     }
 
+    pub async fn resolve_host_info(
+        &self,
+        stored_device_id: DeviceID,
+        stored_device_date: Timestamp,
```
micbakos-rdx, line 27: Related to bullet (3) and (4) of the description of this PR.

src/system/sargon_os/sargon_os.rs
```diff
@@ -160,7 +158,18 @@ impl SargonOS {
...
+            .resolve_host_info(device_info.id, device_info.date)
+            .await;
+
+        let new_device_info = host_info.clone().into();
+        if device_info != new_device_info {
+            debug!("Device info: has been updated {:?}", &new_device_info);
+            secure_storage.save_device_info(&new_device_info).await?;
+        }
+
+        Ok(host_info)
```
micbakos-rdx, line 172: Related to bullet (3) and (4) of the description of this PR.

src/system/sargon_os/sargon_os_profile.rs
```diff
@@ -11,7 +11,8 @@ impl SargonOS {
         profile: &mut Profile,
         device_info: DeviceInfo,
     ) -> bool {
-        let was_needed = profile.header.last_used_on_device != device_info;
+        let was_needed =
+            profile.header.last_used_on_device.id != device_info.id;
```
micbakos-rdx, line 15: Here we need to check against the IDs and not the whole object.

src/system/sargon_os/sargon_os_profile.rs
```diff
@@ -20,12 +21,12 @@ impl SargonOS {
     /// header
     pub async fn claim_profile(&self, profile: &mut Profile) -> Result<()> {
         debug!("Claiming profile, id: {}", &profile.id());
-        let device_info = self.device_info().await?;
-        Self::claim_provided_profile(profile, device_info.clone());
+        let host_info = self.host_info().await?;
+        Self::claim_provided_profile(profile, host_info.clone().into());
```
micbakos-rdx, line 25: Host info needs to be translated `into()` device info for claim logic that exists in profile. So we need to translate sargon os information with a language that profile understands (DeviceInfo).

src/system/sargon_os/sargon_os_profile.rs
```diff
@@ -119,12 +120,15 @@ impl SargonOS {
...
-        let device_info = self.device_info().await?;
+        let host_info = self.host_info().await?;
 
         self.maybe_validate_ownership_update_profile_with(
             false, // we do NOT validate ownership, since this method is claiming
             |mut p| {
-                Ok(Self::claim_provided_profile(&mut p, device_info.clone()))
+                Ok(Self::claim_provided_profile(
+                    &mut p,
+                    host_info.clone().into(),
```
micbakos-rdx, line 130: Same as above.

src/profile/supporting_types/host_info.rs
```diff
@@ -0,0 +1,125 @@
...
+
+    /// The vendor of the host client, e.g. "Apple" for iPhone clients,
+    /// or "Samsung" for Android clients.
+    pub host_vendor: String,
+}
+
+// HostInfo can be converted to DeviceInfo but not the opposite
+// without contacting HostInfoDriver
+#[allow(clippy::from_over_into)]
+impl Into<DeviceInfo> for HostInfo {
```
GhenadieVP, line 36: Usually this is done the other way around by implementing `From` trait, so this would be `impl From<HostInfo> for DeviceInfo` and placed in DeviceInfo file.

src/profile/supporting_types/host_info.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+impl HasSampleValues for HostInfo {
+    fn sample() -> Self {
+        Self {
+            id: DeviceID::from_str("66F07CA2-A9D9-49E5-8152-77ACA3D1DD74")
+                .unwrap(),
+            date: Timestamp::sample(),
+            description: DeviceInfoDescription {
```
GhenadieVP, line 56: I would expected DeviceInfoDescription to also implement HasSampleValues

src/profile/supporting_types/host_info.rs
```diff
@@ -0,0 +1,125 @@
...
+
+    /// The date this description of the device was made, might
+    /// be equal to when the app was first ever launched on the
+    /// device.
+    pub date: Timestamp,
+
+    /// A short description of the device, we devices should
+    /// read the device model and a given name from the device
+    /// if they are able to.
+    pub description: DeviceInfoDescription,
```
GhenadieVP, line 20: A bit unclear why this should be part of HostInfo, let's sync on this.

src/profile/supporting_types/host_info.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// A best effort stable and unique identifier of this
+    /// device.
+    pub id: DeviceID,
+
+    /// The date this description of the device was made, might
+    /// be equal to when the app was first ever launched on the
+    /// device.
+    pub date: Timestamp,
+
+    /// A short description of the device, we devices should
```
sergiupuhalschi-rdx, line 17: typo in "we devices"

src/profile/supporting_types/host_info.rs
micbakos-rdx: It does but does not follow the logic of one sample being for an iOS client and other for Android.

micbakos-rdx:
@GhenadieVP I have pushed the latest commits were 
1. The `HostId` is separate from `HostInfo`. Those two together can create an instance of `DeviceInfo`.
2. The host info driver, now can return a `HostOS` which is an enum with various properties according to the operating system.

GhenadieVP (approved):
Nice, well done!

PR #195, opened 2024-08-05
Add wrapper FfiUrl

matiasbzurovski:
## Description
Adds a wrapper around Rust `Url`, which has to be created on host sides. This way, if the conversion fails, the error is downcasted to host which will handle it appropiately (rather than having Rust code panic).

### DIFF
--- a/apple/Tests/TestCases/Profile/AppPreferencesTests.swift
+++ b/apple/Tests/TestCases/Profile/AppPreferencesTests.swift
@@ -17,17 +17,18 @@ final class AppPreferencesTests: Test<AppPreferences> {
 	}
 	
 	func test_has_gateway_valid() throws {
-		XCTAssertTrue(try SUT.default.hasGateway(with: .init(string: "https://mainnet.radixdlt.com/")!))
-		XCTAssertTrue(try SUT.default.hasGateway(with: .init(string: "https://mainnet.radixdlt.com")!))
-		XCTAssertFalse(try SUT.default.hasGateway(with: .init(string: "https://radixdlt.com")!))
-		
+		XCTAssertTrue(SUT.default.hasGateway(with: try .init(urlPath: "https://mainnet.radixdlt.com/")))
+		XCTAssertTrue(SUT.default.hasGateway(with: try .init(urlPath: "https://mainnet.radixdlt.com")))
+		XCTAssertFalse(SUT.default.hasGateway(with: try .init(urlPath: "https://radixdlt.com/")))
 	}
 	
 	func test_has_gateway_invalid() {
-		let url = URL(string: "Rust considers this invalid")
-		XCTAssertNotNil(url)
-		XCTAssertThrowsError(try SUT.default.hasGateway(with: url!)) { error in
-			XCTAssertEqual(error.localizedDescription, "Failed to convert arg \'url\': relative URL without a base")
+		let urlPath = "invalid input"
+		XCTAssertThrowsError(SUT.default.hasGateway(with: try .init(urlPath: urlPath))) { error in
+			guard let commonError = error as? SargonUniFFI.CommonError else {
+				return XCTFail("Expected CommonError")
+			}
+			XCTAssertEqual(commonError, .InvalidUrl(badValue: urlPath))
 		}
 	}
 }
--- a/apple/Tests/TestCases/Profile/AppPreferencesTests.swift
+++ b/apple/Tests/TestCases/Profile/AppPreferencesTests.swift
@@ -17,8 +17,26 @@ final class AppPreferencesTests: Test<AppPreferences> {
 	}
 	
 	func test_has_gateway() {
-		XCTAssertTrue(SUT.default.hasGateway(with: .init(string: "https://mainnet.radixdlt.com/")!))
-		XCTAssertTrue(SUT.default.hasGateway(with: .init(string: "https://mainnet.radixdlt.com")!))
-		XCTAssertFalse(SUT.default.hasGateway(with: .init(string: "https://radixdlt.com")!))
+		// Valid URLs
+		do {
+			var result = try SUT.default.hasGateway(with: .init(string: "https://mainnet.radixdlt.com/")!)
+			XCTAssertTrue(result)
+			result = try SUT.default.hasGateway(with: .init(string: "https://mainnet.radixdlt.com")!)
+			XCTAssertTrue(result)
+			result = try SUT.default.hasGateway(with: .init(string: "https://radixdlt.com")!)
+			XCTAssertFalse(result)
+		} catch {
+			XCTFail("Unexpected failure")
+		}
+		
+		// Invalid URL
+		do {
+			let url = URL(string: "Rust considers this invalid")
+			XCTAssertNotNil(url)
+			let _ = try SUT.default.hasGateway(with: url!)
+			XCTFail("Should have thrown")
+		} catch let error {
+			XCTAssertEqual(error.localizedDescription, "Failed to convert arg \'url\': relative URL without a base")
+		}
 	}
 }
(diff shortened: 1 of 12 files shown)


src/types/uniffi_url.rs
```diff
@@ -0,0 +1,15 @@
+use crate::prelude::*;
+
+#[derive(uniffi::Object)]
```
CyonAlexRDX, line 3: Should probably be:

```rust
#[derive(Debug, PartialEq, Eq, Hash, uniffi::Object)]
#[uniffi::export(Debug, Display, Eq, Hash)]
pub struct FfiUrl {
    url: url::Url,
}

```

Using `UniFFI::export(Eq)` UniFfi will make the resulting Swift Class conform to Swift Equatable, and analogously for Hash etc, and also for Kotlin

apple/Tests/TestCases/Profile/AppPreferencesTests.swift
```diff
@@ -17,8 +17,26 @@ final class AppPreferencesTests: Test<AppPreferences> {
 	}
 	
 	func test_has_gateway() {
-		XCTAssertTrue(SUT.default.hasGateway(with: .init(string: "https://mainnet.radixdlt.com/")!))
-		XCTAssertTrue(SUT.default.hasGateway(with: .init(string: "https://mainnet.radixdlt.com")!))
-		XCTAssertFalse(SUT.default.hasGateway(with: .init(string: "https://radixdlt.com")!))
+		// Valid URLs
+		do {
```
CyonAlexRDX, line 21: for expected NON-throws I prefer changing the test method to be throwing, and remove the `do {} catch {}`.

apple/Tests/TestCases/Profile/AppPreferencesTests.swift
```diff
@@ -17,8 +17,26 @@ final class AppPreferencesTests: Test<AppPreferences> {
...
+			result = try SUT.default.hasGateway(with: .init(string: "https://mainnet.radixdlt.com")!)
+			XCTAssertTrue(result)
+			result = try SUT.default.hasGateway(with: .init(string: "https://radixdlt.com")!)
+			XCTAssertFalse(result)
+		} catch {
+			XCTFail("Unexpected failure")
+		}
+		
+		// Invalid URL
+		do {
```
CyonAlexRDX, line 33: and I would prefer if this expected throws to be its own test, and use `XCTAssertThrows`

CyonAlexRDX, line 33: and I would prefer if this expected throws to be its own test, and use `XCTAssertThrows`

CyonAlexRDX, line 33: and I would prefer if this expected throws to be its own test, and use `XCTAssertThrows`

src/types/ffi_url.rs
```diff
@@ -0,0 +1,42 @@
...
+/// the code will panic. However, if we send the wrapper instead, we make sure the conversion is safely done on the
+/// host side, dealing with the failing conversion properly rather than panicking.
+#[derive(Debug, PartialEq, Eq, Hash, uniffi::Object, derive_more::Display)]
+#[uniffi::export(Debug, Display, Eq, Hash)]
+pub struct FfiUrl {
+    pub url: Url,
+}
+
+#[uniffi::export]
+impl FfiUrl {
```
CyonAlexRDX, line 22: can probably make this `impl FromStr for FfiUrl` and do `Url::parse(s).map_err(..).map(SUT::new)` ? which results in a bit nicer tests both here and in `src/profile/v100/app_preferences/app_preferences_uniffi_fn.rs`

matiasbzurovski: Added it to clean up the `app_preferences_uniffi_fn.rs` test, but we still want to send a `Url` to this file test as we compare both `Url` match.

Therefore, by adding it to help clean up one test, I ended up needing to add an extra test on this file (to cover a function that is only used in tests). Wonder if that actually helps or adds more noise than before.

CyonAlexRDX: Your call :) but true less is
More. I tend to impl too many traits 😑

src/types/ffi_url.rs
```diff
@@ -0,0 +1,57 @@
+use uniffi::deps::bytes::Buf;
```
GhenadieVP, line 1: Should not be needed anymore

src/types/ffi_url.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use crate::prelude::*;
+use crate::UniFfiTag;
```
GhenadieVP, line 4: Should not be needed

src/types/ffi_url.rs
```diff
@@ -0,0 +1,57 @@
...
+#[derive(Debug, PartialEq, Eq, Hash, uniffi::Object, derive_more::Display)]
+#[uniffi::export(Debug, Display, Eq, Hash)]
+pub struct FfiUrl {
+    pub url: Url,
+}
+
+#[uniffi::export]
+impl FfiUrl {
+    #[uniffi::constructor]
+    pub fn new(url: Url) -> Result<Self> {
```
GhenadieVP, line 24: Do we really want uniffi intialiser through URL? We are again hiding the information here as to why this constructor is fallible.
Why not initialise from String?

src/types/ffi_url.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl FromStr for FfiUrl {
+    type Err = CommonError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let url = parse_url(s)?;
+        Self::new(url)
+    }
+}
```
GhenadieVP, line 36: Is it this only for testing?

src/types/ffi_url.rs
CyonAlexRDX: I reacted to that too. Right, we should use `String`, and then we **can** use `FromStr` impl in this `#[uniffi::constructor]`.

CyonAlexRDX: I reacted to that too. Right, we should use `String`, and then we **can** use `FromStr` impl in this `#[uniffi::constructor]`.

src/types/ffi_url.rs
matiasbzurovski: Yes, gonna remove it as per discussion [here](https://github.com/radixdlt/sargon/pull/195#discussion_r1705285262)

src/types/ffi_url.rs
matiasbzurovski: I don't think we are "hiding" the reason of the failure, we are just exposing -with the wrapper- that the conversion from `Swift.URL` to `Rust.Url` could fail. Initialising from the String seems unnecessary to me, as we are adding the same logic that we define on [lib.rs](https://github.com/radixdlt/sargon/blob/main/src/lib.rs#L263) but in two places. 
By having it as is, we make sure that the wrapper uses the same logic defined for `UniffiCustomTypeConverter`, with the addition of handling the error on host side.

GhenadieVP: This `FfiUrl` should ideally just forward the underlying `Rust.Url` API, so the constructor should be the primary `Rust.Url` constructor.
as for the same logic with the `UniffiCustomTypeConverter`, well you save one function call - `Url::parse`.
IMO, I would opt for clarity here rather than for optimising one line of code.

matiasbzurovski: My reasoning was not because of the extra line of code (in fact, it is going to be one less due to changes on Swift side), but rather to keep the conversion logic on just one place.

GhenadieVP (approved):

PR #196, opened 2024-08-06
Sargon os v1 android

micbakos-rdx:
Introduces all the implementations of drivers needed to boot sargon os on Android.

More details about the drivers
* `AndroidProfileChangeDriver`: Exposes a shared flow for the host app to subscribe to.
* `AndroidFileSystemDriver`: Allows read/write to files stored in app internal storage in `${context.filesDir}/sargon/`
* `AndroidEventBusDriver`: exposes a similar flow as `AndroidProfileChangeDriver`
* `AndroidLoggingDriver`: uses [Timber](https://github.com/JakeWharton/timber) to print logs. Logs can be enabled or not when instantiating the `Bios`.
* `AndroidHostInfoDriver`: implements all the logic that retrieves host information and also reads stored `HostId`.
* `AndroidEntropyProviderDriver`: provides entropy by using android's [SecureRandom](https://developer.android.com/reference/kotlin/java/security/SecureRandom)
* `AndroidStorageDriver`: is a driver that provides access to key/value storage and encrypted or not.
* `AndroidNetworkingDriver`: is the driver that makes the network requests by accepting an `OkHttpClient`.
* `AndroidBiometricAuthorizationDriver` uses `BiometricsHandler` to ask for biometrics.

### `AndroidStorageDriver`
This driver took the most time to develop. Even though sargon exposes an API to read/write to a key/value storage, the contents of the android storage are usually differently stored, like different schema or different data types. For example
    * profile snapshot is encrypted but kept as a string
    * `HostId` used to be stored as `DeviceInfo` so some migration logic should be done. Also host Id used to be stored in `SharedPreferences` and sargon os also provides a test that migrates shared preferences to `Datastore<Preferences>`.
    * `MnemonicWithPassphrase` has different serialization logic.

### `BiometricsHandler`
This is an android specific class which is responsible for asking for biometrics when secure storage needs to read/write data. This is a key difference between iOS and Android. On Android we are not using a storage that manages its own encryption, but we simply encrypt the data and store them to a file as we would store data in any other file that are not encrypted. In order to encrypt securely we use Android's [Keystore System](https://developer.android.com/privacy-and-security/keystore). This keystore manages the encryption outside of the app's sandbox. Some data are tied to keys that not need authorization (like profile snapshot), but other, like mnemonic, need authorisation. We are responsible to chose when and were to ask for biometrics before decrypt/encrypt a mnemonic. So in order to bridge the difference between iOS and Android, whenever a secure storage key requires authorisation, sargon uses `BiometricsHandler` in order to handle exactly that. 

In order to ask for biometrics on Android we need to have access to a `FragmentActivity`. That means that such an activity must have been created in order for us to request biometrics. Since sargon os is booted in application level, were no activity has yet been created, we need to ensure that in some cases sargon may ask the host for biometrics before the first activity is registered. Currently this is possible when the app is booting sargon os for the first time when no profile is available in storage. Immediately sargon creates an ephemeral profile with an ephemeral mnemonic that needs to be stored. This whole operation is quicker that the first activity created. So sargon os needs to suspend until some activity is available in order to handle the biometrics request. Also this handler needs to handle the case were an activity might stop (moved to background) or a new activity is created on top (Current android wallet has only one activity that handles all the screens, but I tried to make it not rely on that assumption).  In order to solve this problem `BiometricsHandler` should be a singleton injected to any activity that can handle biometrics. Such activity should call `register()` in their respective `onCreate`. Inside this class we have two channels communicating together.
* `biometricRequestsChannel` which suspends until an activity is subscribed to biometrics and becomes `Lifecycle.State.ON_CREATE`.
* `biometricsResultsChannel` which waits until the user provides the biometrics, or the system fails for some reason to ask for biometrics (maybe the user did not enrol yet)
These two are combined together in order to expose a simple function `suspend askForBiometrics(): Result<Unit>` which can be called from sargon.

### Android tests
I tried to mock as many classes I could, classes that touch the android system. But there are some tests that would be more reliable to be tested under instrumentation tests. These tests cover all the drivers that need to communicate with the system. In order to run such tests we need a device or an emulator. Such tests run with `./gradlew sargon-android:connectedAndroidTest`. Our CI does not facilitate yet a connection to any device. A good approach here is to make our CI pipeline run these tests some times every day on main branch by connecting to firebase test lab. This can be done in a separate PR.

### Changes in sargon
In sargon I had to do the following changes, in order to successfully boot the sargon os.
1. One of them is #189 . This separated the `HostInfo` from `HostId` and makes it possible for android to provide reliably all the data needed.
2. Sargon os used to operate on the following logic: When booting, ask the storage on what profile id is the active profile id and then if such id exists, load that profile from storage. This is due to the fact that the iOS host used to store multiple profiles but mark what profile is "in use" due to the old backup logic. On android the host only stored the currently used profile. So in order to bridge the gap, sargon os needs to load the currently used profile without any knowledge on what is perceived as active. Android will just load a profile if exists, and iOS needs to perform this active profile id logic in its drivers.
  

### Host integration.
I played around a bit by booting sargon os in the sample application. The host needs to do the following:

* In application scope boot the bios. Take a look at example app's `SargonOsManager`. This works a bit similarly with how android host loads the profile. This will ease the migration of ProfileRepository and sargon os.
* In each activity that can handle biometrics (currently on Android wallet's `MainActivity`) we need to add the following:
```
@AndroidEntryPoint
class MainActivity : FragmentActivity() {
    @Inject
    lateinit var biometricsHandler: BiometricsHandler
    
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)

        biometricsHandler.register(this)
        // ...
    }
}
```
And that's all about it.


CyonAlexRDX (approved):
Nice, did not look into Kotlin code, just read description of PR and Rust changes, look good!

We should probably do a PR into this PR source branch with updates to Swift Sargon / iOS example app! We should decide who does that, if you want wanna do it @micbakos-rdx  ? Should be straightforward I think.

micbakos-rdx:
> Nice, did not look into Kotlin code, just read description of PR and Rust changes, look good!
> 
> We should probably do a PR into this PR source branch with updates to Swift Sargon / iOS example app! We should decide who does that, if you want wanna do it @micbakos-rdx ? Should be straightforward I think.

Happy to do that myself. I need to add more tests which downgrade the coverage. These tests are a bit difficult to emulate. Also I would like us to decide on some things related to booting in today's design session, before jumping into the ios implementation.

jvm/sargon-android/src/main/java/com/radixdlt/sargon/os/driver/AndroidStorageDriver.kt
sergiupuhalschi-rdx, line 1: `SecureStorageDriver` and `UnsafeStorageDriver` could be implemented by 2 separate classes for better separation of concerns, or maybe there are some advantages to keep them together?

sergiupuhalschi-rdx (commented):
@micbakos-rdx Outstanding work! Very well thought out and structured.

PR #197, opened 2024-08-13
Sargon os v1 bootstrap

micbakos-rdx:
* OS can boot without the existence of a profile
* Profile now has states:
    * None: No Profile stored
    * Incompatible: Available stored profile failed to load
    * Loaded: Profile loaded
* The profile change driver has been emitting updates of profile state and not just profile. The host can now subscribe to profile state changes and in turn, to profile changes.
* Some sargon os methods changed/created:
    1. `new_wallet`: Responsible for creating a bdfs and a new profile, and storing them
    2. `import_wallet`: For now it accepts a profile and the option to create a new main bdfs. The reason for those is to bridge integration with current hosts that already parse the imported profile and enforce the main bdfs creation, if needed, in sargon os.
    3. `delete_wallet`: Deletes the profile and all related mnemonics, without creating a new. The profile state is set to None
    4. `set_profile`: Used to update the profile from the host and to propagate the change back to the host.
* `boot` function does not require a profile in order to complete successfully. It tries to load any profile if such exists in secure storage. Remains `async`.
* The `profile()` method returns a `Result<Profile>`. For all methods that use it, if such profile is not yet loaded, then an early return of the error `ProfileStateNotLoaded` is occurred. 
* From our conversations with @GhenadieVP, the `try_read` lock is acquired or a panic will throw. This was a conscious decision, since there should not be any read/write concurrent access into profile, in v1. So I didn't change them to return an error but kept the `expect` function as is. Open to suggestions.


sergiupuhalschi-rdx (approved):

CyonAlexRDX (approved):
LGTM! Good job!

PR #198, opened 2024-08-14
Expose host info to host

micbakos-rdx:
This seems weird. Why host needs information about itself and retrieve it from sargon. 

For compatibility reasons, in order for sargon to be able to retrieve the host id which is stored into android's preferences, we had to put a reasonable amount of code in order to map what was stored on Android to what sargon os needs. So there is a compatibility layer for specifically HostId key. This is written already [here](https://github.com/radixdlt/sargon/blob/main/jvm/sargon-android/src/main/java/com/radixdlt/sargon/os/storage/key/HostIdKeyMapping.kt).

In order to avoid duplication of such logic, it was wise to expose the methods that retrieve host information, instead of keeping the same code natively on Android wallet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2585,7 +2585,7 @@ dependencies = [
 
 [[package]]
 name = "sargon"
-version = "1.1.4"
+version = "1.1.5"
 dependencies = [
  "actix-rt",
  "aes-gcm",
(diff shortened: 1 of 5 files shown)


CyonAlexRDX (approved):
LGTM. 

Yes until we have all logic in Sargon there will be a bit "back and forth". Seemingly strange APIs.

We just have to comfort ourselves that it will go away. And we will have a clean and small API "surface" in the future.

PR #199, opened 2024-08-14
[ABW-3712] Allow to hide/unhide resources on Profile

matiasbzurovski:
## Changelog
- Adds a new `ResourcePreferences` (stored under `Profile.AppPreferences`) that will allow user to hide fungible tokens, non-fungible tokens & pool units. [Zeplin](https://app.zeplin.io/project/639c7e7ab391bb15ac5d25fa/screen/66bb59e46fb28976587354aa)
- Extends `Security` to store new flag to indicate if advanced lock is required.


src/profile/v100/app_preferences/resource_preferences/resource_preferences.rs
```diff
@@ -0,0 +1,268 @@
...
+    Clone,
+    derive_more::Display,
+    uniffi::Record,
+)]
+#[serde(rename_all = "camelCase")]
+#[display("{}", self.description())]
+pub struct ResourcePreferences {
+    /// A dictionary detailing the user preferences for fungible resources.
+    #[serde(default)]
+    pub fungible: HashMap<ResourceAddress, EntityFlags>,
```
CyonAlexRDX, line 22: I think `fungibles`, plural, no? It is a HashMap with settings for fungibles? (Or strictly: fungible assets plural)

src/profile/v100/app_preferences/resource_preferences/resource_preferences.rs
```diff
@@ -0,0 +1,268 @@
...
+                EntityFlags::sample_other(),
+            )]
+            .into(),
+        }
+    }
+}
+
+impl ResourcePreferences {
+    pub fn get_hidden_resources(&self) -> HiddenResources {
+        let fungible = self
```
CyonAlexRDX, line 108: You can probably use a closure unlined here, being a function you reuse for the three collections (Rust does not allow nested functions, but you can use closure).

src/profile/v100/app_preferences/resource_preferences/resource_preferences.rs
CyonAlexRDX, line 22: You can use IndexMap to retain insertion order!

src/profile/v100/app_preferences/security.rs
```diff
@@ -16,14 +16,18 @@ use crate::prelude::*;
...
 #[display(
-    "cloud? {}, dev? {}",
+    "cloud? {}, dev? {}, advanced lock? {}",
     is_cloud_profile_sync_enabled,
-    is_developer_mode_enabled
+    is_developer_mode_enabled,
+    is_advanced_lock_enabled
 )]
 pub struct Security {
     pub is_cloud_profile_sync_enabled: bool,
```
CyonAlexRDX, line 25: I would probably start using new types around bools now, easy to accidentally pass the wrong bool to the wrong value otherwise

CyonAlexRDX (changes requested):
Good start!

src/profile/v100/app_preferences/resource_preferences/resource_preference_kind.rs
```diff
@@ -0,0 +1,10 @@
+use crate::prelude::*;
+
+#[derive(
+    Clone, Debug, PartialEq, Eq, Hash, derive_more::Display, uniffi::Enum,
+)]
+pub enum ResourcePreferenceKind {
```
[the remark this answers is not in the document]
Sajjon: maybe `enum AssetAddress` ?

src/profile/v100/app_preferences/resource_preferences/resource_preferences.rs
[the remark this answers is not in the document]
Sajjon: so it would be something like:

```rust
pub enum AssetVisibility {
    Hidden, 
    // Just an example of a potential future variant
    ConditionallyHiddenIfValueIsLessThan1000Usd
}
pub struct AssetPreference {
    address: AssetAddress, // fungible | nonFungible | poolUnit

    /// This does NOT allow for other flags, it is a mutually exclusive state.
    visibility: AssetVisibility
}

struct AssetPreference {
   assets: IndexMap<AssetAddress, AssetPreference>
}
```

Advantages:
* This retains insertion order of all preferences, "globally" (if ever interesting). 
* Is more compact
* If we want preferences for more than visibility, we add a new enum for that, and a new field "next to" `visibility`.

This is also future proof, since we can always add more enum variants to `AssetVisibility` and we can alway add new enum types and **fiels** to `AssetPreference`. So... should work?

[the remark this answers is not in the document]
matiasbzurovski: Agree that maybe `EntityFlags` aren't a great representation of "flags for an address". However, the reason I used them, is because if we ever extend the configuration of what is defined for entities, it probably applies for assets as well. 

Of course, this is all based on assumptions that we don't know yet how will they turn out. In your `ConditionallyHiddenIfValueIsLessThan1000Usd`, then it definitely makes sense to have the new resource configuration using a new preference definition (as the one you suggested). On my case, I thought a good example could have been "marking a resource/persona/account as favorite". In such case, a new `EntityFlag::IsFavorite` would fit for all of them.

src/profile/v100/app_preferences/security.rs
matiasbzurovski: sorry, couldn't get what you meant here

src/profile/v100/app_preferences/resource_preferences/resource_preferences.rs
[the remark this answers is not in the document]
Sajjon: Shared variants is not the same as semantically the same enum :) they should be able to evolve independently of each other. Also `DeletedByUser` is the intention of account, they dont want it. Initially we gave the button the name Delete, then changed to Hide. Whereas "Hide asset" feature truely is a hide, since it is stored in the immutable Ledger - thus never can be deleted! So it should be named `WhatIsTheNameOfTheEnum::Hidden` and not `::Deleted`.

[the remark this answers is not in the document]
matiasbzurovski: but with that argument hidden accounts/personas should also be named `::Hidden`since they aren't deleted from Ledger

src/profile/v100/app_preferences/security.rs
Sajjon: I mean we should change from:

```rust
pub struct Security {
   ...
    pub is_cloud_profile_sync_enabled: bool,
    pub is_developer_mode_enabled: bool,
    pub is_advanced_lock_enabled: bool,
...
}
```

To:

```rust
#[derive(..., )]
#[serde(transparent)]
pub struct IsCloudSyncEnabled(bool);

#[derive(..., )]
#[serde(transparent)]
pub struct IsDeveloperModeEnabled(bool);
pub struct IsAdvancedLockEnabled(bool);

uniffi::custom_newtype!(IsCloudSyncEnabled, bool);
uniffi::custom_newtype!(IsDeveloperModeEnabled, bool);
...

pub struct Security {
   ...
    pub is_cloud_profile_sync_enabled: IsCloudSyncEnabled,
    pub is_developer_mode_enabled: IsDeveloperModeEnabled,
    pub is_advanced_lock_enabled: IsAdvancedLockEnabled,
...
}
```

Making everything type safe :) 

(we seems suitable to use a macro for it perhaps! `decl newtype_bool!(IsDeveloperModeEnabled); `...

src/profile/v100/app_preferences/resource_preferences/resource_preference_kind.rs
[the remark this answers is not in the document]
matiasbzurovski: there is no such thing as a `FungibleAddress` (so it would represent fungibles as `resource`), nor we want a `NonFungibleResourceAddress` (we want `NonFungibleGlobalId`), so not sure this address_union! macro helps us here.

src/profile/v100/app_preferences/asset_preferences/asset_preferences.rs
```diff
@@ -0,0 +1,162 @@
...
+    Eq,
+    Clone,
+    derive_more::Display,
+    uniffi::Record,
+)]
+#[serde(rename_all = "camelCase")]
+#[display("assets: {:#?}", self.assets)]
+pub struct AssetPreferences {
+    #[serde(default)]
+    pub assets: HashMap<AssetAddress, AssetPreference>,
```
CyonAlexRDX, line 21: should be `IndexMap` else we loose insertion order.

Otherwise might be worth doing like we do with most collection: let `AssetAddress` impl `Identifiable` and use the macro to declare a collection of `AssetPreferences` plural. And use that here.

src/profile/v100/app_preferences/asset_preferences/asset_preferences.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+impl Hash for AssetPreferences {
+    fn hash<H: Hasher>(&self, state: &mut H) {
```
CyonAlexRDX, line 25: and using the `IdentifiedVec` will allow you to derive `Hash` instead of manually  implementing it.

src/profile/v100/app_preferences/asset_preferences/asset_preferences.rs
CyonAlexRDX: oh right this whole thing should probably be `decl_identified_vec_of(AssetPreference)` which creates a `AssetPreferences` which is serde transparent, so it will be `[]` in JSON. Which is pretty nice.

Sajjon (commented):
Nice, getting there! I would add doc an all new types and pub methods

src/profile/v100/app_preferences/asset_preferences/asset_preferences.rs
```diff
@@ -29,17 +29,17 @@ impl AssetPreferences {
     }
 
     pub fn hide_asset(&mut self, asset: AssetAddress) {
-        if !self.update_with(asset.id(), |x| {
-            x.set_visibility(AssetVisibility::Hidden)
-        }) {
+        if !self
+            .update_with(asset.id(), |x| x.visibility = AssetVisibility::Hidden)
```
CyonAlexRDX, line 33: I **THINK** you can use write `update_with(asset, ..` ?

fixtures/vector/only_plaintext_profile_snapshot_version_100.json
```diff
@@ -46,6 +46,13 @@
         },
         "transaction": {
             "defaultDepositGuarantee": "0.975"
+        },
+        "resource": {
```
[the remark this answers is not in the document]
matiasbzurovski: good catch! actually I am gonna revert this change since this JSON isn't used to validate the `AssetsPreferences` and its corresponding test can deal without any `assets` entry on the JSON.

src/profile/v100/app_preferences/asset_preferences/asset_preferences.rs
CyonAlexRDX: Update with takes an impl Borrow. I thought it auto borrowed, but im probably mistaken

CyonAlexRDX (approved):
LGTM! Good job! Congratulations on your first macro!

PR #201, opened 2024-08-23
Hiding/unhiding resource updates

matiasbzurovski:
## Changelog
We don't want to allow hiding a particular NFT (identified by a `NonFungibleGlobalId`), instead we should allow hiding its collection (identified by `ResourceAddress`). 

Then, a Profile no longer has `app_preferences.assets` configuration, but instead it now has `app_preferences.resources`. Such renaming is applied to everything that used to be under `app_preferences/asset` mod, and now stands below `app_preferences/resource` mode. Also, given there is already a struct `ResourcePreference` inside RET's module, the one regarding App Preferences configuration is named `ResourceAppPreference`.


sergiupuhalschi-rdx (approved):

src/profile/v100/app_preferences/resource_preferences/resource_identifier.rs
```diff
@@ -16,38 +16,26 @@ use crate::prelude::*;
 )]
 #[serde(tag = "kind", content = "value")]
 #[serde(rename_all = "camelCase")]
-pub enum AssetAddress {
+pub enum ResourceIdentifier {
```
Sajjon, line 19: now this is an address! So now we can make use of the macro, right?

CyonAlexRDX (approved):
LGTM. We can try to revisit using the address_union! Macro later

PR #203, opened 2024-08-27
Updating the scrypto and ret dependency

0xOmarA:


CyonAlexRDX (approved):
Thx! I realise my mistake, I used `git = "...", version = "1.2.0" }` instead of just version.

PR #205, opened 2024-08-28
Build manifest for claiming the contents of account lockers

sergiupuhalschi-rdx:
[Ticket](https://radixdlt.atlassian.net/browse/ABW-3728)

- Added new LockerAddress entity type
- Extended manifest building with account locker claiming functionality

[ABW-3728]: https://radixdlt.atlassian.net/browse/ABW-3728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/sargon/src/profile/v100/address/locker_address.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// e.g.:
+    /// `"locker_rdx1tz474x29nxxd4k2p2reete9xyz4apawv63dphxkr00qt23vyju49fq"`
+    ///
+    /// A `LockerAddress` has the [Scrypto's `EntityType`][entt] `GlobalAccountLocker`.
+    ///
+    /// Implementation wise we wrap [Radix Engine Toolkit's `CanonicalLockerAddress`][ret], and
+    /// give it UniFFI support, as a `uniffi::Record` (we also own Serde).
+    ///
+    /// [entt]: https://github.com/radixdlt/radix-engine-toolkit/blob/476d779fee08ed1e561ac8fc8730a2a404b7de79/crates/radix-engine-toolkit-uniffi/src/common/entity_type.rs
+    /// [ret]: https://github.com/radixdlt/radix-engine-toolkit/blob/476d779fee08ed1e561ac8fc8730a2a404b7de79/crates/radix-engine-toolkit/src/models/canonical_address_types.rs#L262-L265
```
CyonAlexRDX, line 16: Love that you made sure to update to correct links, thx!

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/account_locker_claimable_resource.rs
```diff
@@ -0,0 +1,51 @@
+use crate::prelude::*;
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, uniffi::Enum)]
+pub enum AccountLockerClaimableResource {
```
CyonAlexRDX, line 4: missing doc on enum and cases

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/manifest_account_locker.rs
```diff
@@ -0,0 +1,151 @@
...
+impl From<AccountAddress> for ScryptoComponentAddress {
+    fn from(value: AccountAddress) -> ScryptoComponentAddress {
+        ScryptoComponentAddress::new_or_panic(value.node_id().0)
+    }
+}
+
+impl TransactionManifest {
+    pub fn account_locker_claim(
+        locker_address: &LockerAddress,
+        claimant: &AccountAddress,
```
CyonAlexRDX, line 15: nice clean API, good job! Feels great to use actual `LockerAddress` types rather than `ComponentAddress`. Much better!

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/account_locker_claimable_resource.rs
sergiupuhalschi-rdx: I was wondering whether I should write doc or not because it looks like all the types in the `wrapped-radix-engine_toolkit` are missing doc. I guess the general rule is always to write doc, right?

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/manifest_account_locker.rs
```diff
@@ -0,0 +1,151 @@
...
+        claimable_resources: Vec<AccountLockerClaimableResource>,
+    ) -> Self {
+        let mut builder = ScryptoManifestBuilder::new();
+        let claimant_arg: ScryptoComponentAddress = (*claimant).into();
+        let claimable_resources =
+            Self::build_claimable_batch(claimable_resources, 50);
+
+        for claimable in claimable_resources.clone() {
```
CyonAlexRDX, line 23: have you run `cargo clippy --fix --allow-staged` ? I think we should change `for claimable in claimable_resources.clone() ` to `for claimable in claimable_resources.iter()` no? (which will be references to `AccountLockerClaimableResource` so you might need to change the match a little bit, e.g. doing `.clone()` below).

Or if you are fine with consuming `claimable_resources` you can `into_iter()` and then all good, which requires you to do the `!claimable_resources.is_empty()` check before calling `into_iter()`. 

i.e.:
`
`
```rust
let should_deposit = !claimable_resources.is_empty();
for claimable_resource in claimable_resources.into_iter {
 /// code here should remain the same you have it
}
```

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/manifest_account_locker.rs
```diff
@@ -0,0 +1,151 @@
...
+impl TransactionManifest {
+    pub fn account_locker_claim(
+        locker_address: &LockerAddress,
+        claimant: &AccountAddress,
+        claimable_resources: Vec<AccountLockerClaimableResource>,
+    ) -> Self {
+        let mut builder = ScryptoManifestBuilder::new();
+        let claimant_arg: ScryptoComponentAddress = (*claimant).into();
+        let claimable_resources =
+            Self::build_claimable_batch(claimable_resources, 50);
```
CyonAlexRDX, line 21: what happens if `claimable_resources.len() > 50`? Should this method not be throwing then? Seems we "truncate" the rest of the resources silently?

CyonAlexRDX: or perhaps at least return a `bool` (or rather an enum!) telling if ALL resources were claimed or if we did truncate some?

sergiupuhalschi-rdx: We truncate the rest of the resources silently and consider the operation successful without notifying the users in any way (at least for now). From UX perspective it will look like this: the user sees that there are some resources to claim for a specific account, they perform the claim transaction -> then they can do that again until they no longer see the notification that something is claimable.

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/manifest_account_locker.rs
sergiupuhalschi-rdx: I assume cloning an entire vector is less efficient than cloning each element from the same vector?

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/manifest_account_locker.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    fn build_claimable_batch(
+        claimable_resources: Vec<AccountLockerClaimableResource>,
+        size: usize,
+    ) -> Vec<AccountLockerClaimableResource> {
+        let mut current_batch_size = 0;
+        claimable_resources
+            .into_iter()
+            .take_while(|claimable| {
+                current_batch_size += match claimable {
```
CyonAlexRDX, line 71: this might in fact benefit to be a method on `AccountLockerClaimableResource`?:

```rust
impl AccountLockerClaimableResource {
    pub fn resource_count(&self) -> usize { match self { ...} }
}
```

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/ret_api.rs
```diff
@@ -168,6 +168,19 @@ pub fn manifest_per_asset_transfers(
     TransactionManifest::per_asset_transfers(transfers)
 }
 
+#[uniffi::export]
+pub fn manifest_account_locker_claim(
+    locker_address: &LockerAddress,
+    claimant: &AccountAddress,
+    claimable_resources: Vec<AccountLockerClaimableResource>,
+) -> TransactionManifest {
+    TransactionManifest::account_locker_claim(
```
CyonAlexRDX, line 177: 🎉

CyonAlexRDX (approved):
Great job!

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/manifest_account_locker.rs
```diff
@@ -0,0 +1,151 @@
...
+    fn from(value: AccountAddress) -> ScryptoComponentAddress {
+        ScryptoComponentAddress::new_or_panic(value.node_id().0)
+    }
+}
+
+impl TransactionManifest {
+    pub fn account_locker_claim(
+        locker_address: &LockerAddress,
+        claimant: &AccountAddress,
+        claimable_resources: Vec<AccountLockerClaimableResource>,
```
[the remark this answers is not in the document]
sergiupuhalschi-rdx: Good point! But that means exposing the `claimable_resources` param through `uniffi` with `IndexSet` type as well. Is it possible? I get the following error and also I couldn't find other usages of this kind with `uniffi`:
`the trait uniffi::TypeId<UniFfiTag> is not implemented for IndexSet<AccountLockerClaimableResource>`

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/manifest_account_locker.rs
```diff
@@ -0,0 +1,151 @@
...
+                        locker_address,
+                        ACCOUNT_LOCKER_CLAIM_NON_FUNGIBLES_IDENT,
+                        (claimant_arg, resource_arg, ids_arg),
+                    );
+                }
+            }
+        }
+
+        if !claimable_resources.is_empty() {
+            builder = builder.deposit_batch(claimant_arg);
```
GhenadieVP, line 58: hmm, should we always use `deposit_batch`?
For asset transfers we do have explicit logic in place that determines if `deposit_batch` or `try_deposit_batch_or_abort` should be used depending on the account deposit config. I think we should do the same here.

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/manifest_account_locker.rs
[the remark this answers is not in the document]
sergiupuhalschi-rdx: Does it make sense to keep `Vec` as the param for this function but then internally converting it to `IndexSet` to remove possible duplicates?

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/manifest_account_locker.rs
sergiupuhalschi-rdx: Well spotted!

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/account_locker_claimable_resource.rs
CyonAlexRDX: yes, all new types and public methods/funcs should be documented, and hopefully we can even add doc to those symbols already missing.

for SargonFOO which is our version of ScryptoFOO or RetFOO you can copy paste the doc of Scrypto | Ret over...

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/manifest_account_locker.rs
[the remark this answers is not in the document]
CyonAlexRDX: sorry for not getting back earlier... was busy with Sargon-MFA. Leave Vec for now.

The pattern I'm using for sets is to let the type be `IdentifiableVec` and let the element be `impl Identifiable for Element` and use the `dec_id_vec!` macro - or whats its name....

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/account_locker/manifest_account_locker.rs
CyonAlexRDX: right but if we return a `bool` indicating that we KNOW that more resources can be claimed, then we can do one less network call from hosts right? :) 

i.e. we can continue to show - or show again - the prompt to claim, based on that bool :)

sergiupuhalschi-rdx: Good point

sergiupuhalschi-rdx: Although I agree it makes sense to tell the hosts whether there are more resources to claim or not, I doubt we need it, so most likely we won't use it in the hosts (at least for now - as per current requirements). We'll still have to perform the network calls as there might be other claimable resources besides the ones already known, we can't rely solely on the `bool` from the previous claiming. The question is if it's worth adding a new uniffi exported model for the sake of correctness and for the eventuality in which it will be needed? Or maybe it's better we add it when we know it's needed for sure?

GhenadieVP (approved):

matiasbzurovski (approved):
I think you need to bump version again, but once that's done it should be good to go

sergiupuhalschi-rdx:
> I think you need to bump version again, but once that's done it should be good to go

I think it's fine, on main it's 1.1.8 and here it's 1.1.9

PR #208, opened 2024-09-04
Accout locker claim non fungibles by amount

sergiupuhalschi-rdx:
Updated non-fungibles claiming from account lockers to take into account the `count` instead of `NonFungibleLocalId`s. Each non-fungible item is still considered an individual resource and we need to adjust the count to not exceed the specified limit.


crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/manifest_account_locker.rs
```diff
@@ -23,73 +24,76 @@ impl TransactionManifest {
...
     ) -> IndexSet<AccountLockerClaimableResource> {
         let mut current_batch_size = 0;
-        claimable_resources
-            .into_iter()
-            .take_while(|claimable| {
-                current_batch_size += claimable.resource_count();
-                current_batch_size <= size
-            })
-            .collect()
+        let mut result = IndexSet::new();
```
CyonAlexRDX, line 69: prefer explicit type annotation here: `let mut result = IndexSet::<AccountLockerClaimableResource>::new();` 

I think it makes code a bit more easy to follow in Github.

CyonAlexRDX (changes requested):

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/manifest_account_locker.rs
```diff
@@ -23,73 +24,58 @@ impl TransactionManifest {
...
-            .into_iter()
-            .take_while(|claimable| {
-                current_batch_size += claimable.resource_count();
-                current_batch_size <= size
-            })
-            .collect()
+        let mut resource_count_left_to_add = max_size;
+        let mut result = IndexSet::<AccountLockerClaimableResource>::new();
+
+        for claimable_resource in claimable_resources {
```
GhenadieVP, line 71: Not sure if I am reading this right - in the case of `resource_count_left_to_add` reaching to be zero, this loop will stop with the assertion in `coerce_resource_count_at_most`, no?

CyonAlexRDX: (I prefer `number_of_foo` over `foo_count`. Also "count left" sounds a bit wrong, I think "left count" sounds better. But `number_of_resources_to_add` (omitting "left") is probably best. "left" signals future action, but so does "to add" as well)

Maybe we want a `if number_of_resources_to_add  <= 0 {  break }` at the end of the loop (or beginning of loop)?

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/sargon_specific_types/account_locker/account_locker_claimable_resource.rs
```diff
@@ -8,18 +8,18 @@ pub enum AccountLockerClaimableResource {
         resource_address: ResourceAddress,
         amount: Decimal192,
     },
-    /// A non-fungible resource with specific claimable IDs
+    /// A non-fungible resource with a specific claimable item count
```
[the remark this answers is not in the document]
sergiupuhalschi-rdx: Aren't the total number of ids in the vault all claimable? Maybe I'm missing something. But I should update the doc to be more precise anyway.

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/manifest_account_locker.rs
sergiupuhalschi-rdx: I shouldn't have missed that. Thank you guys!

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/sargon_specific_types/account_locker/account_locker_claimable_resource.rs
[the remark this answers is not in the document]
GhenadieVP: they are all claimable, but this comment sort of implies that the specified item count will be actually claimed, but it is not always true due the trimming we do.

GhenadieVP (approved):

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/manifest_account_locker.rs
```diff
@@ -23,73 +24,62 @@ impl TransactionManifest {
...
-            })
-            .collect()
+        let mut number_of_items_to_add = max_size;
+        let mut result = IndexSet::<AccountLockerClaimableResource>::new();
+
+        for claimable_resource in claimable_resources {
+            let updated_resource = claimable_resource
+                .coerce_number_of_items_at_most(number_of_items_to_add);
+            result.insert(updated_resource.clone());
+            number_of_items_to_add -= updated_resource.number_of_items();
```
CyonAlexRDX, line 75: maybe a short comment `// can never be negative thanks to clamping performed in `coerce_number_of_items_at_most` ?

CyonAlexRDX (approved):
LGTM, good job!

PR #212, opened 2024-09-10
Sargon OS Boot fixes

micbakos-rdx:
### Sargon Rust changes
* A new function called `derive_wallet` in sargon os. This method can be used when restoring a profile from only a seed phrase and some accounts retrieved from GW.
* A new error is introduced `SecureStorageAccessError`. This is used specifically in Android were the secure storage driver may fail to read from or write to secure storage key which is secured under biometrics. Currently mnemonics fall under this category. This error can be caused by several states of the system. Since I didn't want to make this Android specific, I introduced `SecureStorageAccessErrorKind` which will mostly be used for Android. Most of these kinds are system related and two are related to the user cancelling the authorisation. Although, a suggestion, would be for iOS to return `SecureStorageAccessError(errorKind: .userCancelled)`, maybe when authorisation is cancelled since this seems to be the only possible error on iOS. Whatever the decision is, it will not make any difference. 

### Sargon Android changes
* `SargonOsManager`: Holds a reference to sargon os. Boots sargon when initialised. Implemented as a singleton.
* `AndroidProfileStateChangeDriver`: Holds the state of the current `ProfileState` no need for the host to do that.
* Fixed `encrypt` method to first encrypt and then encode to base64 and not the other way around. This was a bug.
* Fixed biometric requests when the system opens the pattern or pin screens and not the biometrics screen. These screens are usually full screen, causing the app to pause and stop. Before this was buggy because it did not account for that case.
* Inreoduced `KeySpec. reset`. Please take a look at its doc to understand what it does and what problems it fixes especially on `Android API <= 30`


crates/sargon/src/system/sargon_os/sargon_os.rs
```diff
@@ -151,6 +151,38 @@ impl SargonOS {
         Ok(())
     }
 
+    pub async fn derive_wallet(
```
CyonAlexRDX, line 154: Why "derive"? This is not deriving anything..?

Why "wallet"? 

Can we name it something "profile with accounts" ?

micbakos-rdx: Wanted to keep the wallet name consistent with the other "entry" methods like

* `new_wallet`
* `import_wallet`
* `delete_wallet`

crates/sargon/src/profile/v100/profile.rs
```diff
@@ -99,20 +99,28 @@ impl Profile {
...
         if !device_factor_source.is_main_bdfs() {
             panic!("DeviceFactorSource is not main BDFS");
         }
         let bdfs = device_factor_source;
         let header =
             Header::new(DeviceInfo::new_from_info(&host_id, &host_info));
+
+        let mainnet_network = match maybe_accounts {
+            None => ProfileNetwork::new_empty_on(NetworkID::Mainnet),
+            Some(accounts) => ProfileNetwork::new_with_accounts(
```
CyonAlexRDX, line 113: Should we not assert that all accounts are controlled by the BDFS? Maybe not, if accounts passed is a mix of Ledger accounts has software.

Hmmm

crates/sargon/src/system/sargon_os/sargon_os.rs
micbakos-rdx: Unfortunately that is impossible. I wanted to do the same, but in order to do that I need another biometric prompt. The same should apply to import wallet. 

Why is that?
* If we need to ask for biometrics only once, we need to pass the mnemonic to ffi. I am not sure if we want to do that. If we ever did that, then sargon os could check the validity and store it. Thus having total control over it.
* Otherwise we let the host store the mnemonic, as it used to, but now this method will need to read the mnemonic again, asking for biometrics for a second time.

CyonAlexRDX: Right, I think we should do any of those too.

We are already passing the mnemonic across FFI in the form of entropy. So I think best pass a PrivateHierarchicalDeterministicFactorSource from Host, that way we **know** user has intact "wallet"

CyonAlexRDX: Or well, we know if we also check that the accounts are controlled by **that** DeviceFactorSource

crates/sargon/src/profile/v100/profile.rs
GhenadieVP: Issue is that this probably seems like a very generic purpose API, while it should be used only when recovering with a Seed Phrase which by definition can be only BDFS currently. It is meant to replace this Host [logic](https://github.com/radixdlt/babylon-wallet-ios/blob/d50309dd03ffcc306c4849d18f28be5b9062403d/RadixWallet/Clients/ProfileStore/ProfileStore.swift#L161).

crates/sargon/src/system/sargon_os/sargon_os.rs
GhenadieVP: I am in favour of passing the mnemonic over FFI, unless there are some specific concerns. Can't see any in particular.

crates/sargon/src/system/sargon_os/sargon_os.rs
```diff
@@ -151,6 +151,38 @@ impl SargonOS {
...
+            host_info,
+            Some(accounts),
+        );
+
+        self.secure_storage.save_profile(&profile).await?;
+        let profile_state = ProfileState::Loaded(profile);
+        self.profile_state_holder
+            .replace_profile_state_with(profile_state.clone())?;
+
+        self.clients.profile_state_change.emit(profile_state).await;
```
GhenadieVP, line 176: `set_profile` should probably be updated to allow changing to Loaded state with a profile, and then it could be used directly.

jvm/sargon-android/src/main/java/com/radixdlt/sargon/os/driver/AndroidStorageDriver.kt
```diff
@@ -105,12 +88,58 @@ internal class AndroidStorageDriver(
...
+            BiometricPrompt.ERROR_LOCKOUT -> SecureStorageAccessErrorKind.LOCKOUT
+            BiometricPrompt.ERROR_LOCKOUT_PERMANENT -> SecureStorageAccessErrorKind.LOCKOUT_PERMANENT
+            BiometricPrompt.ERROR_NEGATIVE_BUTTON -> SecureStorageAccessErrorKind.NEGATIVE_BUTTON
+            BiometricPrompt.ERROR_NO_BIOMETRICS -> SecureStorageAccessErrorKind.NO_BIOMETRICS
+            BiometricPrompt.ERROR_NO_DEVICE_CREDENTIAL -> SecureStorageAccessErrorKind.NO_DEVICE_CREDENTIAL
+            BiometricPrompt.ERROR_NO_SPACE -> SecureStorageAccessErrorKind.NO_SPACE
+            BiometricPrompt.ERROR_TIMEOUT -> SecureStorageAccessErrorKind.TIMEOUT
+            BiometricPrompt.ERROR_UNABLE_TO_PROCESS -> SecureStorageAccessErrorKind.UNABLE_TO_PROCESS
+            BiometricPrompt.ERROR_USER_CANCELED -> SecureStorageAccessErrorKind.USER_CANCELLED
+            BiometricPrompt.ERROR_VENDOR -> SecureStorageAccessErrorKind.VENDOR
```
GhenadieVP, line 140: Do we actually care to be so granular? or some grouping can be made?

micbakos-rdx: I would like to keep those in order to be able to debug all problems related to biometrics and keystore. These values pass from android driver through ffi to sargon and back to host.

crates/sargon/src/system/sargon_os/sargon_os.rs
micbakos-rdx: > We are already passing the mnemonic across FFI in the form of entropy.

You are right @CyonAlexRDX I forgot about that. Will make the changes.

jvm/sargon-android/src/main/java/com/radixdlt/sargon/os/driver/AndroidBiometricAuthorizationDriver.kt
```diff
@@ -87,24 +48,24 @@ class BiometricsHandler(
...
+            .receiveAsFlow()
+            .flowWithLifecycle(
+                lifecycle = activity.lifecycle,
+                minActiveState = Lifecycle.State.STARTED
+            )
+            .onEach {
+                val result = requestBiometricsAuthorization(activity)
+
+                // Send back the result to sargon os
+                biometricsResultsChannel.send(result)
```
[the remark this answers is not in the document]
GhenadieVP: sure, but the comment then is wrong? this result does not reach the Rust SargonOS, no?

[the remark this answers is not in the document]
micbakos-rdx: It does, in order for the exception to reach the host, it needs to pass through sargon os. Remember this code is executed when sargon requests to read/write something through its secure driver. In order for the host to learn about the error it needs to pass through sargon rust.

jvm/sargon-android/src/main/java/com/radixdlt/sargon/os/driver/AndroidBiometricAuthorizationDriver.kt
```diff
@@ -87,24 +48,24 @@ class BiometricsHandler(
...
-            // Listen to biometric prompt requests while the activity is at least started.
-            activity.lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
-                biometricRequestsChannel.receiveAsFlow().collectLatest {
-                    val result = requestBiometricsAuthorization(activity)
-
-                    // Send back the result to sargon os
-                    biometricsResultsChannel.send(result)
-                }
+        biometricRequestsChannel
+            .receiveAsFlow()
```
giannis-rdx, line 52: why you define the `biometricRequestsChannel` and `biometricsResultsChannel` as `Channel` and then receive the as `Flow`? 
Why not to define them as Flows?

crates/sargon/src/system/sargon_os/sargon_os.rs
micbakos-rdx: @CyonAlexRDX I have implemented the logic to 
1. pass the`PrivateHierarchicalDeterministicFactorSource` into the function 
2. Validate the accounts public keys
3. Ensure that secure storage succeeds saving the mnemonic
 https://github.com/radixdlt/sargon/pull/212/commits/84c7af6878f4e81aeb55eda43049b6df5bf8f6e3
 
 Please let me know if the execution is valid

jvm/sargon-android/src/main/java/com/radixdlt/sargon/os/driver/AndroidBiometricAuthorizationDriver.kt
micbakos-rdx: I wanted to implement a logic of two cold flows. Sargon os can ask for biometrics at any time. So I wanted to solve 2 problems

1. No activity is yet `STARTED` in order to handle biometrics. Remember in order to ask for biometrics, we need a reference to a `FragmentActivity`. In order for the biometrics dialog to show up, we need the activity to be at least started. So I wanted to have a suspending function waiting for an activity to either register and start or return from background. This is why `biometricRequestsChannel` exists. It waits for the request to be handled by one activity.
2. The `biometricsResultsChannel` is waiting for the **USER** to respond. The values of the channel will be emitted by another function.

jvm/sargon-android/src/main/java/com/radixdlt/sargon/os/driver/AndroidBiometricAuthorizationDriver.kt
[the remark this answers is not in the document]
GhenadieVP: I don't mean to be pedantic about this, but there is no such driver as BiometricAuth as far SargonOS is concerned. This result is sent to SecureStorage driver, after that it only is passed to SargonOS, no?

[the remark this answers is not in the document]
micbakos-rdx: It is yes, passed to sargon os, but then a function which uses the driver passes the result back to the host.

For example the function

`sargonOs.newWallet()` can throw such exception which will have travelled all the way from the biometrics handler to the driver to ffi, to sargon rust and back to host's ui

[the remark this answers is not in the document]
GhenadieVP: I understand this, anyway, I am not stubborn about this, just that when reading it for the first time I thought that the result is sent to SargonOS directly and I was confused since there is no such driver in Rust OS.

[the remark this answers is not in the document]
micbakos-rdx: I know the API is difficult to read, since what we are trying to do here is to ask for reading/writing something from sargon rust to the host, without asking for biometrics prior to it. So sargon os is oblivious to asking for biometrics, and does not count for android that does not automatically request for biometrics.

In order to ask for biometrics, it requires communication with the UI layer, and in order to do that with a suspend manner, we need to involve 2 different channels. This result channel sends the result of the biometrics back to the channel which, the secure driver is "subscribed" to and waiting for the response in order to either encrypt/decrypt or fail and communicate to the os why it failed. A simple `FailedToWriteMnemonic` error does not suffice since the host needs to display why it got the error to the user. 

In the case were the error is that the driver failed to encrypt, because the user just cancelled the biometrics, does not need to display that error to the user. But if the error is due to the fact that the user has not enrolled to biometrics, we need to display that to the user, instead of a generic error "Could not create a new wallet" or something similar.

The reason it needs to communicate this failure to the os, is because the host also needs to learn about it.

crates/sargon/src/system/sargon_os/sargon_os.rs
micbakos-rdx: I've made the change here https://github.com/radixdlt/sargon/pull/212/commits/3f32ddb0f28e740af2f04865576f97a199e85a3c. Decided to also use `set_profile` in `new_wallet` too. Let me know your thoughts.

crates/sargon/src/system/sargon_os/sargon_os.rs
```diff
@@ -151,6 +144,51 @@ impl SargonOS {
...
+        let hd_keys: Vec<HierarchicalDeterministicPublicKey> = accounts
+            .iter()
+            .map(|account| {
+                account
+                    .security_state
+                    .into_unsecured()
+                    .map(|c| c.transaction_signing.public_key)
+                    .map_err(|_| CommonError::EntitiesNotDerivedByFactorSource)
+            })
+            .try_collect()?;
```
Sajjon, line 163: cool, I did not know about `try_collect`, we should embrace it!

I think I'd prefer `let ..:  Vec<T> = ....try_collect()?` be rewritten as `let .. = ....try_collect::<Vec<_>>()?`

where `try_collect::<Vec<_>>()` is verbatim, the `_` usage tells rust to fill it in (which it can do based on `.map(|c| c.transaction_signing.public_key)`.

That is the code style I've mostly used elsewhere in Sargon - the advantage is that for long chains of functional calls one can control the return type at the end of the chain, which one typically does by adding yet another map that makes it a bit easier to maintain IMO.

crates/sargon/src/system/sargon_os/sargon_os.rs
```diff
@@ -151,6 +144,51 @@ impl SargonOS {
...
+
+        if !hd_factor_source
+            .mnemonic_with_passphrase
+            .validate_public_keys(hd_keys)
```
Sajjon, line 167: nice! so now we assert that all factor instances are indeed controlled by the specified mnemonic

CyonAlexRDX (approved):
LGTM,I like the change to `PrivateHDFS`.

Left a minor remark about code style

jvm/sargon-android/src/test/java/com/radixdlt/sargon/TransactionManifestTest.kt
```diff
@@ -615,15 +615,15 @@ class TransactionManifestTest : SampleTestable<TransactionManifest> {
...
+                    numberOfItems = 10.toULong()
                 ),
                 AccountLockerClaimableResource.Fungible(
                     resourceAddress = ResourceAddress.init("resource_rdx1tknxxxxxxxxxradxrdxxxxxxxxx009923554798xxxxxxxxxradxrd"),
                     amount = 1500.toDecimal192()
                 ),
                 AccountLockerClaimableResource.NonFungible(
                     resourceAddress = ResourceAddress.init("resource_rdx1n2ekdd2m0jsxjt9wasmu3p49twy2yfalpaa6wf08md46sk8dfmldnd"),
-                    ids = listOf(NonFungibleLocalId.stringId("foobar"))
+                    numberOfItems = 1.toULong()
```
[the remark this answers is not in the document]
micbakos-rdx: Thanks @sergiupuhalschi-rdx wanted to ask you about this

sergiupuhalschi-rdx (approved):
Great job! It's impressive how you found what was the issue with the Keystore secret key generation and fixed it 🏅

crates/sargon/src/system/sargon_os/sargon_os.rs
micbakos-rdx: I like doing this too, but this does not compile. I cannot find a sane way to make this compile.

micbakos-rdx: An example of this working is this:
```
.try_collect::<_, Vec<_>, _>()?;
```
Does this look more readable?

GhenadieVP (approved):
LGTM! Amazing work 🔥!

giannis-rdx (approved):
🦅

micbakos-rdx:
@GhenadieVP Latest commit https://github.com/radixdlt/sargon/pull/212/commits/8505a883c66fbecd346f0131a0c216325c464f25 addresses deleting old ephemeral profile and associated mnemonic when booting again.

PR #213, opened 2024-09-11
Account locker in manifest summary encountered entities

sergiupuhalschi-rdx:


crates/sargon/src/wrapped_radix_engine_toolkit/high_level/sargon_specific_types/manifest_encountered_address.rs
```diff
@@ -0,0 +1,95 @@
+use crate::prelude::*;
+
+address_union!(
+    /// A tagged union of all the encountered addresses in the manifest.
+    /// This is to be primarily used for the "using dApps" section of the wallet's tx review screen.
+    enum ManifestEncounteredAddress: component, locker
```
sergiupuhalschi-rdx, line 6: I'm not sure about the enum name

CyonAlexRDX (changes requested):
Good job!

Missing sample values. Can you also Swift and Kotlinify this.

In Swift conform to Address protocol, see https://github.com/radixdlt/sargon/blob/main/apple/Sources/Sargon/Extensions/Swiftified/Address/AddressOfAccountOrPersona%2BSwiftified.swift

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/sargon_specific_types/manifest_encountered_address.rs
GhenadieVP: Let's not add other types for now, the logic in the Wallet to display the "used dApps" is explicitly for globalGenericComponents + account locker.

At some point, RET had a field called `userApplications` in summary, meant to point that these potentially have associated dApps. Maybe we can have the field in the execution summary more specific to potential dApps used, rather than generic `ManifestEncounteredAddress`, which presumably, by the name, should include all addresses

sergiupuhalschi-rdx: As discussed with @GhenadieVP it seems safer to add only component and locker for now, to prevent showing dApp section in tx review when it's not required.

sergiupuhalschi-rdx: @GhenadieVP do you have an example? I agree, but I still can't think of a better name to reflect what you described, `UserApplicationAddress` doesn't sound better and it lacks context (as the keyword `manifest` offers).

sergiupuhalschi-rdx: `ManifestUserApplicationAddress` / `ManifestDAppAddress` / `ManifestDAppEntityAddress` are any of these better?

CyonAlexRDX (approved):
LGTM ! Good job

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/sargon_specific_types/manifest_encountered_address.rs
Sajjon: `AddressReferencedInTX` ?

.github/workflows/test.yml
```diff
@@ -139,6 +139,13 @@ jobs:
     steps:
       - uses: RDXWorks-actions/checkout@main
 
+      - name: Remove old xcode versions
```
CyonAlexRDX, line 142: We can try this.

But it is possible this breaks test Swift - and possible it works for most runs but fails for some due to non determinism.

If we have troubles we can change so that Kotlin tests are run AFTER Swift tests:

Which can be done by using `needs: test-Swift`

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/sargon_specific_types/manifest_encountered_address.rs
GhenadieVP: For now I guess `ManifestEncounteredComponentAddress`, can't think of a very good name either :).

sergiupuhalschi-rdx: @GhenadieVP `AddressReferencedInTX ` this looks like the most accurate option, is it fine by you?

GhenadieVP: Isn't it too generic? I mean that we are filtering the original addresses after all

sergiupuhalschi-rdx: At first, I thought that `AddressReferencedInTX` is more specific, but now that I think of it all of the addresses referenced in a manifest are referenced in the transaction so I guess you're right and it's better to put emphasis on the fact that the addresses are of `components`. I'll rename it to `ManifestEncounteredComponentAddress`.

sergiupuhalschi-rdx: Yes, it's not a component address, but locker is a component, right? But I see what you're saying and that's why I didn't use `component` when naming it initially.

sergiupuhalschi-rdx: Now the `component` keyword expresses that it's not just encountered addresses, but addresses of components (not resources or other types of addresses there are). I'd leave it as is unless someone has other suggestions that reflect the fact that the addresses were filtered to match some criteria.

PR #217, opened 2024-09-20
[ABW-3809] Update to new receipt

matiasbzurovski:
## Changelog
- Update engine toolkit to `12a9ce36cc8c36c728f095b450caeb45195eb431`
- The execution summary for the `TransactionManifest` now uses `engine_toolkit_receipt` as param (instead of the `encoded_receipt`)


Package.swift
```diff
@@ -55,7 +55,7 @@ let package = Package(
 		.package(url: "https://github.com/SwiftyJSON/SwiftyJSON", from: "5.0.2"),
 		
 		// Multicast / Share of notifications in EventBus
-		.package(url: "https://github.com/sideeffect-io/AsyncExtensions", exact: "0.5.3"),
+		.package(url: "https://github.com/sideeffect-io/AsyncExtensions", exact: "0.5.2"),
```
CyonAlexRDX, line 58: was this intended? I think we want `0.5.3` for Xcode 16 support, but is it not Xcode 15 backwards compat? I think it is?

crates/sargon/src/lib.rs
```diff
@@ -59,6 +59,8 @@ pub mod prelude {
     pub(crate) use uuid::Uuid;
 
     pub(crate) use enum_as_inner::EnumAsInner;
+    pub(crate) use native_radix_engine_toolkit::receipt::RuntimeToolkitTransactionReceipt as RetRuntimeToolkitTransactionReceipt;
```
CyonAlexRDX, line 62: Ah this should be called `ScryptoRuntimeToolkitTransactionReceipt` I think... So I've been using the `RET` or `Ret` prefix for types which comes from a crate in the Radix-Engine-Toolkit Git repository. And I've been using `Scrypto` prefix for types which comes from a crate in the Scrypto Git repo, and `native_radix_engine_toolkit` crate comes from Scrypto repo...

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/transaction_manifest/execution_summary/transaction_manifest_execution_summary.rs
```diff
@@ -5,27 +5,41 @@ use radix_engine_toolkit::functions::manifest::execution_summary as RET_executio
 impl TransactionManifest {
     pub fn execution_summary(
         &self,
-        encoded_receipt: BagOfBytes,
+        engine_toolkit_receipt: BagOfBytes,
     ) -> Result<ExecutionSummary> {
-        let receipt: TransactionReceipt = encoded_receipt.try_into()?;
+        let network_definition = self.network_id().network_definition();
```
CyonAlexRDX, line 10: oh wow, was it really this complicated? Fells like we really should be able to simplify this... is this what results in the least amount of code from Host? We should always optimize for host ofc!

apple/Sources/Sargon/Extensions/Methods/RET/TransactionManifest+Wrap+Functions.swift
```diff
@@ -39,10 +39,10 @@ extension TransactionManifest {
         transactionManifestSummary(manifest: self)
     }
     
-    public func executionSummary(encodedReceipt: Data) throws -> ExecutionSummary {
+    public func executionSummary(engineToolkitReceipt: Data) throws -> ExecutionSummary {
         try transactionManifestExecutionSummary(
```
CyonAlexRDX, line 43: we should document how Hosts retrive and provide this data, here and in Rust fn.

Package.swift
matiasbzurovski: this will be reverted once the draft is ready for review. Needed to drop down to `0.5.2` since Wallet is not compiling with `0.5.3`.

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/transaction_manifest/execution_summary/transaction_manifest_execution_summary.rs
matiasbzurovski: I was able to copy-pasta this from RET [repo](https://github.com/radixdlt/radix-engine-toolkit/commit/12a9ce36cc8c36c728f095b450caeb45195eb431#diff-d09f986affca4b35d7053153c834828e7d0565db366475aea8fcbf0814a60035R107-R121) 😅

apple/Sources/Sargon/Extensions/Methods/RET/TransactionManifest+Wrap+Functions.swift
matiasbzurovski: will add!

crates/sargon/fixtures/transaction/create_single_fungible.rtm
Sajjon, line 1: a bit sad to lose the descriptive comments, but maybe not worth mapping them over

GhenadieVP (approved):

CyonAlexRDX (approved):
LGTM

giannis-rdx (approved):

PR #220, opened 2024-09-25
Check version bump when opening PRs

matiasbzurovski:
Every time we forget to bump version on `Cargo.toml`, we get out of sync between the released version tag and the one indicated in such file.

This PR adds a check to verify every time a PR is opened against `main`, we have changed the `version` inside `Cargo.toml`.

You can see on this [workflow run](https://github.com/radixdlt/sargon/actions/runs/11032774718/job/30642437815) how CI failed when I had opened the PR against `main` (instead of `ci/ubuntu-runners`).

Screenshot 2024-09-25 at 14 17 51


micbakos-rdx:
@matiasbzurovski Should we maybe check to bump the version only if there are changes to rust, kotlin or swift src code? For example in the previous PR I changed the CI and the README, but didn't push any changes to src. What do you think?

matiasbzurovski:
@micbakos-rdx the problem is that every time we merge into `main`, a new release is generated. For example, with your CI/CD change, a new [release](https://github.com/radixdlt/sargon/releases/tag/1.1.18) was generated and since the `Cargo.toml` wasn't changed, it got out of sync.

micbakos-rdx:
@matiasbzurovski At least for sargon-android a new library build is produced and published. This library gets the version from cargo toml directly and appends the commit hash as suffix. So if there is no change to the version the hash will be different.

matiasbzurovski:
on iOS we set the dependency based on the release version. So for example, if we set it to [1.1.18](https://github.com/radixdlt/sargon/releases/tag/1.1.18), the app will use the library from such tag. However, such library has Cargo version set to `1.1.16`, which leads to confusion IMO. 

For example, here we use dependency `1.1.18` but app shows `Sargon: 1.1.16` (given `1.1.18` tag has set `version: 1.1.16` in `Cargo.toml`).

CyonAlexRDX (approved):
cool, lets give it a spin

PR #223, opened 2024-09-26
Add JNA dependency to sargon-desktop-bins

micbakos-rdx:
Since there is no `sargon-desktop` module, there is no information for its dependencies. A dependency that needs to ship with sargon is the jna runtime. 

This can only be mitigated with adding a custom pom dependency for runtime jna.

### DIFF
--- a/jvm/sargon-android/build.gradle.kts
+++ b/jvm/sargon-android/build.gradle.kts
@@ -124,9 +124,7 @@ koverReport {
 }
 
 dependencies {
-    // Cannot use version catalogues for aar. For some reason when published to Maven,
-    // the jna dependency cannot be resolved
-    implementation("net.java.dev.jna:jna:5.13.0@aar")
+    implementation("${libs.jna.get()}@aar")
 
     // For lifecycle callbacks
     implementation(libs.androidx.appcompat)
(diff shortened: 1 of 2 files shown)


jvm/sargon-android/build.gradle.kts
```diff
@@ -192,6 +190,18 @@ publishing {
...
+                withXml {
+                    val dependencies = asNode().appendNode("dependencies")
+
+                    val jni = dependencies.appendNode("dependency")
+                    jni.appendNode("groupId", "net.java.dev.jna")
+                    jni.appendNode("artifactId", "jna")
+                    jni.appendNode("version", libs.versions.jna.get())
+                    jni.appendNode("scope", "runtime")
+                }
+            }
```
micbakos-rdx, line 204: We need to provide a "hardcoded" runtime dependency to jna

giannis-rdx (approved):

jvm/sargon-android/build.gradle.kts
micbakos-rdx: So pom file should look like this:
```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.radixdlt.sargon</groupId>
  <artifactId>sargon-desktop-bins</artifactId>
  <version>1.1.19-6dc74bec</version>
  <dependencies>
    <dependency>
      <groupId>net.java.dev.jna</groupId>
      <artifactId>jna</artifactId>
      <version>5.13.0</version>
      <scope>runtime</scope>
    </dependency>
  </dependencies>
</project>
```

PR #224, opened 2024-09-26
Fixed Sargon iOS example app. (Xcode 16)

CyonAlexRDX:
Fixed Sargon iOS example app which got broken by recent changes to SargonOS.

This can be used to understand SecurityShields a bit. Since it has a working POC of security shields


matiasbzurovski (approved):
minor: seems we are presenting an empty view after adding an account

https://github.com/user-attachments/assets/5fb2ba1b-ddcd-401b-a50c-222ca2a9d9cf

CyonAlexRDX:
mm I saw that too, weird, someone can make take a look and find out why :D I leave it as a exercise :D

PR #225, opened 2024-09-27
Profile safe concurrent access

GhenadieVP:
Make use of [read](https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html#method.read) and [write](https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html#method.write) on the RWLock instead of [try_read](https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html#method.try_read) and [try_write](https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html#method.try_write), which assure proper concurrent read/write access to the profile_state.

### DIFF
--- a/crates/sargon/src/system/sargon_os/profile_state_holder.rs
+++ b/crates/sargon/src/system/sargon_os/profile_state_holder.rs
@@ -94,10 +94,9 @@ impl ProfileStateHolder {
     where
         F: Fn(&Profile) -> T,
     {
-        let guard = self
-            .profile_state
-            .read()
-            .unwrap_or_else(|poison| poison.into_inner());
+        let guard = self.profile_state.read().expect(
+            "Stop execution due to the profile state lock being poisoned",
+        );
 
         let state = &*guard;
         match state {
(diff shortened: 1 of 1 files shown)


crates/sargon/src/system/sargon_os/profile_state_holder.rs
```diff
@@ -94,10 +94,9 @@ impl ProfileStateHolder {
     where
         F: Fn(&Profile) -> T,
     {
-        let guard = self
-            .profile_state
-            .try_read()
-            .expect("Implementing hosts should not read and write Profile from multiple threads.");
+        let guard = self.profile_state.read().expect(
```
GhenadieVP, line 97: We consider any poisoning of the lock as bad state, it is possible to read the poisoned state and possibly use it, but the state can be in an intermediary bad form.  So if any profile mutation panicked on a separate thread, any further operation on the state will be aborted.

micbakos-rdx (approved):

CyonAlexRDX (approved):
lgtm

PR #230, opened 2024-10-08
[ABW-3860] Move submit tx & poll tx status to Sargon

matiasbzurovski:
JIRA ticket: [ABW-3860](https://radixdlt.atlassian.net/browse/ABW-3860)

## Changelog
Adds new function on SargonOS that allow host apps to submit a transaction and poll its status on Sargon.

[ABW-3860]: https://radixdlt.atlassian.net/browse/ABW-3860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/sargon/src/system/sargon_os/sargon_os_transactions.rs
```diff
@@ -0,0 +1,382 @@
...
+// ==================
+// Submit Transaction
+// ==================
+#[uniffi::export]
+impl SargonOS {
+    /// Submits a notarized transaction payload to the network.
+    pub async fn submit_transaction(
+        &self,
+        notarized_transaction: NotarizedTransaction,
+    ) -> Result<IntentHash> {
```
matiasbzurovski, line 13: Might remove this exposed function, and just leave `submit_compiled_transaction` instead

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/ret_api.rs
```diff
@@ -142,14 +142,6 @@ pub fn xrd_address_of_network(network_id: NetworkID) -> ResourceAddress {
     ResourceAddress::xrd_on_network(network_id)
 }
 
-#[uniffi::export]
-pub fn debug_print_compiled_notarized_intent(
-    compiled: CompiledNotarizedIntent,
-) -> String {
-    let notarized = compiled.decompile();
-    format!("{:?}", notarized)
```
matiasbzurovski, line 150: This isn't used on Android, and on iOS we weren't [printing anything](https://github.com/radixdlt/babylon-wallet-ios/blob/main/RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient%2BLive.swift#L58), just declaring a String that Sargon provided (but not logging it), so dediced to remove.

crates/sargon/src/gateway_api/models/logic/response/transaction/status/transaction_status.rs
```diff
@@ -0,0 +1,36 @@
+use crate::prelude::*;
+
+#[cfg(test)]
+mod tests {
```
GhenadieVP, line 4: Why is this test here and nit near the `TransactionStatusResponse` declaration?

crates/sargon/src/system/sargon_os/sargon_os_transactions.rs
GhenadieVP: I would say the other way around, remove the `submit_compiled_transaction` and leave the compiling of the NotarizedTransaction to sargon, which already happens [here](https://github.com/radixdlt/sargon/blob/23d99bd21cffb6170d2caaa1aec5eea994c1e6c4/crates/sargon/src/gateway_api/models/logic/request/transaction/submit/transaction_submit.rs#L5).

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/sargon_specific_types/transaction/transaction_status.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(
+    Clone,
+    Debug,
+    Serialize,
+    Deserialize,
+    PartialEq,
+    Eq,
+    Hash,
+    Ord,
+    PartialOrd,
```
GhenadieVP, line 12: Do we really need the ordering?

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/sargon_specific_types/transaction/transaction_status_reason.rs
```diff
@@ -0,0 +1,79 @@
...
+#[derive(
+    Clone,
+    Debug,
+    Serialize,
+    Deserialize,
+    PartialEq,
+    Eq,
+    Hash,
+    Ord,
+    PartialOrd,
```
GhenadieVP, line 12: Same here

crates/sargon/src/system/sargon_os/sargon_os_transactions.rs
```diff
@@ -0,0 +1,382 @@
...
+        let gateway_client = GatewayClient::new(
+            self.clients.http_client.driver.clone(),
+            network_id,
+        );
+
+        // The delay increment is set to 1 second in production, but 0 seconds in tests.
+        // This will make the tests run with 0s delay, while the production code will have a 2s delay on first call,
+        // a 3s delay on second call, 4s on third and so on.
+        #[cfg(test)]
+        const DELAY_INCREMENT: u64 = 0;
```
Sajjon, line 54: We might wanna unit test  the durations slept (a vec, and the length of which will be the amount we polled (plus or minus 1 :P )), to assert we correctly increase duration between each loop.

Thus you must use a non zero value for `cfg(test)` so I propse we switch to milliseconds instead of seconds. and you can use 1ms for tests...

 `poll_transaction_status` into two methods:
```rust
/// Live impl, but retaining some statistics about how many times we did poll, which
/// are useful for tests
fn poll_transaction_status_with_stats(..) -> Result<(TransasctionStatus, Vec<u64>)> // `Vec<u64>` is duration slept between each poll in milliseconds
```

and  a method `poll_transaction_status` which signature is identical to what you have here, which calls `poll_transaction_status_with_stats` but ignores the stats (return type `Result<TransasctionStatus>`.

crates/sargon/src/system/sargon_os/sargon_os_transactions.rs
```diff
@@ -0,0 +1,382 @@
...
+        #[cfg(not(test))]
+        const DELAY_INCREMENT: u64 = 1;
+
+        let mut delay_duration = DELAY_INCREMENT;
+        loop {
```
Sajjon, line 59: we might wanna include a timeout value to this poll function? where we typically would use `30sec` or so by hosts ?

matiasbzurovski: what would be the benefit of having a timeout? We don't have one on current implementation ([iOS](https://github.com/radixdlt/babylon-wallet-ios/blob/main/RadixWallet/Clients/SubmitTransactionClient/SubmitTransactionClient%2BLive.swift#L23-L51)), and I don't think there is a reason to have one. 
If the network is congested and takes longer than 30s to process, our requests to the gateway would be every time more separated in time each from each other (meaning we won't saturate our GW).

crates/sargon/src/gateway_api/models/logic/response/transaction/status/transaction_status.rs
matiasbzurovski: note that this is located under `logic/response/transaction/status`. If you go to the analogue folder under `types/response/transaction/status` you will also find a `transaction_status.rs` that holds the `TransactionStatusResponse` type.

I assume your confusion comes from the fact that the files are named `transaction_status.rs` and not `transaction_status_response.rs`.  I could rename both to the latter, although we skip the `_response` suffix for other similar cases (such as `TransactionSubmitResponse`), which I think makes sense since the root folder is `response/`. 

Let me know what you guys prefer, and we can follow such standard cc @CyonAlexRDX

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/sargon_specific_types/transaction/transaction_status.rs
matiasbzurovski: definitely not, removing as per you comment on meeting yesterday.

crates/sargon/src/system/sargon_os/sargon_os_transactions.rs
CyonAlexRDX: Yeah maybe not needed. Just feels right :) 

I think gateway team would prefer it too. A max number of polls

crates/sargon/src/gateway_api/models/logic/response/transaction/status/transaction_status.rs
CyonAlexRDX: I think Ghenadie meant that these unit tests are placed in a separate file. And might instead be placed next to the declaration of the type..?

matiasbzurovski: oh! in such case I followed the `type` vs `logic` separation that we already have.

CyonAlexRDX: I think that is fine

CyonAlexRDX (approved):
Good job!

GhenadieVP (approved):

PR #231, opened 2024-10-08
Move transaction preview analysis

sergiupuhalschi-rdx:
[ABW-3864](https://radixdlt.atlassian.net/browse/ABW-3864)

Moved into Sargon the logic of doing the initial transaction analysis for a given raw manifest. At high level it implies the following:

- Creating the TransactionManifest from a raw manifest;
- Extracting the transaction signers for the manifest;
- Executing the transaction preview GW request;
- Running the execution summary with the manifest and receipt.

[ABW-3864]: https://radixdlt.atlassian.net/browse/ABW-3864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/sargon/src/system/sargon_os/sargon_os_transaction.rs
```diff
@@ -0,0 +1,169 @@
...
+            network_id,
+            epoch,
+            Epoch::from(epoch.0 + 10),
+            nonce,
+            notary_public_key,
+            signer_public_keys.is_empty(),
+            0,
+        );
+        let intent = TransactionIntent::new(header, manifest, message)?;
+        let request = TransactionPreviewRequest::new(
```
GhenadieVP, line 135: You need to set the opt in to get the radix_engine_toolkit_receipt in response.

GhenadieVP (commented):
Looks good overall, going in the right direction 👍

crates/sargon/src/system/sargon_os/sargon_os_transaction.rs
sergiupuhalschi-rdx: It's being set by default in the constructor function: `TransactionPreviewRequest::new`. By looking at the wallet code `radix_engine_toolkit_receipt` is always true and never missing, so I guess it's fine to leave it set to true as default.

GhenadieVP: aah right, missed that.

crates/sargon/src/system/sargon_os/sargon_os_transaction.rs
```diff
@@ -0,0 +1,169 @@
...
+        notary_public_key: PublicKey,
+    ) -> Result<TransactionPreviewResponse> {
+        // Get all transaction signers
+        let signer_public_keys =
+            self.extract_transaction_signers(manifest.summary()).await?;
+        let epoch = gateway_client.current_epoch().await?;
+        let header = TransactionHeader::new(
+            network_id,
+            epoch,
+            Epoch::from(epoch.0 + 10),
```
sergiupuhalschi-rdx, line 128: @GhenadieVP Here on Android we add to the current ledger epoch `1` while on iOS `10`. Not sure which one is right.

GhenadieVP: AFAIK, 10 is the right value.

crates/sargon/src/gateway_api/models/types/response/transaction/preview/transaction_status.rs
```diff
@@ -0,0 +1,17 @@
...
+    Deserialize,
+    Serialize,
+    Clone,
+    PartialEq,
+    Eq,
+    Debug,
+    derive_more::Display,
+    uniffi::Enum,
+)]
+pub enum TransactionStatus {
```
matiasbzurovski, line 13: Can this be replaced by the [TransactionStatus](https://github.com/radixdlt/sargon/blob/main/crates/sargon/src/wrapped_radix_engine_toolkit/high_level/sargon_specific_types/transaction/transaction_status.rs#L15) we have from RET?

sergiupuhalschi-rdx: As I can see they are quite different or am I missing something?

matiasbzurovski: I hadn't checked where it was used, that's why I wrongly assumed it was the same. 

It is used under in `TransactionReceipt`, so it isn't actually the `TransactionStatus` but a ` TransactionReceiptStatus`.

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_submit.rs
sergiupuhalschi-rdx, line 1: This has been reviewed in #230. I just moved the files as agreed with Matias.

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_status.rs
sergiupuhalschi-rdx, line 1: This has been reviewed in #230. I just moved the files as agreed with Matias.

crates/sargon/src/gateway_api/models/types/response/transaction/preview/transaction_receipt.rs
```diff
@@ -0,0 +1,17 @@
...
+    Serialize,
+    Clone,
+    PartialEq,
+    Eq,
+    Debug,
+    derive_more::Display,
+    uniffi::Record,
+)]
+#[display("{status}")]
+pub struct TransactionReceipt {
```
CyonAlexRDX, line 14: missing one line doc, would also be great with `HasSampleValues` impl and the standard:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    type SUT = .... type goes here;

    #[test]
    fn equality() {
        assert_eq!(SUT::sample(), SUT::sample());
        assert_eq!(SUT::sample_other(), SUT::sample_other());
    }

    #[test]
    fn inequality() {
    }
}
```

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -0,0 +1,705 @@
+use std::sync::RwLockWriteGuard;
+
+use crate::prelude::*;
+
+#[uniffi::export]
+impl SargonOS {
+    pub async fn analyse_transaction_preview(
```
CyonAlexRDX, line 7: missing docs, maybe at least add one line?

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -0,0 +1,705 @@
...
+            is_wallet_transaction,
+            nonce,
+            notary_public_key,
+        )
+        .await
+    }
+}
+
+impl SargonOS {
+    pub async fn perform_transaction_preview_analysis(
```
CyonAlexRDX, line 29: missing docs, maybe at least add one line?

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -0,0 +1,705 @@
...
+        &self,
+        instructions: String,
+        blobs: Blobs,
+        message: Message,
+        is_wallet_transaction: bool,
```
CyonAlexRDX, line 34: I think we should use the term `host` instead of `wallet`. 

so maybe change `is_wallet_transaction` -> `are_instructions_originating_from_host` or `instructions_originates_from_host` or similar?

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -0,0 +1,705 @@
...
+        nonce: Nonce,
+        notary_public_key: PublicKey,
+    ) -> Result<TransactionPreviewResponse> {
+        let signer_public_keys =
+            self.extract_transaction_signers(manifest.summary()).await?;
+        let epoch = gateway_client.current_epoch().await?;
+        let header = TransactionHeader::new(
+            network_id,
+            epoch,
+            Epoch::from(epoch.0 + 10),
```
CyonAlexRDX, line 103: `10` should probably be a `pub const` on `Epoch` it self:

```rust
impl Epoch {
    /// Cirka 1 hour, since one epoch is cirka 6 minutes.
    pub const DEFAULT_EPOCH_WINDOW_SIZE: u64 = 10;
}
```

And perhaps a pub ctor:

```rust
impl Epoch {
   pub fn window_end_from_start(start: Self) -> Self {
       Self::new(start.0 + Self::DEFAULT_EPOCH_WINDOW_SIZE);
   }
}
```

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -0,0 +1,705 @@
...
+            nonce,
+            notary_public_key,
+            signer_public_keys.is_empty(),
+            0,
+        );
+        let intent = TransactionIntent::new(header, manifest, message)?;
+        let request = TransactionPreviewRequest::new(
+            intent,
+            signer_public_keys,
+            TransactionPreviewRequestFlags::new(true, false, false),
```
CyonAlexRDX, line 113: ah ok, hmm can you please make newtypes for these three flags. That should have been done before IMO. since `true, false, false` is hard to read. Better with `IsFoo(true), IsBar(false), IsBuz(false)`.

There is a macro for this! see:

```rust
decl_bool_type!(IsCloudProfileSyncEnabled, true);
decl_bool_type!(IsDeveloperModeEnabled, false);
decl_bool_type!(IsAdvancedLockEnabled, false);
```
the bool literal is the default value

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -0,0 +1,705 @@
...
+    ) -> CommonError {
+        let message = response
+            .receipt
+            .error_message
+            .unwrap_or_else(|| "Unknown reason".to_string());
+
+        // Quite rudimentary, but it is not worth making something smarter,
+        // as the GW will provide in the future strongly typed errors
+        let is_failure_due_to_deposit_rules = message
+            .contains("AccountError(DepositIsDisallowed")
```
CyonAlexRDX, line 133: right, I would at the very list make these String Literals into const variables. can be global and non public

```
/// PLEASE DOCUMENT ME
const GW_ERR_ACCOUNT_DEPOSIT_DISALLOWED: &str = "AccountError(DepositIsDisallowed";

/// PLEASE DOCUMENT ME
const GW_... = "AccountError(NotAllBucketsCouldBeDeposited";
```

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -0,0 +1,705 @@
...
+        let account_entities: IndexSet<_> = manifest_summary
+            .addresses_of_accounts_requiring_auth
+            .iter()
+            .filter_map(|address| {
+                accounts
+                    .iter()
+                    .find(|account| account.address == *address)
+                    .map(AccountOrPersona::AccountEntity)
+            })
+            .collect();
```
CyonAlexRDX, line 173: put `IndexSet<_>` in the `collect` instead please. so `let variable_name = chain.chain.chain.collect::<TYPE>();`

CyonAlexRDX: I think this **silences** non found entities,. right? we should use a `get_account(..) -> Result<Account>` and use:

```rust
addresses_of_accounts_requiring_auth
.into_iter()
.map(|a| self.profile_state_holder.get_account(a))
.collect::<Result<IndexSet<_>>>()?;
```

which will make this whole method return `Err` if any account was not found.

And analoguosly for Personas.

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -0,0 +1,705 @@
...
+
+    fn extract_transaction_signers_public_keys(
+        &self,
+        signers: IndexSet<AccountOrPersona>,
+    ) -> IndexSet<PublicKey> {
+        signers
+            .iter()
+            .flat_map(|entity| {
+                let public_keys: IndexSet<PublicKey> = entity
+                    .virtual_hierarchical_deterministic_factor_instances()
```
CyonAlexRDX, line 198: ah, you want to use a different function for the signeres, since the current impl of : `virtual_hierarchical_deterministic_factor_instances` includes ROLA keys, with `CAP26KeyKind::AuthenticationSigning` which cannot be used for `TransctionSigning`.

And since this method is called "transaction_signers"` you wanna FILTER to only include the public keys of kind TransactionSigning.

Furthermore this method should probably return `HiearchalDeterministicFactorInstance` instead of `PublicKey`, or rather you should create a `extract_transaction_signers_factor_instances(&self, signers: IndexSet<AccountOrPersona>) -> IndexSet<HiearchalDeterministicFactorInstance>` 

and then the `extract_transaction_signers_public_keys` can simply call `extract_transaction_signers_factor_instances` peforming the `map(|fi| fi.public_key.public_key)`..

crates/sargon/src/profile/v100/entity/has_entity_security_state.rs
```diff
@@ -0,0 +1,24 @@
+use crate::prelude::*;
+
+pub trait HasEntitySecurityState {
+    fn security_state(&self) -> EntitySecurityState;
+
+    fn virtual_hierarchical_deterministic_factor_instances(
+        &self,
```
CyonAlexRDX, line 7: add a `filter_kind: impl Into<Option<CAP26KeyKind>>`  and then instead of returning `factor_instances` you can:

```rust
let Some(key_kind) = filter_key_kind.into() else { return  factor_instances };
factor_instances.into_iter().filter(|fi| f.derivation_path().key_kind == key_kind).collect()
```

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -0,0 +1,705 @@
...
+            result,
+            Err(CommonError::ExecutionSummaryFail {
+                underlying: "InvalidReceipt".to_string()
+            })
+        )
+    }
+
+    #[actix_rt::test]
+    async fn execution_summary_reserved_instructions_error() {
+        let responses = prepare_responses(
```
CyonAlexRDX, line 505: many `ScryptoDecimal192::zero().into(),`, I think it derives `Copy` so you might wanna do a `let zero: DecimalType = ScryptoDecimal192::zero();` in the top here and use `zero` variable. 

You are calling into, which type are we changing into? A `RET` Decimal?

crates/sargon/src/system/drivers/networking_driver/support/test/mock_networking_driver.rs
```diff
@@ -25,10 +25,26 @@ impl MockNetworkingDriver {
...
+        }
+    }
+
     pub fn new(status: u16, body: impl Into<BagOfBytes>) -> Self {
         Self::with_spy(status, body, |_| {})
     }
 
+    pub fn new_with_bodies(status: u16, bodies: Vec<BagOfBytes>) -> Self {
+        Self::with_spy_multiple_bodies(status, bodies, |_| {})
+    }
```
matiasbzurovski, line 46: Seems you could use the `with_responses` I added instead, no?

CyonAlexRDX:
Good job so far! Nothing major, but some low hanging fruit improvements to be made before this pr is merge ready IMO

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -0,0 +1,705 @@
...
+                transaction_manifest.clone(),
+                network_id,
+                message,
+                nonce,
+                notary_public_key,
+            )
+            .await?;
+        let engine_toolkit_receipt = transaction_preview
+            .radix_engine_toolkit_receipt
+            .ok_or(CommonError::FailedToExtractTransactionReceiptBytes)?;
```
matiasbzurovski, line 59: If we are gonna default to `FailedToExtractTransactionReceiptBytes`, then aren't we loosing the error mapped in `map_failed_transaction_preview()`?

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/transaction_manifest/execution_summary/transaction_manifest_execution_summary.rs
```diff
@@ -6,26 +6,34 @@ impl TransactionManifest {
     /// Creates the `ExecutionSummary` based on the `engine_toolkit_receipt` data.
     ///
     /// Such value should be obtained from the Gateway `/transaction/preview` endpoint, under the `radix_engine_toolkit_receipt` field.
-    /// The field will be a JSON that host apps should parse into a .
+    /// The field will be a JSON that host apps should parse into a transaction preview.
```
matiasbzurovski, line 9: Well, that's no longer the case. Now we are going to make the parsing of `radix_engine_toolkit_receipt` in Sargon.

matiasbzurovski: Actually, shouldn't this method be deleted now? We would only build the `ExecutionSummary` through SargonOS, which will internally call `execution_summary_with_engine_toolkit_receipt`

crates/sargon/src/system/drivers/networking_driver/support/test/mock_networking_driver.rs
sergiupuhalschi-rdx: Not really. `with_responses` takes a Vec of a specific type, while I need different types of responses so I pass the serialized objects directly.

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/transaction_manifest/execution_summary/transaction_manifest_execution_summary.rs
sergiupuhalschi-rdx: Right!

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
sergiupuhalschi-rdx: No, because the function will exit before, if `get_transaction_preview` has `Err` result.

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
sergiupuhalschi-rdx: Good point! Yeap, we're changing them into `RET` Decimal.

crates/sargon/src/profile/v100/entity/has_entity_security_state.rs
```diff
@@ -0,0 +1,31 @@
...
+                    factor_instances.insert(authentication_signing);
+                }
+            }
+        }
+        let Some(key_kind) = filter_key_kind.into() else {
+            return factor_instances;
+        };
+        factor_instances
+            .into_iter()
+            .filter(|fi| fi.key_kind() == Some(key_kind))
```
CyonAlexRDX, line 28: oh! riiiiight.... this is a dangerous bug here! 

I forgot `key_kind()` method on `HDFactorInstance` returns `Option`, which it must do because of legacy `DerivatoinPath::BIP44Like`.

But we MUST include BIP44 factors as if they were `CAP26KeyKind::TransactionSigning` if we pass `filter_key_kind: CAP26KeyKind` to this `virtual_hierarchical_deterministic_factor_instances` method.

So we must massage this function / these types a little bit.

Start by adding a unit tests creating `HierarchicalDeterministicFactorInstance` with a `BIP44Like` DerivationPath`. Looks like we do not have a `legacy` account sample value, so start by adding that. And it is so important that we test Legacy accounts in general that I lean towards perhaps even changing our `Account::sample_other` to be a Legacy account - doing that will cascade A LOT. All our Profile sample values will have to be updated. So maybe we should defer it until Ghenadies https://github.com/radixdlt/sargon/pull/237 has been merged.

Anyway, you will need an Account with BIP44 `secp256k1` public key as VECI (Virtual Entity (Account) Creating (Factor) Instance), lets call it `legacy_account` here, and then call `virtual_hierarchical_deterministic_factor_instances` and specify that you care about `TransactionSigning`, the unit test should assert that `legacy_account` is indeed included. 

I think it might be as easy as updating the `key_kind` method https://github.com/radixdlt/sargon/blob/b9d99958230ffc6e10c64cd376679d2357ab7ec0/crates/sargon/src/profile/v100/factors/hierarchical_deterministic_factor_instance.rs#L79
and return `CAP26Path::Account` instead of `None` for `DerivationPath::BIP44Like`
Ensure to document this, something along those lines: "// It is not possible to create any BIP44 for Personas, all BIP44 key are used for Legacy Accounts for TransactionSigning. We note that we return a type called CAP26 for a non-CAP26-type such as BIP44 path, which is an inaccuracy. Alas, this feels OK and we can change `CAP26KeyKind` -> `KeyKind` later if we want."

you will need to update the name and expected value of this unit test:
https://github.com/radixdlt/sargon/blob/b9d99958230ffc6e10c64cd376679d2357ab7ec0/crates/sargon/src/profile/v100/factors/hierarchical_deterministic_factor_instance.rs#L222

Hopefully that is all you need to do.

crates/sargon/src/profile/logic/persona/query_personas.rs
```diff
@@ -0,0 +1,35 @@
+use crate::prelude::*;
+
+impl Profile {
+    /// Returns the non-hidden personas on the current network, empty if no personas
+    /// on the network
+    pub fn personas_on_current_network(&self) -> Result<Personas> {
+        self.current_network().map(|n| n.personas.non_hidden())
```
CyonAlexRDX, line 7: @sergiupuhalschi-rdx sync with @micbakos-rdx about this :D who ever merges first... https://github.com/radixdlt/sargon/blob/0e4ba5a48cc0e2923ae94f4ae57ebf5c2e5c1fa7/crates/sargon/src/profile/logic/account/query_personas.rs#L6

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -0,0 +1,759 @@
...
+            message,
+            is_wallet_transaction,
+            nonce,
+            notary_public_key,
+        )
+        .await
+    }
+}
+
+/// Part of the error message indicating the deposits are denied for the account.
```
CyonAlexRDX, line 33: @sergiupuhalschi-rdx can you update this doc, to also document that it is an error message returned by Gateway.

```rust
/// This is part of an error message returned **by Gateway**, indicating the deposits are denied for the account.
/// We use it part of logic below, matching against this String - we really should upgrade this code to be more
/// structure - we MUST update this value if Gateway where to change this value.
```

or something like that. And same for `GW_ERR_NOT_ALL_COULD_BE_DEPOSITED`

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -0,0 +1,759 @@
...
+            .extract_transaction_signer_entities(manifest_summary.clone())
+            .await?;
+        Ok(self
+            .extract_transaction_signers_factor_instances(signer_entities)
+            .iter()
+            .map(|fi| fi.public_key)
+            .collect())
+    }
+
+    async fn extract_transaction_signer_entities(
```
CyonAlexRDX, line 178: @sergiupuhalschi-rdx @micbakos-rdx important, this method should be moved into the `SignatureCollector` when @micbakos-rdx merges #235 , see here is the near identical code in in that PR (wait for it... GH was slow expanding the link for me): https://github.com/radixdlt/sargon/pull/235/files#diff-2d257ce713ef9ec886a81af3229aae3b41a3e312b8d8ba2a35eccb5ef145bb13R29

when in doubt prefer @micbakos-rdx s implementation. Ensure to test hidden entities... @GhenadieVP should Sargon fail if any of the entities "requiring auth" is hidden?

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -0,0 +1,759 @@
...
+            result,
+            Err(CommonError::ReservedInstructionsNotAllowedInManifest {
+                reserved_instructions: "AccountLockFee".to_string()
+            })
+        )
+    }
+
+    #[actix_rt::test]
+    async fn success() {
+        let responses = prepare_responses(
```
CyonAlexRDX, line 634: same here.. with `zero`. perhaps put the `let zero = RETDecimal::ZERO;` inside the `mod tests` and let it be used by all tests.

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
sergiupuhalschi-rdx: `let ret_zero: AsStr<Decimal> = Decimal::ZERO.into();`
This is the best I could do, am I missing something? Please let me know if there is a better and simpler way to do this.

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
GhenadieVP: > should Sargon fail if any of the entities "requiring auth" is hidden

No, the respective entity would be considered to be a foreign entity.

crates/sargon/src/system/sargon_os/sargon_os_transaction.rs
sergiupuhalschi-rdx: FYI `1` is used on Android only for the transaction preview gateway request, while for the submission `10` is being used same as on iOS.

crates/sargon/src/signing/collector/signatures_collector_preprocessor.rs
```diff
@@ -29,7 +29,53 @@ pub(crate) fn sort_group_factors(
         .collect::<IndexSet<_>>()
 }
 
+pub struct ExtractorOfInstancesRequiredToSignTransactions;
+impl ExtractorOfInstancesRequiredToSignTransactions {
```
CyonAlexRDX, line 33: Let's move
This into a separate file, and add docs and a few unit test for this specifically ? You can use our manifests builders to build manifests requiring auth.

Use the sample values of the SignaturesCollector with securified accounts examples.

Michael created a function which allows you to create manifests requiring auth of personas and accounts

crates/sargon/src/signing/tx_to_sign.rs
```diff
@@ -31,7 +31,22 @@ impl TXToSign {
...
+                profile,
+                intent.manifest_summary().clone(),
+            )?;
+
+        Ok(Self::with(intent_hash, entities_requiring_auth))
+    }
+}
+
+pub struct ExtractorOfEntitiesRequiringAuth;
+impl ExtractorOfEntitiesRequiringAuth {
```
CyonAlexRDX, line 45: Move to separate file, document and add a few unit tests

GhenadieVP (approved):
💪

PR #232, opened 2024-10-09
Git config author in release script

CyonAlexRDX:
Trying to rid of the `Anka` author of our tags in our releases.

For more context see thread: https://rdxworks.slack.com/archives/C06EBEA0SGY/p1728481132742009?thread_ts=1728401660.134659&cid=C06EBEA0SGY


matiasbzurovski (approved):
I think it looks good, although bash is not my greatest strength neither :)

PR #237, opened 2024-10-14
Uniffi separation

GhenadieVP:
Separate Uniffi into a crate:

- Remove any uniffi related code from the sargon crate.
- Redefine the models in sargon-uniffi crate.
- Map From/Into internal models.
- Wrap internal SargonOs and re-expose the methods.
- Defined how an uniffii trait can be mapped to an internal trait.


crates/sargon/src/core/types/bag_of_bytes.rs
CyonAlexRDX, line 1: `BagOfBytes` is a mere `UniFFI` compat construct, so I think we should perhaps delete it from `sargon`. And we should use `Vec<u8>` in this package instead. And then in `sargon-uniffi` we do  conversion.

crates/sargon/src/core/types/exactly_n_bytes.rs
CyonAlexRDX, line 1: So `ExactlyNBytes` types - exactly like `BagOfBytes` - exists only for UniFFI compatability. So we should move them all from `sargon` into `sargon-uniffi`, and in Sargon we will simply use `[u8; N]` directly.

I think, or am I forgetthing anything?

CyonAlexRDX: yeah maybe we like the new types around those and having sample valeus of those.... right. 

hmm. Ok, but we should be able to have ultra ultra lightweight newtypes then :) 

```rust
#[derive(Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
#[serde(transparent)]
pub struct Exactly32Bytes(pub [u8; 32]);
impl HasSampleValues for Exactly32Bytes { ... }
```

crates/sargon/src/core/types/secret_bytes.rs
```diff
@@ -9,139 +9,51 @@ macro_rules! decl_secret_bytes {
...
-                fn from_custom(obj: Self) -> Self::Builtin {
-                    BagOfBytes::from(obj.0.as_slice())
-                }
-            }
-
             $(
                 #[doc = $expr]
             )*
-            #[derive(Zeroize, derive_more::Debug, derive_more::Display, uniffi::Record)]
+            #[derive(Zeroize, Clone, Eq, PartialEq, derive_more::Debug, derive_more::Display)]
```
Sajjon, line 15: @GhenadieVP check if you can add `ZeroizeOnDrop` which is the only REALLY useful trait, but UniFFI prevented me from adding it.

crates/sargon/src/hierarchical_deterministic/bip39/bip39_word/bip39_word.rs
```diff
@@ -1,7 +1,7 @@
 use crate::prelude::*;
 
 /// A word in the BIP39 word list of `language` at known `index` (0-2047).
-#[derive(Zeroize, Clone, Debug, PartialEq, Eq, Hash, uniffi::Record)]
+#[derive(Zeroize, Clone, Debug, PartialEq, Eq, Hash)]
```
Sajjon, line 4: @GhenadieVP try find replace all "Zeroize" -> "ZeroizeOnDrop, Zeroize" in the whole `sargon` crate, or otherwise manually go through all derives and try adding ZeroizeOnDrop.

crates/sargon/src/profile/v100/address/legacy_olympia_account_address.rs
```diff
@@ -20,10 +19,11 @@ use crate::prelude::*;
 )]
 #[display("{}", self.bech32_address())]
 #[debug("{}", self.bech32_address())]
-pub struct LegacyOlympiaAccountAddressSecretMagic {
-    public_key: Secp256k1PublicKey,
+pub struct LegacyOlympiaAccountAddress {
+    pub value: Secp256k1PublicKey,
```
Sajjon, line 23: I think it makes sense to keep the name `public_key` to indicate that this address can be parsed from and into a public_key.

crates/sargon/src/core/types/bag_of_bytes.rs
GhenadieVP: In Sargon BagOfBytes is still useful as wrapper around Vec<u8> to provide additional functionality, such as initializing from hex, converting to hex, have ZeroizeDrop and so on. I am not sure we can just remove it.

crates/sargon/src/core/types/exactly_n_bytes.rs
GhenadieVP: same for these, we do additional functionality, it is not just a wrapper for uniffi.

crates/sargon/src/core/types/secret_bytes.rs
GhenadieVP: Right, thanks for reminding me

crates/sargon/src/profile/v100/address/wrap_ret_address.rs
Sajjon, line 1: I wonder if we should try removing all of our address types later on. And using `RET` addresses directly, or with a VERY THIN new type wrapper around each perhaps some traits to augment them with some methods if we want, and thin wrapper doing serde...

crates/sargon/src/profile/v100/entity/persona/persona_data/persona_data_identified_collection_types.rs
```diff
@@ -28,7 +28,6 @@ macro_rules! declare_collection_of_identified_entry {
             Eq,
             derive_more::Display,
             derive_more::Debug,
-            uniffi::Record,
```
Sajjon, line 31: We do not need to do it now, but this whole macro should be removed. This is one of many macros which can now be replaced with generics thx to not being UniFFI compatible anymore.

micbakos-rdx:
@GhenadieVP Impressive work!. If you want to build and run tests for Android too the only change you need to apply to generate bindings is to add in `jvm/sargon-android/build.gradle.kts` the following line in `L284`
```
"-p", "sargon-uniffi",
```

So the exec command will look like this
```
exec {
    workingDir = rootDir.parentFile
    commandLine(
        "cargo", "run",
        "-p", "sargon-uniffi",
        "--features", "build-binary",
        "--bin", "sargon-bindgen",
        "generate", "--library", binaryFile.toString(),
        "--language", "kotlin",
        "--out-dir", "${buildDir}/generated/src/kotlin"
    )
}
```

crates/sargon-uniffi/src/core/error/common_error.rs
```diff
@@ -0,0 +1,771 @@
...
+//     fn into(self) -> InternalCommonError {
+//         unimplemented!()
+//     }
+// }
+
+#[repr(u32)]
+#[derive(
+    Clone, Debug, ThisError, PartialEq, InternalConversionV2, uniffi::Error,
+)]
+pub enum CommonError {
```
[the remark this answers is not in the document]
Sajjon: @GhenadieVP have you bumped UniFFI? works on non enum types too: https://github.com/mozilla/uniffi-rs/blob/c344737d7c72a9ffedc08f840601318f2a16d4e9/docs/manual/src/udl/errors.md?plain=1#L102 looks like it at least

crates/sargon/src/core/types/bag_of_bytes.rs
Sajjon: sure!

crates/sargon-uniffi/src/core/hash.rs
```diff
@@ -0,0 +1,54 @@
...
+/// exposed in Swift/Kotlin as its own struct/data class, with
+/// hidden secret magic.
+#[derive(
+    Clone, PartialEq, Eq, std::hash::Hash, InternalConversion, uniffi::Record,
+)]
+pub struct Hash {
+    pub(crate) value: Exactly32Bytes,
+}
+
+impl From<InternalHash> for Hash {
```
Sajjon, line 17: is the `impl From<InternalHash> for Hash`  and `impl Into<InternalHash> for Hash` needed ? Cannot `InternalConversion` proc macro generate that for us?

Sajjon: maybe it can but you just have not done it yet?

crates/sargon-uniffi/src/system/sargon_os/sargon_os.rs
```diff
@@ -0,0 +1,72 @@
...
+            internal_bios,
+        ))
+        .await;
+        Arc::new(SargonOS {
+            wrapped: internal_sargon_os,
+        })
+    }
+
+    pub async fn new_wallet(&self) -> Result<()> {
+        self.wrapped.new_wallet().await.into_result()
```
CyonAlexRDX, line 33: @GhenadieVP we might wanna `impl Deref<Target = InternalSargonOS> for SargonOS`, which will allow `self.new_wallet().await.into_result()`, i.e. we can skip `.wrapped` for all methods. which might we worth it. And not only here, potentially in other places inside of `sargon-uniffi`.

CyonAlexRDX: @GhenadieVP there is also the [`delegate` crate](https://crates.io/crates/delegate) (super cool crate) [which we are using in `Decimal`](https://github.com/radixdlt/sargon/blob/8540902716af79fd375eff3a749b84a384decdcc/crates/sargon/src/core/types/decimal192.rs#L456)

which might allow use to forward ALL methods in a very convenient manner, **with the `into_inner` calls**, something like:

```rust
#[uniffi::export]
impl SargonOS {
    delegate! {
        to self.wrapped {
            pub async fn new_wallet(&self) -> Result<()>;
            pub async fn import_wallet(&self, profile: &Profile,  bdfs_skipped: bool) -> Result<()>;
// just list the signature! :D 
        }
    }
} 
```

crates/sargon-uniffi/src/core/result/from_internal_result.rs
Sajjon, line 1: empty file?

CyonAlexRDX (approved):
Amazing job! Great first iteration! One of the more
Important PRs yet

crates/sargon-uniffi/src/core/hash.rs
GhenadieVP: The macro cannot do it yet, to allow conversion of internal types that do not match in structure the types defined in uniffi. These are few, but definitely worth in future iterations to add such capability.

crates/sargon-uniffi/src/system/sargon_os/sargon_os.rs
GhenadieVP: Sadly this does not work as we have discussed, but I have plans of writing a custom macro that can achieve the same, but in a separate PR :)

crates/sargon-uniffi/src/core/error/common_error.rs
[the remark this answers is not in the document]
GhenadieVP: well it does not work, the [expand_error function explicitly expects an enum](https://github.com/mozilla/uniffi-rs/blob/c344737d7c72a9ffedc08f840601318f2a16d4e9/uniffi_macros/src/error.rs#L17)

sergiupuhalschi-rdx (approved):
Definitely a huge improvement. Cargo check is much faster when run only on the sargon crate. Can't wait for it to be merged!

PR #236, opened 2024-10-11
[ABW-3838] Add TX V2 RET models

danvleju-rdx:
JIRA ticket: [ABW-3838](https://radixdlt.atlassian.net/browse/ABW-3838)

### Description
- Added transaction V2 models required for Pre-authorization.
- Aligned V1 models with the latest Scrypto and RET updates.
- Enhanced `ManifestSummary` to include information about deposits and withdrawals from the static analysis.
- Removed `radix-engine-toolkit-json`, previously used for parsing `NonFungibleGlobalId` from a String. This parsing is now handled within Sargon.

[ABW-3838]: https://radixdlt.atlassian.net/browse/ABW-3838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #238, opened 2024-10-15
[ABW-3835] Pre-Authorization request models

matiasbzurovski:
JIRA ticket: [ABW-3835](https://radixdlt.atlassian.net/browse/ABW-3835)

## Dependencies
This PR depends on https://github.com/radixdlt/sargon/pull/236. It is opened against it, to allow checking the actual difference.

## Changelog
Adds the Pre-Authorization request models, defined in [CAP-39](https://radixdlt.atlassian.net/wiki/spaces/AT/pages/3702816769/CAP-39+Pre-authorization+request).

[ABW-3835]: https://radixdlt.atlassian.net/browse/ABW-3835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/sargon/src/radix_connect/wallet_interaction/dapp_wallet_interaction/dapp_to_wallet/interaction_items/pre_authorization/expiration.rs
```diff
@@ -0,0 +1,66 @@
...
+/// An enum that represents the different ways a subintent can expire.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, uniffi::Enum)]
+#[serde(tag = "discriminator", content = "value", rename_all = "camelCase")]
+pub enum DappToWalletInteractionSubintentExpiration {
+    /// The subintent expires at a specific time.
+    #[serde(rename = "expireAtTime")]
+    AtTime(Timestamp),
+
+    /// The subintent expires X seconds after its signature.
+    #[serde(rename = "expireAfterSignature")]
```
CyonAlexRDX, line 12: not your model I guess, but this is really a strange name... signing is a 100% offline action. An Elliptic Curve Signature has no anchor in any system, alas this name makes it sound like it has. So I feel this is completely wrong.

Or have we built a system where Dapp receiving signed subitents will check the expiry and SHOULD skip using it if `AfterSignature.0` time has passed ? Some dapps can technically chose to ignore this right? Or is it done by our own tooling, in RDT?

matiasbzurovski: You are right that signing is a 100% offline action. It isn't actually expiring after the signature, but after the submission of the response (which has been signed). Considering the submission takes place after signing it, I don't have a big issue against the naming.


Regarding how it works, I don't think it is the dApp responsibility to check the expiration. It is just that the _parent transaction_ would fail if at least one of the subintents has already expired.

For example, a dApp sends a subintent for `User A` to approve sending 100 XRD with 1 hour expiration, and a subintent for `User B` to approve sending 2 USDT with same expiration.

If both users sign their subintents within one hour from each other, the transaction to exchange 100 XRD over 2 USDT will succeed. Otherwise, it would fail.

Note that with this expiration, the TX will be valid if both users decide to not check their wallet for N days, and open it to approve the subintent at the same time (with a 1 hour window). If the dApp wants to prevent this, it would use the other `DappToWalletInteractionSubintentExpiration::ExpireAtTime`. But in any case, it wouldn't be the dApps responsibility to check/ignore this: the TX would fail if any of the children doesn't meet the expiration.

Sajjon: We should call this `SubmissionSyncWindowForParties` really. It is about submission, not signing, so we should use the word lemma "submit" not "sign". And it is a sync window for parties - as you described above, all parties have to sign within the that time window.

Sajjon: and what you described should be put as doc on the variant :)

CyonAlexRDX: AllSuubintentsSignedWithin

matiasbzurovski: well, to be honest, I prefer `Expiration` over `SubmissionSyncWindowForParties`. I would keep the enum as is, since it is also the name used on the JSON field. 

I believe the main confusion is on the `AfterSignature` variant. I would refer to the other subintents (`AllSubintents..`) since it is this subintent the one that will expire. Maybe we we can rename to `TimeAfterSubmission` instead?

Sajjon: I mean to say that the enum should continue be named `Expiration` but this variant (case) `AfterSignature` ought to be renamed `AllSubintentsSubmittedWithin`

sergiupuhalschi-rdx (commented):
Looks good!

sergiupuhalschi-rdx (approved):

crates/sargon/src/radix_connect/wallet_interaction/dapp_wallet_interaction/dapp_to_wallet/interaction_items/pre_authorization/pre_authorization.rs
```diff
@@ -0,0 +1,44 @@
+use crate::prelude::*;
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, uniffi::Record)]
+pub struct DappToWalletInteractionPreAuthorizationItems {
+    pub subintent: Option<DappToWalletInteractionSubintentRequestItem>,
```
GhenadieVP, line 5: optional?

GhenadieVP (commented):
LGTM! not sure about the subintent being optional in the request interaction.

danvleju-rdx (approved):

crates/sargon/src/wrapped_radix_engine_toolkit/high_level/manifest_building/assert_manifest.rs
```diff
@@ -56,3 +56,33 @@ pub fn instructions_eq(
...
+    instructions: InstructionsV2,
+    expected: impl AsRef<str>,
+) {
+    let trim =
+        |s: &str| s.replace(' ', "").replace('\t', " ").trim().to_owned();
+    pretty_assertions::assert_eq!(
+        trim(&instructions.to_string()),
+        trim(expected.as_ref())
+    );
+}
```
danvleju-rdx, line 88: Should not be needed. I've removed this fn on my branch.

PR #241, opened 2024-10-18
Use TransactionIntent instead of IntentHash

micbakos-rdx:
* Some key places of signatures collector have been changed to support `TransactionIntent`s and not just `IntentHash`es.
* New type that wraps the compiled intent has been created called `CompiledTransactionIntent` which is the result of the `TransactionIntent::compile` function. Decompile is available by forwarding the call to RET.
* The key change is in `PetitionForTransaction`. Each such petition contains a unique `TransactionIntent`. Such transaction intents are not kept around in other files in memory.
* Each `PetitionForEntity` still references an `IntentHash`
* In order to facilitate such change the intermediary types had to change:
    * `TXToSign` now keeps a `TransactionIntent`. Since `TransactionIntent` is not `Hash` it was impossible to use `IndexSet` so to ensure uniqueness `TXToSign` conforms to `Identifiable` and the ID is the computed `IntentHash`.
    * `TransactionSignRequestInput` is the one that should be passed into uniffi layer. I took the initiative to use `CompiledTransactionIntent` instead of `TransactionIntent` since the payload should be smaller and the host can always decompile before signing.
    
    
 > [!Tip]
 > The base branch is `ABW-3873-rola`


crates/sargon/src/signing/collector/signatures_collector.rs
```diff
@@ -95,12 +95,12 @@ impl SignaturesCollector {
         let transactions = transactions
             .into_iter()
             .map(extract_signers)
-            .collect::<Result<IndexSet<TXToSign>>>()?;
+            .collect::<Result<Vec<TXToSign>>>()?;
```
CyonAlexRDX, line 98: did you get a compilation error when using  `.collect::<Result< IdentifiedVecOf <TXToSign>>>()?` (directly, instead of via `Vec`)?

crates/sargon/src/signing/tx_to_sign.rs
```diff
@@ -1,20 +1,28 @@
 use crate::prelude::*;
 
-#[derive(Clone, Debug, PartialEq, Eq, std::hash::Hash)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub(crate) struct TXToSign {
-    pub(crate) intent_hash: IntentHash,
+    pub(crate) intent: TransactionIntent,
     entities_requiring_auth: Vec<AccountOrPersona>, // should be a set but Sets are not `Hash`.
```
CyonAlexRDX, line 6: ah, you can now change `Vec<AccountOrPersona>` into an `IndexSet<AccountOrPersona>` :)

crates/sargon/src/signing/tx_to_sign.rs
```diff
@@ -1,20 +1,28 @@
...
 }
 
+impl Identifiable for TXToSign {
+    type ID = IntentHash;
+
+    fn id(&self) -> Self::ID {
+        self.intent.intent_hash()
```
CyonAlexRDX, line 13: minor optimization is to eagerly calc the `intent_hash` inside the ctor, and keep it as a private field, so that we need not calculate it multiple times like we do here now.

micbakos-rdx: Right was thinking about it. 💯

crates/sargon/src/types/samples/transaction_intent_samples.rs
```diff
@@ -1,12 +1,13 @@
...
 use radix_engine_toolkit::models::canonical_address_types::NetworkId;
 use reqwest::Identity;
 
 impl TransactionIntent {
     /// Returns a sample intent that its transaction summary will involve all the
     /// `accounts_requiring_auth` and `personas_requiring_auth` in entities requiring auth.
     /// This can be accomplished by building a manifest that constructs owner keys from these
-    /// entities
+    /// entities. All entities set the same `PublicKeyHash` for the sake of simplicity.
     pub fn entities_requiring_auth<'a, 'p>(
```
CyonAlexRDX, line 11: maybe this ctor can be prefixed with `sample_` ?

crates/sargon/src/types/samples/transaction_intent_samples.rs
```diff
@@ -21,55 +22,63 @@ impl TransactionIntent {
     /// `account_addresses_requiring_auth` and `identity_addresses_requiring_auth` in
     /// entities requiring auth.
     /// This can be accomplished by building a manifest that constructs owner keys from these
-    /// entity addresses.
+    /// entity addresses. All entities set the same `PublicKeyHash` for the sake of simplicity.
     pub fn new_requiring_auth(
```
CyonAlexRDX, line 26: maybe prefix with `sample_` ?

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/compiled_transaction_intent.rs
```diff
@@ -0,0 +1,106 @@
...
+    Deserialize,
+    Clone,
+    Debug,
+    PartialEq,
+    Eq,
+    Hash,
+    derive_more::Display,
+    derive_more::FromStr,
+)]
+pub struct CompiledTransactionIntent(pub BagOfBytes);
```
CyonAlexRDX, line 14: I think we would wanna try hiding the constructor. And then move the `impl TransactionIntent` method this file (`pub fn compile(&self) -> CompiledTransactionIntent {`)

This lacks UniFFI compatability which will be needed if we are gonna use it in Interactors.

Not sure what is easiest with @GhenadieVP s branch, if it is a newtype or if it is a single fields object. but `BagOfBytes` need be private.
And make `pub fn new` non-pub, so that the only way of creating `CompiledTransactionIntent` (outside of this file) is to call `compile` on a `TransactionIntent`.

crates/sargon/src/signing/collector/signatures_collector.rs
micbakos-rdx: No just dumb me.

CyonAlexRDX (approved):
Nice! Just think we ought to hide the CTOR of `CompiledTransactionIntent`.

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/compiled_transaction_intent.rs
micbakos-rdx: Thanks @CyonAlexRDX did the changes you mentioned. Additionally I am planning to add uniffi support for `CompiledTransactionIntent` like this:

```
use crate::prelude::*;
use sargon::CompiledTransactionIntent as InternalCompiledTransactionIntent;

uniffi::custom_newtype!(CompiledTransactionIntent, BagOfBytes);

#[derive(Clone, PartialEq, Eq, InternalConversion)]
pub struct CompiledTransactionIntent(BagOfBytes);

#[uniffi::export]
pub fn new_compiled_transaction_intent_sample() -> TransactionIntent {
    InternalCompiledTransactionIntent::sample().into()
}

#[uniffi::export]
pub fn new_compiled_transaction_intent_sample_other() -> TransactionIntent {
    InternalCompiledTransactionIntent::sample_other().into()
}

#[uniffi::export]
pub fn compiled_transaction_intent_decompile(compiled_intent: &CompiledTransactionIntent) -> TransactionIntent {
    compiled_intent.into_internal().decompile().into()
}
```
@GhenadieVP Do you think that is sufficient? Also @CyonAlexRDX I used `custom_newtype` like we do with `Epoch `

micbakos-rdx: Hmm it seems that private fields are not accessible from uniffi with `InternalConversion`.

micbakos-rdx: Revised implementation, with the downside that the internal type has a public constructor. @CyonAlexRDX 
```
uniffi::custom_newtype!(CompiledTransactionIntent, BagOfBytes);

#[derive(Clone, PartialEq, Eq)]
pub struct CompiledTransactionIntent(BagOfBytes);

impl From<InternalCompiledTransactionIntent> for CompiledTransactionIntent {
    fn from(value: InternalCompiledTransactionIntent) -> Self {
        Self(value.bytes().into())
    }
}

impl CompiledTransactionIntent {
    pub fn into_internal(&self) -> InternalCompiledTransactionIntent {
        self.clone().into()
    }
}

impl Into<InternalCompiledTransactionIntent> for CompiledTransactionIntent {
    fn into(self) -> InternalCompiledTransactionIntent {
        InternalCompiledTransactionIntent::new(self.0.into())
    }
}

#[uniffi::export]
pub fn new_compiled_transaction_intent_sample() -> CompiledTransactionIntent {
    InternalCompiledTransactionIntent::sample().into()
}

#[uniffi::export]
pub fn new_compiled_transaction_intent_sample_other() -> CompiledTransactionIntent {
    InternalCompiledTransactionIntent::sample_other().into()
}

#[uniffi::export]
pub fn compiled_transaction_intent_decompile(compiled_intent: &CompiledTransactionIntent) -> TransactionIntent {
    compiled_intent.into_internal().decompile().into()
}
```

micbakos-rdx: I think we nailed the export by doing the following:

* The internal type's constructor is public, because it is needed by uniffi crate, but marked as `unsafe` so no-one can use it by accident.
* The uniffi type is exported as a record, and not a type alias, but with `secret_magic` field that is post processed as not public in kotlin and swift.

crates/sargon-uniffi/src/wrapped_radix_engine_toolkit/low_level/compiled_transaction_intent.rs
```diff
@@ -0,0 +1,47 @@
+use crate::prelude::*;
+use sargon::CompiledTransactionIntent as InternalCompiledTransactionIntent;
+
+#[derive(Clone, PartialEq, Eq, uniffi::Record)]
+pub struct CompiledTransactionIntent {
```
GhenadieVP, line 5: use `decl_conversion_tests_for!(CompiledTransactionIntent)` to generate the roundtrip conversion tests.

crates/sargon-uniffi/src/wrapped_radix_engine_toolkit/low_level/compiled_transaction_intent.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl CompiledTransactionIntent {
+    pub fn into_internal(&self) -> InternalCompiledTransactionIntent {
+        self.clone().into()
+    }
+}
+
+impl Into<InternalCompiledTransactionIntent> for CompiledTransactionIntent {
+    fn into(self) -> InternalCompiledTransactionIntent {
+        unsafe {
```
GhenadieVP, line 25: `unsafe`?

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/compiled_transaction_intent.rs
```diff
@@ -0,0 +1,123 @@
...
+)]
+pub struct CompiledTransactionIntent(BagOfBytes);
+
+impl CompiledTransactionIntent {
+    /// ONLY Use this in a test or when converting from the homonymous uniffi type.
+    ///
+    /// # Safety
+    /// This is completely unsafe, a random `BagOfBytes` cannot construct a valid
+    /// `CompiledTransactionIntent` and cannot decompile to a valid `TransactionIntent`.
+    pub unsafe fn new(bytes: BagOfBytes) -> Self {
```
Sajjon, line 22: I realize we can make this `Result<Self>` right? you can do:

```rust
pub fn new(bytes: BagOfBytes) -> Result<Self> {
    RET_decompile_intent(bytes).map(Self).map_err(|e| ..)
}
```

crates/sargon-uniffi/src/wrapped_radix_engine_toolkit/low_level/compiled_transaction_intent.rs
micbakos-rdx: Changed to https://github.com/radixdlt/sargon/pull/241#discussion_r1806664802

giannis-rdx (approved):

GhenadieVP (approved):

CyonAlexRDX:
We can probably merge this?

sergiupuhalschi-rdx (approved):

micbakos-rdx:
> We can probably merge this?

Sure will do, prevented from doing so since I didn't want to "pollute" the ROLA PR.

PR #243, opened 2024-10-23
New `HDPathComponent` and HD Paths

CyonAlexRDX:
# Abstract
Changes 

```diff
-HDPathComponent(u32)
+enum HDPathComponent { Unsecurified(Unsecurified), Securified(SecurifiedU30) }
```

And simplified / cleans up all our HD Path implementations.

# Rationale

Soon I will migrate over `FactorInstancesProvider` from Sargon-MFA into Sargon, which contains logic which does `next free derivation entity index = max derivation entity index  + 1`, [see here](https://github.com/radixdlt/sargon-mfa/blob/8350c7be549ce802d0bba0fe59297ad368dac422/src/factor_instances_provider/next_index_assigner/next_derivation_entity_index_cache_analyzing_assigner.rs#L35).

This increment, `add_one` or [`add_n`](https://github.com/radixdlt/sargon-mfa/blob/8350c7be549ce802d0bba0fe59297ad368dac422/src/factor_instances_provider/next_index_assigner/next_derivation_entity_index_assigner.rs#L82) is flawed post MFA! Not that it is likely that someone would have 2^30 accounts (in fact no consumer phone/computer would be able to contain it, because our 1000 accounts test Profile is 2.3 MB so this would be a ~2500 PETA-byte large Profile 😅), but it is still semantically wrong, because it is theoretically possible to increase by one and go from Unsecurified KeySpace to Securified Keyspace

# `HDPathComponent` / U30 / U31 etc...

```diff
-pub type HDPathValue = u32;
-pub type HDPathComponent(HDPathValue);

+pub enum HDPathComponent {
+ Unsecurified(Unsecurified),
+ Securified(SecurifiedU30),
+}
```

```rust
pub enum HDPathComponent {
    Unsecurified(Unsecurified),
    Securified(SecurifiedU30),
}

/// A 30 bit unsigned integer
pub struct U30(u32);

/// A 31 bit unsigned integer
pub struct U31(u32);

pub struct SecurifiedU30(pub U30);

pub enum Unsecurified {
    Unhardened(Unhardened),
    Hardened(UnsecurifiedHardened),
}

pub struct Unhardened(pub U31);
pub struct UnsecurifiedHardened(pub U30);
```

There exists a "cousin" enum called `Hardened` which is used in as `index` in Account and Identity paths:

```rust
pub enum Hardened {
    Unsecurified(UnsecurifiedHardened),
    Securified(SecurifiedU30),
}
```

All these types hold a BIP32 path component value, ultimately a `u32` **without the offset added to it**. Meaning the `U30` inside the `UnsecurifiedHardened`, holds `5` for `5H`, **not** `5 + 0x80000000`. We construct all these using either `from_local_key_space(value: impl Into<U31>, extra_params: ExtraParam) -> Result` or using `from_global_key_space(value: u32) -> Result`.

If we want to construct `5H` as `UnsecurifiedHardened` using `from_global_key_space`, we would do:
```rust
assert_eq!(
    UnsecurifiedHardened::from_global_key_space(
            5 + GLOBAL_OFFSET_HARDENED
        )
        .unwrap()
        .to_string(),
    "5H"
);

```

Or alternatively with `from_local_key_space`

```rust
assert_eq!(
    UnsecurifiedHardened::from_local_key_space(U30::from(5u16))
        .unwrap()
        .to_string(),
    "5H"
);
```

Ultimately we wanna use a SLIP10 implementation to derive an actual key from the path, thus we must map our `HDPath` which just holds a `Vec<HDPathComponent>` into a `Vec<u32>`, we this by calling `map_to_global_key_space` on these types.

```rust
assert_eq!(
    UnsecurifiedHardened::from_local_key_space(U30::from(5u16))
        .unwrap()
        .map_to_global_key_space(),
    5 + GLOBAL_OFFSET_HARDENED
);
```

> [!IMPORTANT]
> Please note that the otherwise max hardened value now fails to deserialise `2147483647H`, because it is `U30::MAX` to large for a `H` - hardened unsecurified space. To get the same global value value, we denote it `1073741823S`. I updated out test vectors to accommodate this. https://github.com/radixdlt/sargon/commit/d401ee7003036dcd3f5e685d36ffa750cfa82b67#diff-5969d03c17c3c4c43c24da98f0080fb4f6f37f0ee3ad9b562cd020e674b5c0c4R748
> PLEASE LET ME KNOW if you want me to create a lenient JSON (and/or FromStr) parser which automatically maps -> `n H` -> `n-U30::MAX S` space. I think this is fine because as noted above, no user has over 1 billion accounts....

# Paths

## Removed `CAP26Path`
I've removed the `enum CAP26Path`, I dont think we need it.

## Changed `DerivationPath`
I've changed `DerivationPath`:

```diff
pub enum DerivationPath {
-   CAP26 { value: CAP26Path },
+   Account { value: AccountPath },
+   Identity { value: IdentityPath },
    BIP44Like { value: BIP44LikePath },
}
```

## GetIDPath no `Serde` no UniFFI.
I've removed `GetIDPath` from UniFFI layer, we wont need it. It is only used to calculate the FactorSourceID of FactorSources, which Sargon should do and should not need be exposed to hosts. Furthermore we need never store it in JSON, so no need for Serde. Therefor no need to be inside `DerivationPath`.

## AccountPath / IdentityPath changes
```diff
pub struct AccountPath {
-    pub path: HDPath,
    pub network_id: NetworkID,
-    pub entity_kind: CAP26EntityKind,
    pub key_kind: CAP26KeyKind,
-    pub index: HDPathValue,
+    pub index: HDPathComponent,
}
```


crates/sargon-uniffi/src/hierarchical_deterministic/bip32/hd_path_component.rs
CyonAlexRDX, line 1: split into many files

crates/sargon/Cargo.toml
```diff
@@ -24,11 +24,17 @@ log = "0.4.20"
 # pretty_env_logger = "0.5.0"
 pretty_env_logger = { git = "https://github.com/seanmonstar/pretty-env-logger/", rev = "0e238400e18649415dc710c025e99c009a1bb744" }
 
-# derive_more = "1.0.0-beta.6"
-derive_more = { git = "https://github.com/JelteF/derive_more", rev = "1196b2dd7a366c06db621093884adbc379fc0f0a", features = [
+# derive_more = "1.0.0"
+derive_more = { git = "https://github.com/JelteF/derive_more", rev = "d7f5b9e94d024790682f6fc4dcca13941cce64c8", features = [
+  "add",
```
CyonAlexRDX, line 29: Derive more has become very powerful! We should use these ! `as_ref` and `deref` are very nice!

crates/sargon/src/hierarchical_deterministic/bip32/hd_path.rs
CyonAlexRDX, line 1: probably not so useful to review this with "split mode" probably best to review the new code in isolation.

crates/sargon/src/hierarchical_deterministic/bip32/hd_path_component.rs
CyonAlexRDX, line 1: probably not so useful to review this with "split mode" probably best to review the new code in isolation.

crates/sargon/src/hierarchical_deterministic/bip32/unsigned_ints/u30.rs
```diff
@@ -0,0 +1,171 @@
+use crate::prelude::*;
+
+#[derive(
+    Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Deref, AsRef,
+)]
+pub struct U30(pub u32);
```
CyonAlexRDX, line 6: missing doc

crates/sargon/src/hierarchical_deterministic/bip32/unsigned_ints/u31.rs
```diff
@@ -0,0 +1,181 @@
+use crate::prelude::*;
+
+#[derive(
+    Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Deref, AsRef,
+)]
+pub struct U31(pub u32);
```
CyonAlexRDX, line 6: missing doc

crates/sargon/src/hierarchical_deterministic/bip39/bip39_seed.rs
```diff
@@ -36,59 +39,97 @@ impl BIP39Seed {
...
 }
 
+pub trait Derivation:
+    Clone + Into<DerivationPath> + HasDerivationPathSchemeObjectSafe
+{
+    fn derivation_path(&self) -> DerivationPath {
+        self.clone().into()
+    }
+}
+impl<T: Clone + Into<DerivationPath> + HasDerivationPathSchemeObjectSafe>
```
CyonAlexRDX, line 128: This is a pattern we ought to start using, I recently learned it! Here we are saying that ALL types which impl `Clone + Into<DerivationPath> + HasDerivationPathSchemeObjectSafe` also impl `Derivation`

crates/sargon/src/hierarchical_deterministic/cap26/paths/traits.rs
```diff
@@ -0,0 +1,44 @@
...
+    fn get_entity_kind(&self) -> CAP26EntityKind;
+}
+
+impl<T: HasEntityKind> HasEntityKindObjectSafe for T {
+    fn get_entity_kind(&self) -> CAP26EntityKind {
+        T::entity_kind()
+    }
+}
+
+pub trait NewEntityPath: Sized {
```
CyonAlexRDX, line 17: split to separate file

crates/sargon/src/hierarchical_deterministic/cap26/paths/unvalidated_cap26_path.rs
```diff
@@ -0,0 +1,171 @@
+use crate::prelude::*;
+
+/// A derivation path consisting of CAP26 components, however, not validated
+/// as canonical.
```
CyonAlexRDX, line 4: explain I mean, not validated to be either AccountPath or IdentityPath

crates/sargon/src/hierarchical_deterministic/derivation/derivation_path.rs
CyonAlexRDX, line 1: declared with macro, if we ever want a `AnyAccountPath { CAP26(AccountPath), BIP44(BIP44Like) }` we can declare that with same macro. 

I'm open to not using a macro, but it was in fact simpler doing it with macro.

CyonAlexRDX, line 1: probably not so useful reviewing in "split" mode....

crates/sargon/src/profile/logic/profile_next_derivation.rs
```diff
@@ -67,7 +67,7 @@ impl Profile {
         kind: EntityKind,
         network_id: NetworkID,
         factor_source_id: FactorSourceIDFromHash,
-    ) -> HDPathValue {
```
CyonAlexRDX, line 70: these methods are soon to be deleted, replaced by the FactorInstancesProvider in a coming PR.

crates/sargon/src/profile/v100/factors/hd_transaction_signing_factor_instance.rs
```diff
@@ -1,25 +1,55 @@
 use crate::prelude::*;
 
+pub trait HasKeyKind {
```
CyonAlexRDX, line 3: move to seperate file

crates/sargon/src/profile/v100/factors/hd_transaction_signing_factor_instance.rs
```diff
@@ -1,25 +1,55 @@
...
+    fn key_kind(&self) -> CAP26KeyKind;
+}
+
+pub trait IsEntityPath:
```
CyonAlexRDX, line 7: move to separate file

crates/sargon/tests/vectors/main.rs
```diff
@@ -81,25 +63,14 @@ mod cap26_tests {
...
-                        curve: S::curve(), // will be `Secp256k1` for `cap26_secp256k1.json` vectors!
-                        hd_path: v.path.hd_path().clone(),
-                        derivation_path: DerivationPath::CAP26 {
-                            value: v.path.clone(),
-                        },
-                    };
-
-                    let derived = seed.derive_private_key(&derivation);
+                    let derived =
+                        seed.derive_private_key_curve(curve, v.path.clone());
```
CyonAlexRDX, line 73: much cleaner!

crates/sargon/src/hierarchical_deterministic/derivation/derivation_path.rs
```diff
@@ -1,88 +1,191 @@
...
+                }
+            }
+
+            impl FromBIP32Str for $union_name {
+                fn from_bip32_string(s: impl AsRef<str>) -> Result<Self> {
+                    let s = s.as_ref();
+                    Result::<Self>::Err(CommonError::InvalidBIP32Path { bad_value: s.to_owned() })
+                    $(
+                        .or($variant_type::from_bip32_string(s).map(Self::[< $variant_name:snake >]))
+                    )+
```
matiasbzurovski, line 83: correct me if I am wrong, but in Rust you can return
```Rust
Err(failure_path)
 .or(happy_path)
```
and it will try to perform the `happy_path` even though the `failure_path` is set first?

crates/sargon-uniffi/src/hierarchical_deterministic/bip44/bip44_like_path.rs
```diff
@@ -29,12 +29,14 @@ use sargon::BIP44LikePath as InternalBIP44LikePath;
 /// ```
 #[derive(Clone, PartialEq, Eq, Hash, InternalConversion, uniffi::Record)]
 pub struct BIP44LikePath {
-    pub path: HDPath,
```
CyonAlexRDX, line 32: no need to store the full path, we know the purpose of cointype...

crates/sargon-uniffi/src/hierarchical_deterministic/bip32/hd_path_component.rs
```diff
@@ -1,16 +1,822 @@
...
+    value: u32,
+) -> HDPathComponent {
+    InternalHDPathComponent::from_global_key_space(value)
+        .into_result()
+        .expect("Should always be able to create an HDPathComponent from a u32")
+}
+
+#[uniffi::export]
+pub fn hd_path_component_to_bip32_string(component: HDPathComponent) -> String {
+    Into::<InternalHDPathComponent>::into(component).to_bip32_string()
```
GhenadieVP, line 84: you can use `component.into_internal().to_bip32_string()`

CyonAlexRDX: nice

crates/sargon-uniffi/src/hierarchical_deterministic/bip32/hd_path_component.rs
```diff
@@ -1,16 +1,822 @@
...
+
+#[uniffi::export]
+pub fn hd_path_component_get_key_space(component: HDPathComponent) -> KeySpace {
+    Into::<InternalHDPathComponent>::into(component)
+        .key_space()
+        .into()
+}
+
+#[cfg(test)]
+mod tests5 {
```
GhenadieVP, line 118: While more tests cannot hurt, these really feel unnecessary given that uniffi is just a mapper layer.
We would need to find a way to add tests for functions without the need of copying the internal tests here.

crates/sargon-uniffi/src/hierarchical_deterministic/bip32/hd_path_component.rs
```diff
@@ -1,16 +1,822 @@
...
+        .index_in_local_key_space()
+        .into()
+}
+
+#[uniffi::export]
+pub fn hd_path_component_get_key_space(component: HDPathComponent) -> KeySpace {
+    Into::<InternalHDPathComponent>::into(component)
+        .key_space()
+        .into()
+}
```
GhenadieVP, line 115: Out of curiosity, would we need any of these in the Wallet late down the line? I assume this is needed now given that it affects the Wallet.

crates/sargon/src/hierarchical_deterministic/cap26/paths/account_path.rs
```diff
@@ -0,0 +1,439 @@
...
+/// ```
+///
+/// The `AccountPath` struct is parametrized by Radix network id and account index, but fixes the other
+/// constants in the path as follows:
+///
+/// ```text
+/// m / 44' / 1022' / NETWORK_ID' / 525' / 1460' / ACCOUNT_INDEX'
+/// ```
+///
+/// More generally:
```
sergiupuhalschi-rdx, line 22: Thanks for such a detailed explanation! 🙏

crates/sargon/src/hierarchical_deterministic/derivation/derivation_path.rs
Sajjon: yes that is what I'm doing. Sorry what is the review? :)

matiasbzurovski: wasn't a review, just a Rust question since I am surprised this works 🙂 
I was thinking of the `.or()` in Rust as a `??` in Swift, meaning I would have though this would always return an `Err`.

CyonAlexRDX: looks like there is an error in the documentation of `or` https://doc.rust-lang.org/std/result/enum.Result.html#method.or

it returns the first `Ok`

matiasbzurovski (approved):

micbakos-rdx (approved):

PR #245, opened 2024-10-28
Address extend

GhenadieVP:


Sajjon:
Did we get any performance boost btw?

GhenadieVP:
No improvement in performance

PR #246, opened 2024-10-28
Updates to Trancaction status polling

matiasbzurovski:
## Changelog
Updates the Transaction status polling logic so that, if the endpoint response is a failure, we poll again after the corresponding delay. 

In other words, a failure response will now be considered as a successful response with `TransactionStatusResponsePayloadStatus::Unknown`. This is the behavior that host apps had before using Sargon to poll status.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2718,7 +2718,7 @@ dependencies = [
 
 [[package]]
 name = "sargon"
-version = "1.1.35"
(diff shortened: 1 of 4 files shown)


CyonAlexRDX (approved):
LGTM

sergiupuhalschi-rdx (approved):

PR #247, opened 2024-10-28
[ABW-3886] Poll subintent status

matiasbzurovski:
JIRA ticket: [ABW-3866](https://radixdlt.atlassian.net/browse/ABW-3886)

## Changelog
Polls the given `SubintentHash` until it is either successfully committed within a Transaction or it expires.

[ABW-3866]: https://radixdlt.atlassian.net/browse/ABW-3866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/sargon-uniffi/src/system/sargon_os/sargon_os_transactions.rs
```diff
@@ -1,35 +0,0 @@
-use crate::prelude::*;
-
-// ==================
-// Submit Transaction
-// ==================
-#[uniffi::export]
```
matiasbzurovski, line 6: This file was no longer used, as its functiones were migrated to different files: `sargon_os_transaction_submit` & `sargon_os_transaction_status`.

It probably slipped after merging main into UniFFI separation branch.

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
```diff
@@ -0,0 +1,392 @@
...
+}
+
+// ==================
+// Poll PreAuthorization Status (Internal)
+// ==================
+impl SargonOS {
+    /// Polls the state of a Subintent until it is submitted
+    ///
+    /// It returns the `TransactionIntentHash`, but also the list of delays between each poll.
+    async fn poll_subintent_status_with_delays(
```
CyonAlexRDX, line 29: If much rather we put the logic in a shared implementation used by tx v1 and subintents.

Because this looks like big copy paste with same logic ? Or did I miss something?

matiasbzurovski: I've added a note [here](https://github.com/radixdlt/sargon/pull/247/files/bdbe33d2a28d5c2ba9e92d796ad0bec4282f428f#diff-927cc7b4a971b1591828417003462b78720089019f792be45afdcd0eadef7ce8R79) explaining that once we know for sure how the endpoint is going to work, we may want to reuse the same logic.

However, as we still don't know if it is gonna be a different endpoint/model, it doesn't make too much sense to extract the common logic yet.

crates/sargon/src/gateway_api/methods/transaction_methods.rs
```diff
@@ -1,5 +1,6 @@
 use crate::prelude::*;
 use radix_engine_toolkit_common::receipt::SerializableToolkitTransactionReceipt;
+use radix_transactions::model::IntentHash;
```
CyonAlexRDX, line 3: @matiasbzurovski delete this and use `ScryptoIntentHash` instead I think

matiasbzurovski: actually this import isn't necessary, since we need to use our internal `SubintentHash` (and not the one from `radix_transactions`).

I believe RustRover auto imported it when copy-pasting some code, will keep an eye on this

CyonAlexRDX (approved):
LGTM

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
```diff
@@ -0,0 +1,392 @@
...
+        intent_hash: SubintentHash,
+        expiration: Option<DappToWalletInteractionSubintentExpiration>,
+    ) -> Result<TransactionStatus> {
+        let (status, _) = self
+            .poll_subintent_status_with_delays(intent_hash, expiration)
+            .await?;
+
+        Ok(status)
+    }
+}
```
GhenadieVP, line 20: This function is not very useful IMO, unless you plan to extend it

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
```diff
@@ -0,0 +1,392 @@
...
+
+        loop {
+            // Check if the subinent hasn't expired already
+            if Timestamp::now_utc() > expiration_timestamp {
+                return Err(CommonError::ExpiredSubintent);
+            }
+
+            // Increase delay by 1 second on subsequent calls
+            delay_duration += DELAY_INCREMENT;
+            let sleep_duration = Duration::from_millis(delay_duration);
```
GhenadieVP, line 62: I am not entirely sure this logic is correct for subintent polling. We should likely take into account the expiration duration and adjust the delay accordingly. If say the dApp specified the expiration to be 25 seconds, and it submits the parent transaction at second 21 and it is committed fully to the network before expiration, say second 24, it seems that this polling will consider the subintent as expired, because we will have the following times for checking:
- t1 = 0
- t2 = 2
- t3 = 5
- t4 = 9
- t5 = 14
- t6 = 20
- t7 = 27

And the checking at t7 will hit the expiration check above while the subintent was actually submitted. Let me know if I reading this code wrongly :).

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
matiasbzurovski: We need an internal fn that also returns the delays to test it. If we are going to use this internal one (`poll_pre_authorization_status_with_delays`) from UniFFI, the `into_result()` conversion wouldn't be as straightforward right?

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
matiasbzurovski: good catch! Added this consideration to the expiration check

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
```diff
@@ -52,15 +52,17 @@ impl SargonOS {
...
             delay_duration += DELAY_INCREMENT;
             let sleep_duration = Duration::from_millis(delay_duration);
 
+            // Check if the subinent has already expired. We subtract the `sleep_duration` to consider the case
+            // where the subinent was submitted after last poll, but the `expiration_timestamp` was also reached before this poll.
+            // We must check the subinent status before assuming it had expired.
+            if Timestamp::now_utc().sub(sleep_duration) > expiration_timestamp {
+                return Err(CommonError::ExpiredSubintent);
+            }
+
```
GhenadieVP, line 65: so this would allow polling after `Timestamp::now_utc() > expiration_timestamp`, but how would this align with the countdown in UI? Where I would expect that we show the subintent is being expired if `Timestamp::now_utc() >= expiration_timestamp`

matiasbzurovski: the countdown in the UI, which only applies for the case of `Expiration::AtTime`, will be displayed before the subintent is signed.

This logic, to poll for its status, will be performed after the subintent is signed.

GhenadieVP: We would show the countdown in the polling slide up also AFAIK

crates/sargon/src/gateway_api/models/logic/response/ledger_state.rs
```diff
@@ -1,6 +1,7 @@
 use crate::prelude::*;
 
-#[cfg(test)]
+// TODO: Uncomment the following line once we don't need to mock response of GatewayClient::get_pre_authorization_status()
+//#[cfg(test)]
```
CyonAlexRDX, line 4: is this accurate? are we still mocking?

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
```diff
@@ -0,0 +1,360 @@
...
+        expiration: DappToWalletInteractionSubintentExpiration,
+    ) -> Result<(PreAuthorizationStatus, Vec<u64>)> {
+        let network_id = self.current_network_id()?;
+        let gateway_client = GatewayClient::new(
+            self.clients.http_client.driver.clone(),
+            network_id,
+        );
+
+        // We are going to play safe and leave an extra second to make sure we check the status one second after it has theoretically expired.
+        // This is to avoid considering expired a subintent that got committed in the last instant.
```
sergiupuhalschi-rdx, line 41: Nice one

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
```diff
@@ -0,0 +1,360 @@
...
+                    if has_expired {
+                        // If it has expired, we return the corresponding status.
+                        return Ok((PreAuthorizationStatus::Expired, delays));
+                    } else {
+                        // Otherwise, we determine the delay before next call.
+                        // This will either be the default delay or the remaining time until expiration (whatever is less).
+                        // Example: We have already polled 4 times for a subintent that expires after 10 seconds.
+                        // Seconds until expiration = 10 + 1 = 11
+                        // Accumulated delay = 0 + 2 + 3 + 4 = 9
+                        // Next polling should be in 5 seconds, but instead we will make it in 2 to check immediately after expiration.
```
sergiupuhalschi-rdx, line 85: Special thanks for the explanation

sergiupuhalschi-rdx (approved):

crates/sargon/src/system/sargon_os/pre_authorization/support/pre_authorization_status.rs
```diff
@@ -0,0 +1,12 @@
+use crate::prelude::*;
+
+/// This is the result of the Pre-Authorization status polling for a given Subintent.
+#[derive(Clone, Debug, PartialEq)]
+pub enum PreAuthorizationStatus {
+    /// The Pre-Authorization has successfully been submitted in a transaction.
+    /// The associated value corresponds to the IntentHash of the given Transaction.
+    Success(String),
```
GhenadieVP, line 8: Need to use a strong type

GhenadieVP (changes requested):

crates/sargon/src/gateway_api/models/logic/response/ledger_state.rs
matiasbzurovski: should have heard my own comment: uncommenting line 4

crates/sargon/src/core/utils/constants.rs
```diff
@@ -23,3 +23,7 @@ pub const GATEWAY_ENTITY_DETAILS_CHUNK_ADDRESSES: u64 = 20;
 /// Minimum XRD balance required for a fee payer to execute an account deletion transaction.
 /// Includes a safety margin above the estimated 3 XRD fee for maximum resource transfer.
 pub const MIN_REQUIRED_XRD_FOR_ACCOUNT_DELETION: f64 = 4.0;
+
+/// The delay increment among polling requests is set to 1 second.
+/// This means that there will be a 2s delay after first call, a 3s delay after second call, 4s after third and so on.
+pub const POLLING_DELAY_INCREMENT: u64 = 1;
```
CyonAlexRDX, line 29: Best put `_IN_SEC` as prefix. When reviewing code - outside of IDE that is - I think it is nice to be able to know more about the const based on its name.

crates/sargon/src/gateway_api/models/logic/response/transaction/status/transaction_status.rs
```diff
@@ -1,5 +1,89 @@
 use crate::prelude::*;
 
+#[cfg(test)]
+impl TransactionStatusResponse {
+    // Helper functions to create sample responses
+
+    pub fn sample_empty() -> TransactionStatusResponse {
```
CyonAlexRDX, line 7: Prefer `Self` in return position and in construction.

And also, since you construct so many `TransactionStatusResponse` here, I would probably have used a ctor!

And for case of `error_message: None` you can use a specific ctor, which calls `new` called `success`?

crates/sargon/src/core/utils/constants.rs
matiasbzurovski: I went back and foward between this and `POLLING_DELAY_INCREMENT_IN_SECONDS` at least 3 times 😛 will leave it with it.

crates/sargon/src/gateway_api/models/logic/response/transaction/subintent_status/subintent_status.rs
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    pub fn sample_committed_success_corrupt() -> Self {
+        Self::committed_success(None)
+    }
+
+    fn committed_success(
+        finalized_at_transaction_intent_hash: Option<String>,
+    ) -> Self {
+        Self {
```
CyonAlexRDX, line 28: use ctor?

crates/sargon/src/gateway_api/models/logic/response/transaction/subintent_status/subintent_status.rs
```diff
@@ -0,0 +1,34 @@
+use crate::prelude::*;
+
+#[cfg(test)]
+impl SubintentStatusResponse {
+    // Helper functions to create sample responses
+
+    pub fn sample_unknown() -> Self {
+        Self {
```
CyonAlexRDX, line 8: use ctor?

crates/sargon/src/gateway_api/models/types/request/transaction/status/transaction_status.rs
```diff
@@ -2,7 +2,7 @@ use crate::prelude::*;
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct TransactionStatusRequest {
-    /** Bech32m-encoded hash. */
+    /// Bech32m-encoded hash.
```
CyonAlexRDX, line 5: 👍

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
```diff
@@ -0,0 +1,360 @@
...
+    pub async fn poll_pre_authorization_status(
+        &self,
+        intent_hash: SubintentHash,
+        expiration: DappToWalletInteractionSubintentExpiration,
+    ) -> Result<PreAuthorizationStatus> {
+        let (status, _) = self
+            .poll_pre_authorization_status_with_delays(intent_hash, expiration)
+            .await?;
+
+        Ok(status)
```
CyonAlexRDX, line 18: Absolutely nothing wrong with the code:

```rust
let (status, _) = self
	.poll_pre_authorization_status_with_delays(intent_hash, expiration)
	.await?;

Ok(status)
```
that you have here.

But I prefer:

```rust
self
	.poll_pre_authorization_status_with_delays(intent_hash, expiration)
	.await
	.map(|status, _| status)
```

A bit shorter :) a more "functional" which we try to make use of in Sargon :)

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
```diff
@@ -0,0 +1,360 @@
...
+                        .finalized_at_transaction_intent_hash
+                        .ok_or(CommonError::Unknown)?;
+                    return Ok((
+                        PreAuthorizationStatus::Success(
+                            transaction_intent_hash,
+                        ),
+                        delays,
+                    ));
+                }
+                SubintentStatus::Unknown => {
```
CyonAlexRDX, line 69: I prefer as little code as possible inside of matches, especially when the number of lines is assymetrical - as in the variant `CommittedSuccess` returns quite early, but `Unknown` is 4x as many lines. Futhermore, we are in a tripple indendation here, a little bit pyramid of doom. But we note that we can just break out of the match in this variant, and continue the code outside of the match (but sill inside the loop), right? 

```rust
loop {
    match status {
        SubintentStatus::CommittesSuccess => { 
            ... return;
         }
        SubintentStatus::Unknown => { /* not done yet, continue outside of match */ }
    }
    // handle unknown here
    let accumulated_delay = delays.iter().sum::<u64>();
}

File: crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs [hunk 18]
```diff
@@ -0,0 +1,360 @@
...
+        &self,
+        expiration: DappToWalletInteractionSubintentExpiration,
+    ) -> u64 {
+        match expiration {
+            DappToWalletInteractionSubintentExpiration::AtTime(at_time) => {
+                let current_time = seconds_since_unix_epoch();
+                if at_time.unix_timestamp_seconds > current_time {
+                    at_time.unix_timestamp_seconds - current_time
+                } else {
+                    0 // Avoid overflow in case we check after expiration
```
CyonAlexRDX, line 119: 👍

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
```diff
@@ -0,0 +1,360 @@
...
+#[cfg(test)]
+mod poll_pre_authorization_status_with_delays {
+    use super::*;
+    use actix_rt::time::timeout;
+    use std::{future::Future, time::Duration};
+
+    #[allow(clippy::upper_case_acronyms)]
+    type SUT = SargonOS;
+    #[allow(clippy::upper_case_acronyms)]
+    type SSR = SubintentStatusResponse;
```
CyonAlexRDX, line 138: 👍

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
```diff
@@ -0,0 +1,360 @@
...
+            expiration_after_delay(10),
+        )
+        .await
+        .unwrap();
+
+        // The result is Success with the mocked TransactionIntentHash
+        assert_eq!(
+            result.0,
+            PreAuthorizationStatus::Success(
+                TransactionIntentHash::sample().to_string()
```
CyonAlexRDX, line 204: I like asserting exact values like this. Just wanna present an alternative solution, which you might or might not know if.

If it would be hard to know the exact expected value of the IntentHash, you can use the `matches!` macro, which is really really great in situation like this! You would then write like this:

```rust
assert!( // note, not using assert_eq! macro
    matches!(
        result.0,
        PreAuthoriziationStatus::Success(_)
    )
);
```

CyonAlexRDX (approved):
Still an approve - good job! 

I mostly left some Rust nit comments

crates/sargon/src/gateway_api/models/types/response/transaction/subintent_status/status.rs
```diff
@@ -0,0 +1,7 @@
+use crate::prelude::*;
+
+#[derive(Deserialize, Serialize, Clone, PartialEq, Eq, Debug)]
+pub enum SubintentStatus {
+    Unknown,
```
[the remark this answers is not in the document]
matiasbzurovski: There isn't actually doc for this enum, but only for the [usage](https://radix-babylon-gateway-api.redoc.ly/#operation/TransactionSubintentStatus) of it (which I added on `SubintentStatusResponse`). Will add something similar here as well

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
matiasbzurovski: right, I had used the `matches!` macro at other places where we don't care/know about the associated value. Makes sense here since the associated value is actually set on mocks (and no need to test our mocks)

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_pre_authorization_status.rs
```diff
@@ -0,0 +1,360 @@
...
+
+        let mut delay_duration = POLLING_DELAY_INCREMENT;
+
+        loop {
+            // Check the subintent status
+            let response = gateway_client
+                .subintent_status(SubintentStatusRequest::new(
+                    intent_hash.to_string(),
+                ))
+                .await?;
```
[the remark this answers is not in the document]
matiasbzurovski: Updated to default to `Unknown` if GW returns an error.

GhenadieVP (approved):

PR #249, opened 2024-10-29
Update CI to check version on both Cargo.toml

matiasbzurovski:
Check that version has been bumped on both `Cargo.toml` files.


CyonAlexRDX (approved):
LGTM

PR #255, opened 2024-11-06
[ABW-3899] Create subintent API

matiasbzurovski:
JIRA ticket: [ABW-3899](https://radixdlt.atlassian.net/browse/ABW-3899)

## Changelog
Exports function on SargonOS to create a Subinent

[ABW-3899]: https://radixdlt.atlassian.net/browse/ABW-3899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_create_subintent.rs
```diff
@@ -0,0 +1,250 @@
...
+            network_id,
+        );
+        let current_epoch = gateway_client.current_epoch().await?;
+
+        // Calculate the seconds until the expiration of the subintent.
+        let expiry_time_from_now_in_seconds =
+            self.expiry_time_from_now_in_seconds(expiration);
+        if expiry_time_from_now_in_seconds == 0 {
+            return Err(CommonError::SubintentExpired);
+        }
```
GhenadieVP, line 27: This could be moved up, so we avoid making a gateway call is the subintent is expired already

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_create_subintent.rs
```diff
@@ -0,0 +1,250 @@
...
+            network_id,
+            start_epoch_inclusive: current_epoch,
+            end_epoch_exclusive: end_ranges.0,
+            min_proposer_timestamp_inclusive: None,
+            max_proposer_timestamp_exclusive: Some(end_ranges.1),
+            intent_discriminator,
+        };
+
+        // Build subintent
+        Subintent::new(header, subintent_manifest, MessageV2::None)
```
GhenadieVP, line 46: Message should not be none, it should be set to whatever message was received with the WalletInteraction

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_create_subintent.rs
```diff
@@ -0,0 +1,250 @@
...
+        const MAX_EPOCH_DIFF: u64 =
+            30 * 24 * 60 * 60 / EXPECTED_EPOCH_DURATION_IN_SECONDS;
+
+        // 1 epoch for the fact that it's min_inclusive and max_exclusive; 1 more for the fact that we might be very close to the end of the epoch already
+        const MIN_EPOCH_DIFF: u64 = 2;
+
+        let epoch_diff = MAX_EPOCH_DIFF.min(
+            MIN_EPOCH_DIFF
+                + (expiry_time_from_now_in_seconds
+                    / EXPECTED_EPOCH_DURATION_IN_SECONDS),
```
GhenadieVP, line 86: This is good, maybe we should use MIN_EPOCH_DIFF only if:
```rust
(expiry_time_from_now_in_seconds / EXPECTED_EPOCH_DURATION_IN_SECONDS) < MIN_EPOCH_DIFF
```
?

matiasbzurovski: I thought about this as well, but I concluded we want to always add 2 more epochs than the epochs until expiration. 

Let's imagine we are in `Epoch(1)` and we have an expiration of 14 minutes. If we apply the suggested change, the result would be `Epoch(3)`. Imagine we are 2 minutes away from next epoch, that means that the user will have the following time period to sign the subintent:
```
Epoch 1: 2 minutes
Epoch 2: 5 minutes
Epoch 3: 5 minutes
```

This is, a total of 12 minutes, while the expiration is in 14 minutes. That's why I believe @dhedey always adds the 2 extra epochs, which he justifies with the constant comment:
```Rust 
// 1 epoch for the fact that it's min_inclusive and max_exclusive; 
// 1 more for the fact that we might be very close to the end of the epoch already
```

GhenadieVP (approved):

crates/sargon/src/system/sargon_os/pre_authorization/sargon_os_create_subintent.rs
```diff
@@ -0,0 +1,250 @@
...
+        &self,
+        current_epoch: Epoch,
+        expiry_time_from_now_in_seconds: u64,
+    ) -> (Epoch, Instant) {
+        // 5 minutes
+        const EXPECTED_EPOCH_DURATION_IN_SECONDS: u64 = 300;
+
+        // As per the transaction validation configuration, epoch diff should be less than 1 month.
+        const MAX_EPOCH_DIFF: u64 =
+            30 * 24 * 60 * 60 / EXPECTED_EPOCH_DURATION_IN_SECONDS;
```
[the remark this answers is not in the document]
dhedey: (For some background, it's not a guaranteed length if e.g. the network stalls or goes _very_ slowly, but in practice it's very accurate)

PR #256, opened 2024-11-07
Use V1 analysis for enclosed manifests

GhenadieVP:
This is a temporary trick so we have properly working Enclosed Pre-Auth preview instead of a stubbed execution summary. We should replace with using the proper endpoint for v2 analysis when it is in place.


crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -55,21 +56,14 @@ impl SargonOS {
         }
 
         // Get the transaction preview
-        let transaction_preview = self
-            .get_transaction_preview(
+        let execution_summary = self
```
matiasbzurovski, line 59: Update comment on line above

matiasbzurovski (approved):

PR #265, opened 2024-11-15
Delete account classification

GhenadieVP:
Introduce DeleteAccount clasification if there are account badges deposited into the account itself.

### DIFF
--- a/apple/Sources/Sargon/Drivers/Networking/NetworkingDriver+URLSession.swift
+++ b/apple/Sources/Sargon/Drivers/Networking/NetworkingDriver+URLSession.swift
@@ -25,6 +25,7 @@ extension URLSession: NetworkingDriver {
 			throw SargonError.NetworkRequestGenericFailure(
 				underlying: String(describing: error))
 		}
+        print(String.init(data: response.0, encoding: .utf8));
 		return try NetworkResponse(response: response)
 	}
 }
--- a/apple/Sources/Sargon/Drivers/Networking/NetworkingDriver+URLSession.swift
+++ b/apple/Sources/Sargon/Drivers/Networking/NetworkingDriver+URLSession.swift
@@ -25,7 +25,6 @@ extension URLSession: NetworkingDriver {
 			throw SargonError.NetworkRequestGenericFailure(
 				underlying: String(describing: error))
 		}
-        print(String.init(data: response.0, encoding: .utf8));
 		return try NetworkResponse(response: response)
 	}
 }
--- a/crates/sargon-uniffi/src/wrapped_radix_engine_toolkit/low_level/transaction_classes/detailed_manifest_class.rs
+++ b/crates/sargon-uniffi/src/wrapped_radix_engine_toolkit/low_level/transaction_classes/detailed_manifest_class.rs
@@ -94,3 +94,5 @@ pub enum DetailedManifestClass {
         account_addresses: Vec<AccountAddress>,
     },
 }
+
+delegate_debug_into!(DetailedManifestClass, InternalDetailedManifestClass);
--- a/crates/sargon-uniffi/src/wrapped_radix_engine_toolkit/low_level/transaction_classes/detailed_manifest_class.rs
+++ b/crates/sargon-uniffi/src/wrapped_radix_engine_toolkit/low_level/transaction_classes/detailed_manifest_class.rs
@@ -95,4 +95,7 @@ pub enum DetailedManifestClass {
     },
 }
 
-delegate_debug_into!(DetailedManifestClass, InternalDetailedManifestClass);
+delegate_display_debug_into!(
+    DetailedManifestClass,
+    InternalDetailedManifestClass
+);
--- a/crates/sargon-uniffi/src/wrapped_radix_engine_toolkit/low_level/transaction_classes/detailed_manifest_class.rs
+++ b/crates/sargon-uniffi/src/wrapped_radix_engine_toolkit/low_level/transaction_classes/detailed_manifest_class.rs
@@ -87,4 +87,10 @@ pub enum DetailedManifestClass {
         /// The redemptions observed in the transaction
         pool_redemptions: Vec<TrackedPoolRedemption>,
     },
+
+    /// A manifest that deletes accounts.
+    DeleteAccounts {
+        /// The addresses of the accounts that are being deleted
+        account_addresses: Vec<AccountAddress>,
+    },
 }
--- a/crates/sargon/src/core/error/common_error.rs
+++ b/crates/sargon/src/core/error/common_error.rs
@@ -735,7 +735,7 @@ pub enum CommonError {
     #[error("Unable to make {amount} transfers in one single transaction")]
     MaxTransfersPerTransactionReached { amount: u64 } = 10208,
 
-    #[error("Transaction Manifest class is forbidden: {}", class.name())]
+    #[error("Transaction Manifest class is forbidden: {class}")]
     ForbiddenManifestClass { class: DetailedManifestClass } = 10209,
 }
 
--- a/crates/sargon/src/core/error/common_error.rs
+++ b/crates/sargon/src/core/error/common_error.rs
@@ -735,8 +735,8 @@ pub enum CommonError {
     #[error("Unable to make {amount} transfers in one single transaction")]
     MaxTransfersPerTransactionReached { amount: u64 } = 10208,
 
-    #[error("Transaction Manifest class is forbidden: {class}")]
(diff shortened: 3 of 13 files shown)


apple/Sources/Sargon/Drivers/Networking/NetworkingDriver+URLSession.swift
```diff
@@ -25,6 +25,7 @@ extension URLSession: NetworkingDriver {
 			throw SargonError.NetworkRequestGenericFailure(
 				underlying: String(describing: error))
 		}
+        print(String.init(data: response.0, encoding: .utf8));
```
CyonAlexRDX: prefer `log.debug`

crates/sargon/src/profile/v100/address/non_fungible_local_id.rs
```diff
@@ -43,6 +43,33 @@ impl NonFungibleLocalId {
         NonEmptyMax64Bytes::try_from(bytes.as_ref())
             .map(|value| Self::Bytes { value })
     }
+
+    pub fn as_bytes(&self) -> Option<NonEmptyMax64Bytes> {
+        if let NonFungibleLocalId::Bytes { value } = self {
```
CyonAlexRDX, line 48: prefer `EnumAsInner` on `NonFungibleLocalId` then this method should have been generated for us

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/execution_summary/execution_summary.rs
```diff
@@ -110,6 +110,37 @@ impl ExecutionSummary {
     }
 }
 
+impl ExecutionSummary {
+    pub fn classify_delete_accounts_if_present(&mut self) {
+        if self.detailed_classification.is_empty() {
```
CyonAlexRDX, line 115: prefer return early to avoid "pyramid of doom" (indentation)

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/execution_summary/execution_summary.rs
```diff
@@ -110,6 +110,37 @@ impl ExecutionSummary {
...
+                .filter_map(|deposit| {
+                    let (account_address, resources) = deposit;
+
+                    resources
+                        .iter()
+                        .filter_map(|resource| {
+                            resource.get_non_fungible_indicator()
+                        })
+                        .flat_map(|indicator| indicator.ids())
+                        .any(|id| id.derives_account_address(&account_address))
```
CyonAlexRDX, line 128: a comment here in code would be helpful

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/execution_summary/execution_summary.rs
```diff
@@ -110,6 +110,37 @@ impl ExecutionSummary {
     }
 }
 
+impl ExecutionSummary {
+    pub fn classify_delete_accounts_if_present(&mut self) {
```
CyonAlexRDX, line 114: this method is easier to test if you make one pure fuction, and then another which is `&mut self` which uses the pure func.

crates/sargon-uniffi/src/wrapped_radix_engine_toolkit/low_level/transaction_classes/detailed_manifest_class.rs
```diff
@@ -87,4 +87,10 @@ pub enum DetailedManifestClass {
         /// The redemptions observed in the transaction
         pool_redemptions: Vec<TrackedPoolRedemption>,
     },
+
+    /// A manifest that deletes accounts.
+    DeleteAccounts {
+        /// The addresses of the accounts that are being deleted
+        account_addresses: Vec<AccountAddress>,
+    },
```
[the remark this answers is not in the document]
sergiupuhalschi-rdx: The UI doesn't allow multiple accounts deletion, so I think it's fine to leave the preview type as it is.

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/execution_summary/execution_summary.rs
matiasbzurovski: why are we only checking when there is no other detailed classification btw?

GhenadieVP: we don't want to override the RET classification

apple/Sources/Sargon/Drivers/Networking/NetworkingDriver+URLSession.swift
GhenadieVP: is removed.

crates/sargon/src/profile/v100/address/non_fungible_local_id.rs
GhenadieVP: temporary, will use Michael's code once once pushed

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/execution_summary/execution_summary.rs
GhenadieVP: This will be tested through manifest analysis, not meant to be tested in isolation.

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/execution_summary/execution_summary.rs
matiasbzurovski: but we are pushing, not overriding. Won't the manifest be classified with RET before doing this manual check ourselves?

CyonAlexRDX: Hosts use `.first` on classifcation. I.e. Hosts does not care about multiple classifcations, always reads the first.

CyonAlexRDX: https://github.com/radixdlt/babylon-wallet-ios/blob/ef47ccba90a8ef35eacd48a505212c49c95580f3/RadixWallet/Core/SargonExtensions/ExecutionSummary%2BExtensions.swift#L5

crates/sargon/src/core/error/common_error.rs
```diff
@@ -734,6 +734,9 @@ pub enum CommonError {
 
     #[error("Unable to make {amount} transfers in one single transaction")]
     MaxTransfersPerTransactionReached { amount: u64 } = 10208,
+
+    #[error("Transaction Manifest class is forbidden: {}", class.name())]
+    ForbiddenManifestClass { class: DetailedManifestClass } = 10209,
```
GhenadieVP, line 739: Some derived manifest classes can only happen as result of the interaction coming from the Wallet.

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/execution_summary/execution_summary.rs
GhenadieVP: @matiasbzurovski I meant that if the manifest was classified by RET, avoid doing custom classification.

danvleju-rdx (approved):

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -218,7 +221,23 @@ impl SargonOS {
...
+            .iter()
+            .find(|classification| classification.is_forbidden());
+
+        if let Some(forbidden_manifest_class) = forbidden_manifest_class
+            && !are_instructions_originating_from_host
+        {
+            return Err(CommonError::ForbiddenManifestClass {
+                class: forbidden_manifest_class.clone(),
+            });
+        }
```
GhenadieVP, line 238: Note:
Currently, for delete account this is paranoia check. There are no tests for it, as it cannot happen for the following reasons:
- AccountSecurify is a reserved instruction, so it our current delete manifest can be submitted only by the Wallet, and is pre-validated when we do check for reserved instructions.
- Then, the other way when a manifest can be classified as Delete Accounts, is doing a transfer of the badge, which is not possible - the access controller does not allow withdrawing a non-fungible and transferring it, and we cannot withdraw the badge from a deleted account because it is tombstoned.

sergiupuhalschi-rdx (approved):

CyonAlexRDX (approved):
LGTM

micbakos-rdx (approved):

PR #266, opened 2024-11-15
Tombstone accounts functionality and sync

micbakos-rdx:
This PR allows the host to 
1. ask sargon os to tombstone an account
2. ask sargon os to sync account visibility with the status on ledger

* New flag added to `EntityFlag`
* Function that returns the accounts on current network is updated to also ignore tombstoned accounts.
* Helper methods to mark one or many accounts as tombstoned
* Sync functionality:
   1. For an account address we query its vault address on ledger that holds the `ACCOUNT_OWNER_BADGE`.
   2. If an account holds that resource badge, we need to fetch the NFTs in that vault
   3. For those NFTs if we find at least one of them referencing the account, we decide that this account is "deleted" on ledger. The badge NFT id is a byte type and the bytes derive the account address. (That way we don't need to fetch the data for that NFT)
* Since the vault request does not accept an array of account addresses, when we check against the profile we need to run the same request for all accounts. That is why join is used. 

# To be implemented:
Test the sync method. In order to do that I need to change the `MockNetworkingDriver`. It is currently implemented to expect sequential requests.

> [!note]
> The  base branch is `ABW-3926-delete-account`

### DIFF
--- a/crates/sargon/src/gateway_api/endpoints/state_endpoints.rs
+++ b/crates/sargon/src/gateway_api/endpoints/state_endpoints.rs
@@ -48,6 +48,27 @@ impl GatewayClient {
         self.post("state/entity/page/non-fungibles/", request, res_id)
             .await
     }
+
+    pub(crate) async fn state_entity_page_non_fungible_vaults(
+        &self,
+        request: StateEntityNonFungibleResourceVaultsPageRequest,
+    ) -> Result<
+        PageResponse<
+            NonFungibleResourcesCollectionItemVaultAggregatedVaultItem,
+        >,
+    > {
+        //Result<StateEntityNonFungibleResourceVaultsPageResponse> { TODO ask
+        self.post("state/entity/page/non-fungible-vaults/", request, res_id)
+            .await
+    }
+
+    pub(crate) async fn state_entity_page_non_fungible_vaults_ids(
+        &self,
+        request: StateEntityNonFungibleIdsPageRequest,
+    ) -> Result<PageResponse<NonFungibleLocalId>> {
+        self.post("state/entity/page/non-fungible-vault/ids", request, res_id)
+            .await
+    }
 }
 
 impl GatewayClient {
--- a/crates/sargon/src/gateway_api/methods/state_methods.rs
+++ b/crates/sargon/src/gateway_api/methods/state_methods.rs
@@ -54,6 +54,67 @@ impl GatewayClient {
             .await
             .map(|x| x.unwrap_or(Decimal192::zero()))
     }
+
+    /// Looks up on ledger whether this `account_address` is deleted, by looking up the NFTs
+    /// it owns and checking if its owner badge is one of them.
+    pub async fn check_account_is_deleted(
+        &self,
+        account_address: AccountAddress,
+    ) -> Result<(AccountAddress, bool)> {
+        // Construct the owner badge resource address
+        let owner_badge_resource_address = ResourceAddress::new(
+            SCRYPTO_ACCOUNT_OWNER_BADGE,
+            account_address.network_id(),
+        )?;
+
+        // Get the vaults that hold the owner badge in this account_address
+        let vaults = self
+            .state_entity_page_non_fungible_vaults(
+                StateEntityNonFungibleResourceVaultsPageRequest::new(
+                    Address::Account(account_address),
+                    owner_badge_resource_address,
+                    None,
+                    None,
+                    None,
+                    None,
+                ),
+            )
+            .await?;
+
+        if let Some(vault) = vaults.items.first() {
+            let ledger_state_selector =
+                Some(LedgerStateSelector::from(vaults.ledger_state));
+
+            // When such vault exists, load all NF IDs
+            let ids = self
+                .load_all_pages(
+                    None,
+                    ledger_state_selector.clone(),
+                    |cursor, ledger_state_selector| {
+                        self.state_entity_page_non_fungible_vaults_ids(
+                            StateEntityNonFungibleIdsPageRequest::new(
+                                Address::from(account_address),
+                                vault.vault_address,
+                                owner_badge_resource_address,
+                                ledger_state_selector,
+                                cursor,
+                                None,
+                            ),
+                        )
+                    },
+                )
+                .await?;
+
+            // Check if any id derives the account address.
+            // If that is the case, then this account means that it was swallowed its key
+            let swallows_badge = ids
+                .iter()
+                .any(|id| id.derives_account_address(account_address));
+            Ok((account_address, swallows_badge))
+        } else {
+            Ok((account_address, false))
+        }
+    }
 }
 
 impl GatewayClient {
--- a/crates/sargon/src/gateway_api/models/types/request/state/entity/page/non_fungible_vault_ids.rs
+++ b/crates/sargon/src/gateway_api/models/types/request/state/entity/page/non_fungible_vault_ids.rs
@@ -1,0 +1,47 @@
+use crate::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct StateEntityNonFungibleIdsPageRequest {
+    /// Bech32m-encoded human readable version of the address.
+    pub(crate) address: Address,
+
+    /// Bech32m-encoded human readable version of the address.
+    pub(crate) vault_address: VaultAddress,
+
+    /// Bech32m-encoded human readable version of the address.
+    pub(crate) resource_address: ResourceAddress,
+
+    /// This allows for a request to be made against a historic state. If a constraint is specified,
+    /// the Gateway will resolve the request against the ledger state at that time.
+    /// If not specified, requests will be made with respect to the top of the committed ledger.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) at_ledger_state: Option<LedgerStateSelector>,
+
+    /// This cursor allows forward pagination, by providing the cursor from the previous request.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) cursor: Option<String>,
+
+    /// The page size requested.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) limit_per_page: Option<u64>,
+}
+
+impl StateEntityNonFungibleIdsPageRequest {
+    pub fn new(
+        address: Address,
+        vault_address: VaultAddress,
+        resource_address: ResourceAddress,
+        at_ledger_state: impl Into<Option<LedgerStateSelector>>,
+        cursor: impl Into<Option<String>>,
+        limit_per_page: impl Into<Option<u64>>,
+    ) -> StateEntityNonFungibleIdsPageRequest {
+        StateEntityNonFungibleIdsPageRequest {
+            address,
+            vault_address,
+            resource_address,
+            at_ledger_state: at_ledger_state.into(),
+            cursor: cursor.into(),
+            limit_per_page: limit_per_page.into(),
+        }
+    }
+}
--- a/crates/sargon/src/gateway_api/models/types/request/state/entity/page/non_fungible_vaults.rs
+++ b/crates/sargon/src/gateway_api/models/types/request/state/entity/page/non_fungible_vaults.rs
@@ -1,0 +1,63 @@
+use crate::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct StateEntityNonFungibleResourceVaultsPageRequest {
+    /// Bech32m-encoded human readable version of the address.
+    pub(crate) address: Address,
+
+    /// Bech32m-encoded human readable version of the address.
+    pub(crate) resource_address: ResourceAddress,
+
+    /// This allows for a request to be made against a historic state. If a constraint is specified,
+    /// the Gateway will resolve the request against the ledger state at that time.
+    /// If not specified, requests will be made with respect to the top of the committed ledger.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) at_ledger_state: Option<LedgerStateSelector>,
+
+    /// This cursor allows forward pagination, by providing the cursor from the previous request.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) cursor: Option<String>,
+
+    /// The page size requested.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) limit_per_page: Option<u64>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) opt_ins: Option<StateEntityNonFungibleResourceVaultsPageOptIns>,
+}
+
+impl StateEntityNonFungibleResourceVaultsPageRequest {
+    pub fn new(
+        address: Address,
+        resource_address: ResourceAddress,
+        at_ledger_state: impl Into<Option<LedgerStateSelector>>,
+        cursor: impl Into<Option<String>>,
+        limit_per_page: impl Into<Option<u64>>,
+        opt_ins: impl Into<Option<StateEntityNonFungibleResourceVaultsPageOptIns>>,
+    ) -> StateEntityNonFungibleResourceVaultsPageRequest {
+        StateEntityNonFungibleResourceVaultsPageRequest {
+            address,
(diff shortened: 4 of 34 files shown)


crates/sargon/src/system/sargon_os/sargon_os_sync_accounts.rs
```diff
@@ -0,0 +1,52 @@
...
+            self.clients.http_client.driver.clone(),
+            network_id,
+        );
+
+        let active_account_addresses =
+            accounts.iter().map(|a| a.address).collect::<Vec<_>>();
+        let on_ledger_checks = active_account_addresses
+            .iter()
+            .map(|address| {
+                gateway_client.check_account_is_deleted(address.clone())
```
CyonAlexRDX, line 25: this does `N` many requests. Dont we wanna do **one single** batch request instead?

CyonAlexRDX: I really think we do, imagine 100 accounts...

micbakos-rdx: How? Gateway does not allow this...

crates/sargon/src/system/sargon_os/sargon_os_sync_accounts.rs
```diff
@@ -0,0 +1,52 @@
...
+                .filter_map(|(account_address, is_deleted)| {
+                    if *is_deleted {
+                        Some(account_address.clone())
+                    } else {
+                        None
+                    }
+                })
+                .collect_vec();
+
+        let any_account_tombstoned = !account_addresses_to_tombstone.is_empty();
```
CyonAlexRDX, line 43: I prefer `is_any_account_tombstoned `

crates/sargon/src/profile/logic/account/account_visibility.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl Account {
+    /// Marks the account as hidden
+    pub fn mark_as_hidden(&mut self) {
+        self.flags.insert_flag(EntityFlag::DeletedByUser);
+    }
+
+    /// Marks the account as tombstoned
+    pub fn mark_as_tombstoned(&mut self) {
+        self.flags.insert_flag(EntityFlag::TombstonedByUser);
```
CyonAlexRDX, line 11: we should probably unhide it before marking it tombstoned. I mean we should avoid `[.deleted, .tombstoned]` we should have only `[.tombstoned]`

crates/sargon/src/gateway_api/methods/state_methods.rs
```diff
@@ -54,6 +54,67 @@ impl GatewayClient {
...
+                    owner_badge_resource_address,
+                    None,
+                    None,
+                    None,
+                    None,
+                ),
+            )
+            .await?;
+
+        if let Some(vault) = vaults.items.first() {
```
CyonAlexRDX, line 84: prefer early return with `let Some(vault) = vaults.items.first() else {  Ok((account_address, false)) } ;`

crates/sargon/src/profile/logic/account/account_visibility.rs
micbakos-rdx: Wanted to ask about this, thanks for reminding me.

crates/sargon/src/gateway_api/methods/state_methods.rs
```diff
@@ -54,6 +54,67 @@ impl GatewayClient {
...
+                    ledger_state_selector.clone(),
+                    |cursor, ledger_state_selector| {
+                        self.state_entity_page_non_fungible_vaults_ids(
+                            StateEntityNonFungibleIdsPageRequest::new(
+                                Address::from(account_address),
+                                vault.vault_address,
+                                owner_badge_resource_address,
+                                ledger_state_selector,
+                                cursor,
+                                None,
```
CyonAlexRDX, line 101: `None` == page size ? @matiasbzurovski declared `GATEWAY_PAGE_REQUEST_LIMIT`, should that not be used here?

crates/sargon/src/gateway_api/models/types/request/state/entity/page/non_fungible_vaults.rs
```diff
@@ -0,0 +1,63 @@
...
+
+impl StateEntityNonFungibleResourceVaultsPageRequest {
+    pub fn new(
+        address: Address,
+        resource_address: ResourceAddress,
+        at_ledger_state: impl Into<Option<LedgerStateSelector>>,
+        cursor: impl Into<Option<String>>,
+        limit_per_page: impl Into<Option<u64>>,
+        opt_ins: impl Into<Option<StateEntityNonFungibleResourceVaultsPageOptIns>>,
+    ) -> StateEntityNonFungibleResourceVaultsPageRequest {
```
CyonAlexRDX, line 37: Prefer `Self`

crates/sargon/src/gateway_api/models/types/request/state/entity/page/non_fungible_vaults.rs
```diff
@@ -0,0 +1,63 @@
...
+        StateEntityNonFungibleResourceVaultsPageRequest {
```
CyonAlexRDX, line 38: Prefer `Self`

crates/sargon/src/gateway_api/methods/state_methods.rs
micbakos-rdx: We only request one item, do we need it?

crates/sargon/src/gateway_api/models/types/request/state/entity/page/non_fungible_vault_ids.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl StateEntityNonFungibleIdsPageRequest {
+    pub fn new(
+        address: Address,
+        vault_address: VaultAddress,
+        resource_address: ResourceAddress,
+        at_ledger_state: impl Into<Option<LedgerStateSelector>>,
+        cursor: impl Into<Option<String>>,
+        limit_per_page: impl Into<Option<u64>>,
+    ) -> StateEntityNonFungibleIdsPageRequest {
```
CyonAlexRDX, line 37: Prefer `Self` x2

crates/sargon/src/gateway_api/methods/state_methods.rs
```diff
@@ -54,6 +54,67 @@ impl GatewayClient {
             .await
             .map(|x| x.unwrap_or(Decimal192::zero()))
     }
+
+    /// Looks up on ledger whether this `account_address` is deleted, by looking up the NFTs
+    /// it owns and checking if its owner badge is one of them.
+    pub async fn check_account_is_deleted(
+        &self,
+        account_address: AccountAddress,
```
CyonAlexRDX, line 62: is it possible to use batch of many account addresses?

crates/sargon/src/system/sargon_os/sargon_os_accounts.rs
```diff
@@ -292,6 +292,76 @@ impl SargonOS {
...
+            for updated_account in updated.clone() {
+                if p.update_account(&updated_account.address, |old| {
+                    *old = updated_account.clone()
+                })
+                .is_none()
+                {
+                    result = Err(CommonError::UnknownAccount);
+                    break;
+                }
+            }
```
matiasbzurovski, line 314: if an array with 5 account has only the third one as missing, this would update the first 2 and return an error. Wouldn't that leave the Profile in a corrupted state?

crates/sargon/src/profile/v100/entity/account/account.rs
```diff
@@ -172,40 +173,54 @@ impl Account {
...
     /// A HIDDEN `Mainnet` account named "Diana", a sample used to facilitate unit tests, with
     /// derivation index 3.
     pub fn sample_mainnet_diana() -> Self {
-        Self::sample_at_index_name(3, "Diana", true)
+        Self::sample_at_index_name(3, "Diana", true, false)
+    }
+
+    /// A tombestoned account named Sean, with derivation index 4.
+    pub fn sample_mainnet_sean() -> Self {
+        Self::sample_at_index_name(4, "Sean", false, true)
```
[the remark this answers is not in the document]
CyonAlexRDX: `erin` please :)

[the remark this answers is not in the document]
CyonAlexRDX: https://en.wikipedia.org/wiki/Alice_and_Bob

crates/sargon/src/profile/v100/entity/entity_flag.rs
```diff
@@ -19,9 +19,13 @@ use crate::prelude::*;
 )]
 #[serde(rename_all = "camelCase")]
 pub enum EntityFlag {
-    /// The entity is marked as deleted by user. Entity should still be kept in Profile
+    /// The entity is marked as hidden by user. Entity should still be kept in Profile
+    /// The user can "unhide" the entity and continue involving it in transactions on ledger.
     DeletedByUser,
```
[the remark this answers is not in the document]
CyonAlexRDX: agreed

crates/sargon/src/system/sargon_os/sargon_os_sync_accounts.rs
micbakos-rdx: Not possible with the current gateway endpoint. We could maybe batch those requests ourselves.

CyonAlexRDX: awaiting https://rdxworks.slack.com/archives/C06EBEA0SGY/p1731673177430949?thread_ts=1731668714.440949&cid=C06EBEA0SGY

crates/sargon/src/system/sargon_os/sargon_os_accounts.rs
CyonAlexRDX: hmm better than nothing I think? The more the merrier so to speak. Better mark `N` of `M` tombstoned accounts as such, even if `N < M`.

CyonAlexRDX: hmm better than nothing I think? The more the merrier so to speak. Better mark `N` of `M` tombstoned accounts as such, even if `N < M`.

crates/sargon/src/gateway_api/endpoints/state_endpoints.rs
```diff
@@ -48,6 +48,26 @@ impl GatewayClient {
         self.post("state/entity/page/non-fungibles/", request, res_id)
             .await
     }
+
+    pub(crate) async fn state_entity_page_non_fungible_vaults(
+        &self,
+        request: StateEntityNonFungibleResourceVaultsPageRequest,
```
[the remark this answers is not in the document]
micbakos-rdx: I started doing that and I saw the rest of the methods were not following this rule, that is why I didn't follow it either.

To add to that most of the responses are `PageResponse<T>` so the name of the response cannot be the same as the endpoint function's signature.

crates/sargon/src/gateway_api/models/types/response/state/entity/page/non_fungible_vaults.rs
```diff
@@ -0,0 +1,21 @@
+use crate::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct StateEntityNonFungibleResourceVaultsPageResponse {
```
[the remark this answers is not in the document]
micbakos-rdx: Actually not needed. Since we use `PageResponse`.

crates/sargon/src/profile/v100/entity/account/account.rs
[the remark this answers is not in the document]
micbakos-rdx: Wanted to name a character that always dies first in films. That's why I chose Sean :P

crates/sargon/src/profile/logic/profile_network/profile_network_get_entities.rs
```diff
@@ -6,7 +6,7 @@ impl ProfileNetwork {
     }
 
     pub fn accounts_non_hidden(&self) -> Accounts {
-        self.accounts.non_hidden()
+        self.accounts.visible()
```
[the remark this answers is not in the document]
micbakos-rdx: We can expose such method, but currently hosts don't use it.

crates/sargon/src/profile/v100/entity/account/account.rs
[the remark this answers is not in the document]
CyonAlexRDX: haha, or `Kenny` "Oh my god, the killed Kenny!". But lets go with `erin`, we use it elsewhere for the index 4.

crates/sargon/src/system/sargon_os/sargon_os_accounts.rs
matiasbzurovski: for this case yeah, but as a general rule (this function is to update accounts, not only to tombstone them) not sure if it is better partially updating.

crates/sargon/src/profile/logic/profile_network/profile_network_get_entities.rs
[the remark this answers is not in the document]
matiasbzurovski: actually it won't be a problem because we unhide it before tombstoning it 👍

crates/sargon/src/gateway_api/endpoints/state_endpoints.rs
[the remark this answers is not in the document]
micbakos-rdx: Exactly just wanted to note that most of the responses are PageResponse<T> and we usually name the T on something related to the item in the page. We don't call it as *Response.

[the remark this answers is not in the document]
matiasbzurovski: yes, but we can still order the `XYZRequest` so that they match the endpoint path

crates/sargon/src/system/sargon_os/sargon_os_accounts.rs
GhenadieVP: I would opt for having a consistency check done before calling update_profile. 
Additionally what would be a scenario when we can end up calling `update_accounts` with a missing account? If there is a scenario, should we throw an error, or doing an update on a non-existing account can be considered noop?

GhenadieVP (approved):
Amazing work 💯!

crates/sargon/src/profile/v100/address/non_fungible_local_id.rs
```diff
@@ -53,6 +64,25 @@ impl NonFungibleLocalId {
...
+        account_address: AccountAddress,
+    ) -> bool {
+        self.as_bytes()
+            .map(|local_id_bytes| {
+                let bytes = local_id_bytes.bag_of_bytes.bytes();
+                if bytes.len() == ScryptoNodeId::LENGTH {
+                    bytes[..ScryptoNodeId::LENGTH]
+                        == account_address.node_id().0
+                } else {
+                    false
```
Sajjon, line 79: missing test?

crates/sargon/src/lib.rs
```diff
@@ -51,6 +52,7 @@ pub mod prelude {
     pub(crate) use derive_more::derive::{
         AsRef, Debug as MoreDebug, Deref, Display,
     };
+    pub(crate) use futures::future::{join_all, try_join_all};
```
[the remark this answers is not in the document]
micbakos-rdx: Right i will remove the futures crate too

crates/sargon/src/system/sargon_os/sargon_os_accounts.rs
CyonAlexRDX: We SHOULD know if we should use ADD or UPDATE, that is really something we should keep track of. So I think `update` ought to always be failing for unknown entities. It is a programmer error.

CyonAlexRDX: >  if the account is missing.
sounds like @GhenadieVP you thought I meant to fail if parameter `updated: Accounts` missed some of the existing accounts. That is not what I meant. That would be strange!

I'm saying if `updated: Accounts` contains a **new** account, which is not in profile, we ought to fail.

And I'm also saying that I think we should revert the whole operation if an update of any account fails. Which feels more prudent and **easy to reason about**.

CyonAlexRDX (approved):
LGTM

PR #268, opened 2024-11-15
Allow deleting virtual accounts

matiasbzurovski:
## Changelog
The [resource-preferences](https://radix-babylon-gateway-api.redoc.ly/#operation/AccountResourcePreferencesPage) & [authorized-depositors](https://radix-babylon-gateway-api.redoc.ly/#operation/AccountAuthorizedDepositorsPage) endpoints are returning 404 if the address provided is a virtual account (not registered in Ledger as it doesn't have any resources).

This would prevent users from deleting such account, so we are adding this temporary workaround until GW doesn't return 404 anymore for these cases.

### DIFF
--- a/crates/sargon/src/gateway_api/endpoints/state_endpoints.rs
+++ b/crates/sargon/src/gateway_api/endpoints/state_endpoints.rs
@@ -99,19 +99,8 @@ impl GatewayClient {
         &self,
         request: AccountPageResourcePreferencesRequest,
     ) -> Result<PageResponse<AccountResourcePreference>> {
-        // The GW is currently returning a 404 when this endpoint is called with a virtual account.
-        // This is a temporary workaround until the GW is fixed.
-        // More info on thread: https://rdxworks.slack.com/archives/C06EBEA0SGY/p1731686360114749
-        match self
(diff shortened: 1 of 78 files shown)


GhenadieVP (approved):

crates/sargon/src/gateway_api/endpoints/state_endpoints.rs
```diff
@@ -62,8 +62,19 @@ impl GatewayClient {
...
-        self.post("state/account/page/resource-preferences", request, res_id)
+        // The GW is currently returning a 404 when this endpoint is called with a virtual account.
+        // This is a temporary workaround until the GW is fixed.
+        // More info on thread: https://rdxworks.slack.com/archives/C06EBEA0SGY/p1731686360114749
+        match self
+            .post("state/account/page/resource-preferences", request, res_id)
             .await
+        {
+            Ok(response) => Ok(response),
+            Err(CommonError::NetworkResponseBadCode { code: 404 }) => {
```
Sajjon, line 73: I think the Slack thread mentioned `500`, should we really ONLY map for `404`?

sergiupuhalschi-rdx (approved):

PR #269, opened 2024-11-18
Remove references to deleted accounts + emit profile changes

GhenadieVP:
- Emit profile updates when an account is hidden/tombstoned.
- Remove account references from dApps when hidden/tombstoned.

### DIFF
--- a/crates/sargon/src/profile/v100/networks/network/authorized_dapp/authorized_persona_simple.rs
+++ b/crates/sargon/src/profile/v100/networks/network/authorized_dapp/authorized_persona_simple.rs
@@ -1,3 +1,5 @@
+use radix_engine_interface::blueprints::account;
+
 use crate::prelude::*;
 
 /// Simple data representation of a Persona the user has shared with a Dapp.
@@ -34,6 +36,21 @@ pub struct AuthorizedPersonaSimple {
 
     /// ID to PersonaData entries to user has shared with a Dapp.
     pub shared_persona_data: SharedPersonaData,
+}
+
+impl AuthorizedPersonaSimple {
+    /// Removes the referenced account for this dApp
+    pub(crate) fn remove_shared_account(
+        &mut self,
+        account_address: &AccountAddress,
+    ) -> bool {
+        if let Some(shared_accounts) = self.shared_accounts.as_mut() {
+            shared_accounts.remove_entry(&account_address);
+            true
+        } else {
+            false
+        }
+    }
 }
 
 impl AuthorizedPersonaSimple {
--- a/crates/sargon/src/profile/v100/networks/network/authorized_dapp/authorized_persona_simple.rs
+++ b/crates/sargon/src/profile/v100/networks/network/authorized_dapp/authorized_persona_simple.rs
@@ -1,5 +1,3 @@
-use radix_engine_interface::blueprints::account;
-
 use crate::prelude::*;
 
 /// Simple data representation of a Persona the user has shared with a Dapp.
--- a/crates/sargon/src/profile/v100/networks/network/authorized_dapp/authorized_persona_simple.rs
+++ b/crates/sargon/src/profile/v100/networks/network/authorized_dapp/authorized_persona_simple.rs
@@ -37,7 +37,7 @@ pub struct AuthorizedPersonaSimple {
 }
 
 impl AuthorizedPersonaSimple {
-    /// Removes the referenced account for this dApp
+    /// Removes the referenced account from the shared accounts
     pub(crate) fn remove_shared_account(
         &mut self,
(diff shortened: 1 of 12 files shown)


crates/sargon/src/profile/v100/networks/network/authorized_dapp/authorized_persona_simple.rs
```diff
@@ -1,3 +1,5 @@
+use radix_engine_interface::blueprints::account;
```
Sajjon, line 1: probably incorrect import?

crates/sargon/src/system/sargon_os/sargon_os_accounts.rs
```diff
@@ -331,34 +331,56 @@ impl SargonOS {
...
-        self.update_account(account).await
+        self.update_profile_with(|profile| {
+            profile.networks.hide_account(&account_address);
+            Ok(())
+        })
+        .await?;
+
+        self.clients
+            .profile_state_change
+            .emit(ProfileState::Loaded(self.profile()?))
```
Sajjon, line 342: 😞 at least put a `// TODO: This is hacky and temporary and should be removed in the future, replaced by finer granularity Events` or similar?

CyonAlexRDX: Right, but still ought to be a function, with good name; and we document it, why we need it.

micbakos-rdx (approved):

sergiupuhalschi-rdx (approved):

crates/sargon/src/system/sargon_os/sargon_os_accounts.rs
GhenadieVP: Move to `update_profile_with`. You mean to add the reasoning outlined in the ticket -> https://radixdlt.atlassian.net/browse/ABW-3591?

crates/sargon/src/profile/v100/networks/network/profile_network.rs
```diff
@@ -174,6 +174,32 @@ impl ProfileNetwork {
...
+    ) -> Option<Account> {
+        let account = self.update_account(account_address, |account| {
+            account.mark_as_hidden();
+        });
+        self.authorized_dapps
+            .remove_referenced_account(account_address);
+        account
+    }
+
+    /// Tombostones the account associated with the `account_address`
```
matiasbzurovski, line 191:
```suggestion
    /// Tombstones the account associated with the `account_address`
```

applies in other places as well

matiasbzurovski (approved):

crates/sargon/src/system/sargon_os/sargon_os_accounts.rs
Sajjon: Or am I missing something?

GhenadieVP: > the emit call is not needed at all.

Not sure what you mean, the hosts need be notified with the profile is updated internally by Sargon. Up until this PR the hosts where notified only on wallet creation, and when the Hosts called `set_profile` after doing the mutation on their side.

CyonAlexRDX: duh 🤦‍♂️ haha I missed we did not call inside of 

```rust
     match state {
            ProfileState::Loaded(ref mut profile) => mutate(profile),
            _ => Err(CommonError::ProfileStateNotLoaded {
                current_state: state.to_string(),
            }),
        }
```

ok so my point is, instead of "sprinkling" emits, why dont we add a single emit inside of that match? inside of `update_profile_with` in `ProfileStateHolder` ?

crates/sargon/src/profile/v100/networks/network/profile_network.rs
CyonAlexRDX: strange that `typos` step in CI did not catch this.

danvleju-rdx (approved):

crates/sargon/src/system/sargon_os/sargon_os_accounts.rs
GhenadieVP: I mentioned above that I moved it in [sargon_os update_profile_with](https://github.com/radixdlt/sargon/blob/b4fbe078d80867c34f561035001e34e7dce1652a/crates/sargon/src/system/sargon_os/sargon_os_profile.rs#L145).

CyonAlexRDX (approved):
lgtm

giannis-rdx (approved):

micbakos-rdx:
@GhenadieVP Tested on Android host. Seems good to go. Are we planning on merging it?

PR #276, opened 2024-11-21
Use the correct artifact path on the release step

muzuke:


.github/workflows/release.yml
```diff
@@ -105,13 +103,13 @@ jobs:
       - name: Create release
         if: github.ref == 'refs/heads/main'
         run: |
-          # This MUST match whatever you we have declared in Package.swift
```
CyonAlexRDX, line 108: This is an important comment which you ignore and now deleted?

.github/workflows/release.yml
```diff
@@ -105,13 +103,13 @@ jobs:
...
-          SWIFT_SARGON_BINARY_ASSET_NAME="libsargon-rs.xcframework.zip"
+          set -eux
+
+          ls -l target/swift/*.zip
```
CyonAlexRDX, line 108: I think we wanna move the zip from target/Swift to repo root.

Then it will work. With Package.swift unchanged.

CyonAlexRDX: I mean this path:

https://github.com/radixdlt/sargon/blob/main/Package.swift#L27

But maybe that will be the path even though it looks like it will be target/swift?

But I think probably safest to `mv` the zip from `target/swift` to Proj **root** here in GH workflow instead

CyonAlexRDX (approved):
🤞

PR #277, opened 2024-11-21
Remove single quotes from release title variable

muzuke:
The single quotes were treating the variable as literal.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -78,6 +78,9 @@ jobs:
         run: |
           set -euo pipefail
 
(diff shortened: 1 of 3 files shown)


Sajjon:
@muzuke the CD for iOS is broken actually, this `let useLocalFramework = true` never got changed to `let useLocalFramework = false`.

So this `sed` did not work:
https://github.com/radixdlt/sargon/blob/main/.github/workflows/release.yml#L55

wrong working dir??

CyonAlexRDX (changes requested):
remove the last `sed` to fix invalid iOS releases please :)

PR #261, opened 2024-11-13
Update PreAuthorization response model

matiasbzurovski:
## Changelog
Updates `WalletToDappInteractionPreAuthorizationResponseItems` to include one new layer (under `"resposne"` field) when sending back the `signedPartialTransaction`, as defined in [CAP-39](https://radixdlt.atlassian.net/wiki/spaces/AT/pages/3702816769/CAP-39+Pre-authorization+request#Wallet-response).

PR #279, opened 2024-11-27
[ABW-4003] Update WalletToDappInteractionSubintentResponseItem

matiasbzurovski:
Jira Ticket: [ABW-4003](https://radixdlt.atlassian.net/browse/ABW-4003)

## Changelog
Updates model to follow latest definition of [CAP-39](https://radixdlt.atlassian.net/wiki/spaces/AT/pages/3702816769/CAP-39+Pre-authorization+request). 

Instead of retaining the `encoded_signed_partial_transaction`, the model will now hold the `SignedSubintent` object and serialize into the mentioned field plus the new `subintentHash`.

[ABW-4003]: https://radixdlt.atlassian.net/browse/ABW-4003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/sargon/src/radix_connect/wallet_interaction/dapp_wallet_interaction/wallet_to_dapp/success_response/pre_authorization/subintent_response.rs
```diff
@@ -1,35 +1,73 @@
 use crate::prelude::*;
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct WalletToDappInteractionSubintentResponseItem {
-    /// A hex encoded signed partial transaction.
-    #[serde(rename = "signedPartialTransaction")]
-    pub encoded_signed_partial_transaction: String,
+    /// A signed subintent
+    pub signed_subintent: SignedSubintent,
```
matiasbzurovski, line 6: This is he change you had requested in https://github.com/radixdlt/sargon/pull/261 @CyonAlexRDX 

I wasn't able to implement it using `#[serde_as]`, and had to add custom `Serialize`/`Deserialize` implementations. Let me know if you know of an easier way of doing so.

crates/sargon/src/radix_connect/wallet_interaction/dapp_wallet_interaction/wallet_to_dapp/success_response/pre_authorization/subintent_response.rs
```diff
@@ -52,9 +90,15 @@ mod tests {
...
+            &SUT::sample_other(),
+            r#"
+            {
+                "signedPartialTransaction": "4d220e03210221012105210607f20a00000000000000000a0a000000000000002200002200000ab168de3a00000000202000220000202000202200202100202200202000",
+                "subintentHash": "subtxid_sim1kdwxe9mkpgn2n5zplvh4kcu0d69k5qcz679xhxfa8ulcjtjqsvtq799xkn"
+            }
+            "#,
+        );
     }
 }
```
Sajjon, line 104: Lets try to see why you did not get derive serde to work. but if we really cannot then:

please add unit tests for FAILING json decoding e.g. missing key `"subintentHash"` and another missing `"signedPartialTransaction"` and some where those keys exist but wrong value kind. That is really what one ought to do when writing a custom serde.

crates/sargon/src/wrapped_radix_engine_toolkit/low_level/v2/signed_subintent.rs
```diff
@@ -31,6 +31,10 @@ impl SignedSubintent {
         )
         .expect("Compiling after initialization is always valid")
     }
+
+    pub fn decompiled(compiled: Vec<u8>) -> Result<Self> {
```
Sajjon, line 35: hmm do we use this verb tense for ctor? maybe we do in a few places. But I would prefer `decompling` I think

crates/sargon/src/radix_connect/wallet_interaction/dapp_wallet_interaction/wallet_to_dapp/success_response/pre_authorization/subintent_response.rs
Sajjon: is `SignedSubintent` Serde? if not, that is why you could not use derive here. If it is not, why not? is out our type?

CyonAlexRDX: ah sorry, I reviewed a bit hastily. I see now... but still the custom Serde should go on `SignedSubintent` and not `WalletToDappInteractionSubintentResponseItem` (which **can** derive Serde then, since it only consists of serde fields)

crates/sargon/src/radix_connect/wallet_interaction/dapp_wallet_interaction/wallet_to_dapp/success_response/pre_authorization/subintent_response.rs
CyonAlexRDX: keep this simple single `json_roundtrip` test, but put more unit tests of failing serde inside `SignedSubintent` (see https://github.com/radixdlt/sargon/pull/279/files#r1860603707 )

crates/sargon/src/radix_connect/wallet_interaction/dapp_wallet_interaction/wallet_to_dapp/success_response/pre_authorization/subintent_response.rs
matiasbzurovski: if we add the custom Serialize/Deserialize inside `SignedSubintent` (rather than here), then every usage of it will also include the `subintentHash` fields (which isn't really necessary). 
To avoid this, we would need to add custom serialization in both places (and include the `subintentHash` here), which seems a bit unnecessary.

Given we don't need serialization of `SignedSubintent`, I guess it is fine to keep as is, but let me know what you think.

crates/sargon/src/radix_connect/wallet_interaction/dapp_wallet_interaction/wallet_to_dapp/success_response/pre_authorization/subintent_response.rs
matiasbzurovski: actually we don't _need_ the `subintentHash` field to deserialize, but will add tests for the case without `signedPartialTransaction` or where its content is invalid

PR #280, opened 2024-11-27
Lenient `assert_factor_instances_valid`

CyonAlexRDX:
Change so that we only **assert** that factor instances are not shared between entities in these two contexts:
* Securifying entitites
* Adding new entities
And we only assert NEW instances against existing. We do not assert all existing against each other.

Downgrade **assert** to *diagnostics* of instances shared between entities when we set the profile in ProfileHolder.

If we identify a duplicate we log it with `error!` also prepare for possibility of introducing a "diagnostic driver" in the future.

# Future directions
We should make these future improvements:

## DEBUG / PROD aware config
The hosts should boot bios with a `environment_flags` or `runtime_flags` or similar object which tells Sargon if host runs `DEBUG` (Alpha, PreAlpha) or `PROD`. And based on this we might in fact handle some "checks" differently, e.g. we could propagate more errors / panic more etc for `DEBUG`. Typically we could use FULL assertion of factor instances shared between entities - all against all (not only new against existing).

## Diagnostics Driver
We use `diagnostics_for_factor_instances_valid_with_handler` and pass a `on_duplicate` handler which uses a new `DiagnosticsClient` using a new `DiagnosticsDriver` which for Swift uses [Swift Issue Reporting]( https://github.com/pointfreeco/swift-issue-reporting) and `reportIssue` to halt execution when running from Xcode. And for Android host maybe raise an exception or something other cool/useful tool.


crates/sargon/src/system/sargon_os/sargon_os_accounts.rs
```diff
@@ -778,7 +792,7 @@ impl SargonOS {
                 Ok(())
             }
             .and_then(|_| {
-                p.assert_factor_instances_valid().inspect_err(|_| { p.networks = networks_backup; })
+                p.assert_new_factor_instances_not_already_used(instances_of_new_entities.clone()).inspect_err(|_| { p.networks = networks_backup; })
```
CyonAlexRDX, line 795: We used to assert that FactorInstances are not shared between entities against the WHOLE profile, now we assert the NEW against the current only.

crates/sargon/src/system/sargon_os/profile_state_holder.rs
```diff
@@ -143,31 +143,38 @@ impl ProfileStateHolder {
     /// Sets the `ProfileState` held by this `ProfileStateHolder` to the latest `profile_state`.
     pub(super) fn replace_profile_state_with(
         &self,
+        is_android: bool,
         profile_state: ProfileState,
     ) -> Result<()> {
         let mut lock = self.state.write().expect(
             "Stop execution due to the profile state lock being poisoned",
         );
-        Self::assert_factor_instances_valid(&profile_state)?;
```
CyonAlexRDX, line 151: We used to ASSERT on profile replaced, now instead, we "diagnose" which uses `log` to log - and in the future we can add runtime warnings if we pass such a new driver from hosts. (on iOS we can use https://github.com/pointfreeco/swift-issue-reporting to halt execution when running `DEBUG` from Xcode)

crates/sargon/src/factor_instances_provider/provider/factor_instances_provider_unit_tests.rs
```diff
@@ -151,6 +151,12 @@ impl SargonOS {
 
         assert!(security_structures_of_factor_instances.is_empty());
 
+        // Assert that none of the NEW FactorInstances collide with the existing ones
```
CyonAlexRDX, line 154: Context: this is inside:

```rust
#[cfg(test)]
impl Profile {
    async fn __OFFLINE_ONLY_securify_accounts(
        &self,
        account_addresses: IndexSet<AccountAddress>,
        shield: &SecurityStructureOfFactorSources,
    ) -> Result<(Accounts, FactorInstancesProviderOutcome)> {
}
```

crates/sargon/src/profile/v100/profile_legacy_state_bugs.rs
```diff
@@ -0,0 +1,496 @@
...
+            warn!(
+                "Duplicated FactorInstances found {:?} but due to Android bug",
+                duplicate_instances
+            );
+        } else {
+            error!(
+                "Duplicated FactorInstances found {:?}",
+                duplicate_instances
+            );
+        }
```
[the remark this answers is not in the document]
CyonAlexRDX: what do you want the body of:

```rust
 pub(crate) fn diagnostics_for_factor_instances_valid_with_handler(
        &self,
        is_android: bool,
        mut on_duplicate: impl FnMut(DuplicateInstancesWithKnownOs),
    ) {
    }
```

to be then?

crates/sargon/src/profile/v100/profile.rs
```diff
@@ -234,48 +234,167 @@ impl<T: IsEntity> IdentifiedVecOf<T> {
...
+    }
+}
+
+/// Like `DuplicateInstances` but also contains a flag if the current OS is Android.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub(crate) struct DuplicateInstancesWithKnownOs {
+    /// The duplicated instances found
+    pub(crate) duplicate_instances: DuplicateInstances,
+    /// If HostInfo driver returns that the current OS is Android
+    pub(crate) is_android: bool,
```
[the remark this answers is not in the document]
CyonAlexRDX: because should REALLY react to this scenario if we were to detect it on iOS, was my thinking. But I realize this is only relevant for us... so can only happen if we are given an already ""corrupt"" Profile and run it with cable. hmm

crates/sargon/src/profile/v100/profile_legacy_state_bugs.rs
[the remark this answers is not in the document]
GhenadieVP: ok, not explained myself clearly, I meant that we should have just `error!()` instead of check for android and use `warn!` instead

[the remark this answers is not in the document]
CyonAlexRDX: maybe you just wanna do:

```rust
{
    let Some(duplicate_instances) = self.check_for_duplicated_instances() else {
        return;
    };

    error!(
        "Duplicated FactorInstances found {:?}",
        duplicate_instances
    );

    on_duplicate(DuplicateInstancesWithKnownOs {
        duplicate_instances,
        is_android,
    })
}
```

crates/sargon/src/system/sargon_os/profile_state_holder.rs
```diff
@@ -143,31 +143,38 @@ impl ProfileStateHolder {
     /// Sets the `ProfileState` held by this `ProfileStateHolder` to the latest `profile_state`.
     pub(super) fn replace_profile_state_with(
         &self,
+        is_android: bool,
```
matiasbzurovski, line 146: maybe put this param after `profile_state` so that the fn reads `replace_profile_state_with(profile_state,..`)

CyonAlexRDX: I might remove this bool

GhenadieVP (approved):

PR #282, opened 2024-11-28
Determine expiration from SignedSubintent

matiasbzurovski:
Last bit of [ABW-4001](https://radixdlt.atlassian.net/browse/ABW-4003). 

We will use the `max_proposer_timestamp_exclusive` from the `SignedSubintent` to determine when it expires.

### DIFF
--- a/crates/sargon/src/radix_connect/wallet_interaction/dapp_wallet_interaction/wallet_to_dapp/success_response/pre_authorization/subintent_response.rs
+++ b/crates/sargon/src/radix_connect/wallet_interaction/dapp_wallet_interaction/wallet_to_dapp/success_response/pre_authorization/subintent_response.rs
@@ -7,13 +7,13 @@ pub struct WalletToDappInteractionSubintentResponseItem {
     pub signed_subintent: SignedSubintent,
 
     /// The timestamp at which the subintent expires
-    pub expiration_timestamp: Timestamp,
+    pub expiration_timestamp: Instant,
 }
 
 impl WalletToDappInteractionSubintentResponseItem {
     pub fn new(
         signed_subintent: SignedSubintent,
-        expiration_timestamp: Timestamp,
+        expiration_timestamp: Instant,
     ) -> Self {
         Self {
             signed_subintent,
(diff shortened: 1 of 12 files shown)


crates/sargon/src/radix_connect/wallet_interaction/dapp_wallet_interaction/wallet_to_dapp/success_response/pre_authorization/subintent_response.rs
```diff
@@ -29,11 +29,18 @@ impl WalletToDappInteractionSubintentResponseItem {
...
-            .as_seconds_f64() as u64
+impl From<SignedSubintent> for WalletToDappInteractionSubintentResponseItem {
+    fn from(value: SignedSubintent) -> Self {
+        let expiration_timestamp = value
+            .subintent
+            .header
+            .max_proposer_timestamp_exclusive
+            .expect(
+                "A SignedSubintent must have max_proposer_timestamp_exclusive",
+            );
```
matiasbzurovski, line 42: As discussed on meeting, added this force cast to avoid adding defensive programming against our own code. Otherwise, we can also make this a `TryFrom` and handle/ignore the error on host apps.

@GhenadieVP

crates/sargon/src/radix_connect/wallet_interaction/dapp_wallet_interaction/wallet_to_dapp/success_response/pre_authorization/subintent_response.rs
```diff
@@ -116,33 +123,8 @@ mod tests {
 
     #[test]
     fn json_success() {
-        // Test roundtrip for sample (whose timestamp doesn't include milliseconds).
         assert_json_roundtrip(&SUT::sample());
-
-        // Test encoding of sample_other (whose timestamp includes 123 milliseconds).
-        let result =
-            serde_json::to_string_pretty(&SUT::sample_other()).unwrap();
-        let json = r#"
```
GhenadieVP, line 125: I think this test was better, we should assert against an actual representation of the json that would be sent. Doing the json roundtrip can be successful, while the json that is sent to the dApp does not match the RDT expectation.

matiasbzurovski: makes sense, bringing it back

sergiupuhalschi-rdx (approved):

GhenadieVP (approved):

PR #287, opened 2024-12-02
[ABW-3848] Security problems logic

matiasbzurovski:
Jira Ticket: [ABW-3848](https://radixdlt.atlassian.net/browse/ABW-3848)

## Changelog
Adds the `SecurityProblem` enum, and the static logic to determine which problems does the Host wallet have at a given point in time.

## Rationale and next steps
The current solution takes a `CheckSecurityProblemsInput` to determine the problems at a given point in time. Host apps must call the new function to check the problems every time this input changes. This is a first step into a _smarter_ solution where Sargon defines the input itself, and just require Host apps to emit events on a stream every time a condition related to the input changes.

For example, in order to determine entities that are not recoverable, host apps are monitoring their secure/unsecure storage to see which factor sources are present/recoverable. Sargon won't ever have access to such information internally, but it could ask for Host apps to inform it via corresponding streams. Then, the logic to determine the status of the entities would be defined based on the information provided by Host apps + information available on Profile.

The reasoning to do a preliminary static solution:
- When implementing the solution on host apps, there was a lot of iteration over the logic to determine **when** to check security problems. In other words, lots of bugs were resolved before getting to the current stable state. Moving this to Sargon will probably not come for free, and resolving these edge cases again will be even more complicated with the logic  split among host apps/Sargon.
- Takes a lot more time to migrate everything to Sargon, and best case we have the same output than now. 

## Host implementations
The changes in iOS Wallet to use this logic has been added in https://github.com/radixdlt/babylon-wallet-ios/pull/1404. However, this PR is yet marked as draft until Android updates as well and confirms the proposed solution works for both platforms.

[ABW-3848]: https://radixdlt.atlassian.net/browse/ABW-3848?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/sargon-uniffi/src/security_center/support/input.rs
```diff
@@ -0,0 +1,63 @@
+use crate::prelude::*;
+use sargon::CheckSecurityProblemsInput as InternalCheckSecurityProblemsInput;
+use sargon::IsCloudProfileSyncEnabled;
+
+#[derive(Clone, PartialEq, Eq, uniffi::Record)]
+pub struct CheckSecurityProblemsInput {
+    /// Whether the cloud profile sync is enabled.
+    pub is_cloud_profile_sync_enabled: bool,
```
GhenadieVP, line 8: You should be able to use InternalConversion if you add From/Into bool for IsCloudProfileSyncEnabled

GhenadieVP (commented):
Looks very good! hopefully Android can easily accommodate this

crates/sargon-uniffi/src/security_center/support/backup_result.rs
```diff
@@ -0,0 +1,39 @@
...
+#[derive(Clone, PartialEq, Eq, uniffi::Record)]
+pub struct BackupResult {
+    /// The identifier of the backup.
+    pub save_identifier: String,
+
+    /// Whether this backup matches the one on Profile.
+    pub is_current: bool,
+
+    /// Whether this backup has failed.
+    pub is_failed: bool,
```
giannis-rdx, line 18: what do you define as failed in iOS?
If the cloud backup is off, is it a fail?
Is cloud service error a fail? 
Is it a fail, if cloud backup is on but not updated?
All the above?
What else?

just for reference here is how in android we define the [BackupState](https://github.com/radixdlt/babylon-wallet-android/blob/main/core/src/main/java/rdx/works/core/domain/cloudbackup/BackupState.kt) (cloud + manual)

crates/sargon-uniffi/src/security_center/support/backup_result.rs
```diff
@@ -0,0 +1,39 @@
...
+use sargon::IsBackupResultCurrent;
+use sargon::IsBackupResultFailed;
+
+/// A struct that represents the result of a given backup.
+///
+/// Reference for iOS: it is a combination of `BackupStatus` and `BackupResult` (all in one).
+#[derive(Clone, PartialEq, Eq, uniffi::Record)]
+pub struct BackupResult {
+    /// The identifier of the backup.
+    pub save_identifier: String,
```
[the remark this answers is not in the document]
matiasbzurovski: just an identifier of the backup, I've added it in case we wanted to identify among different `BackupResult`, but we can remove since it isn't used.

crates/sargon-uniffi/src/security_center/support/backup_result.rs
matiasbzurovski: In iOS we store in local storage information about the last manual/cloud backup. This information is updated whenever a new backup is started, and when it finishes (successfully or not). The idea is to create this `BackupResult` based on such information.

> what do you define as failed in iOS?

In iOS we have 3 status for a backup: `success`, `failed` and `started`.
We consider it `is_failed` when the `status == failed` or when `status == started` and it has been in such state for 5 minutes or more ([ref](https://github.com/radixdlt/babylon-wallet-ios/blob/main/RadixWallet/Clients/UserDefaults%2BDependency%2BExtension/UserDefaults%2BDependency%2BExtension.swift#L261-L269)).

> Is cloud service error a fail?
> Is it a fail, if cloud backup is on but not updated?
> All the above?
> What else?

Note that is only refers to the result of the last backup attempt. The considerations to whether it is a problem or not based on cloud backups being disabled/enabled is done [here](https://github.com/radixdlt/sargon/pull/287/files#diff-0703cbaefcade0f2714715250fb3956bda33b19756427ef0c4fc563ee88a98c5R22-R28).

In other words, `BackupResult` is agnostic of the Wallet backup configuration. It only reflects the state of the last backup attempt (either manual or cloud). This was easy to do on iOS because as I mentioned, we save such information on local storage. Is Android keeping track of this information as well?

crates/sargon/src/security_center/client.rs
```diff
@@ -0,0 +1,331 @@
...
+
+        let has_problem_5 = || {
+            if !is_cloud_profile_sync_enabled {
+                return false;
+            }
+            let Some(cloud_backup) = input.last_cloud_backup else {
+                return true;
+            };
+            *cloud_backup.is_failed
+        };
```
giannis-rdx, line 29: The logic for problem 5 is not correct. 🔍 

In Android we define as problem 5 when cloud backup has a service error (e.g. google cloud is down),  which is exactly what is described in [confluence](https://radixdlt.atlassian.net/wiki/spaces/AT/pages/3392569357/Security-related+Problem+States+in+the+Wallet#Security-Problem-5). 

this is how we express the cloud related security problems 
```
sealed interface CloudBackupNotWorking : SecurityProblem {
        // security problem 5
        data class ServiceError(val isAnyActivePersonaAffected: Boolean) : CloudBackupNotWorking
        // security problem 6 & 7
        data class Disabled(
            val isAnyActivePersonaAffected: Boolean,
            val hasManualBackup: Boolean
        ) : CloudBackupNotWorking
    }
```

and this is the logic to define when we have a security problem 5:
```
if (cloudBackupState is BackupState.CloudBackupEnabled && cloudBackupState.hasAnyErrors) {
     SecurityProblem.CloudBackupNotWorking.ServiceError(isAnyActivePersonaAffected = activePersonasNeedCloudBackup > 0)
}
```

cloud backup is enabled (meaning toggle in settings screen is on) but we have errors

crates/sargon-uniffi/src/security_center/support/backup_result.rs
giannis-rdx: ok in Android it is different. We have two backup states that holds the information that is needed to derive the problems later.

```data class CloudBackupEnabled(
        override val email: String,
        val hasAnyErrors: Boolean = false,  <--------- if true then security problem 5
        override val lastCloudBackupTime: Timestamp? = null,
        override val lastManualBackupTime: Instant? = null,
        override val lastModifiedProfileTime: Timestamp? = null,
    ) : BackupState()

    data class CloudBackupDisabled(
        override val email: String?,
        override val lastCloudBackupTime: Timestamp?,
        override val lastManualBackupTime: Instant?,
        override val lastModifiedProfileTime: Timestamp?
    ) : BackupState()
```

but based on the way you implemented the consideration whether it is a problem or not I think it should be doable to "map" the `BackupState` to `BackupResult`

giannis-rdx (approved):

giannis-rdx (approved):
🚢  it

PR #292, opened 2024-12-05
Check prerequisites for building a Security Shield

matiasbzurovski:
## Changelog
- Adds `FactorSourceCategory` & `SecurityShieldPrerequisitesStatus` to allow checking if a Security Shield can be built on the current Profile.

> [!Note]
> target branch is `ABW-3848-security-problems` to avoid conflicts.


crates/sargon/src/profile/v100/factors/factor_source_category.rs
```diff
@@ -0,0 +1,17 @@
+use crate::prelude::*;
+
+/// An enum representing the **category** of a `FactorSource`/`FactorSourceKind`.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum FactorSourceCategory {
```
CyonAlexRDX, line 5: FactorSourceKindCategory imo

CyonAlexRDX (changes requested):

crates/sargon/src/system/sargon_os/sargon_os_security_structures.rs
```diff
@@ -100,6 +100,35 @@ impl SargonOS {
...
+    /// Returns the status of the prerequisites for building a Security Shield.
+    ///
+    /// According to [definition][doc], a Security Shield can be built if the user has, asides from
+    /// the Identity factor, "2 or more factors, one of which must be Hardware"
+    ///
+    /// [doc]: https://radixdlt.atlassian.net/wiki/spaces/AT/pages/3758063620/MFA+Rules+for+Factors+and+Security+Shields#Factor-Prerequisites
+    pub fn get_security_shield_prerequisites_status(
+        &self,
+    ) -> Result<SecurityShieldPrerequisitesStatus> {
+        let factor_sources = self.factor_sources()?;
```
Sajjon, line 113: I think we should try to have a thin functions on SargonOs as possible. So i would put this on Profile. And go through ProfileStateHolder

crates/sargon/src/profile/v100/factors/factor_source_category.rs
```diff
@@ -0,0 +1,17 @@
...
+    /// Something I am.
+    Identity,
+
+    /// Something I have.
+    Hardware,
+
+    /// Something I know.
+    Information,
+
+    /// Someone else.
```
Sajjon, line 15: Maybe:
`/// Someone I trust`

crates/sargon/src/profile/v100/factors/factor_source_category.rs
matiasbzurovski: I thought about this but went with `FactorSourceCategory` because `kind` and `category` are synonyms, so it sounded a bit weird. Also, given a `FactorSource` can be categorized as well (in fact, it is how we use it) I think this makes more sense.

CyonAlexRDX (approved):
LGTM, good job!

PR #293, opened 2024-12-06
`TransationManifest::securify_unsecurified_entity ` & `ProvisionalSecurifiedConfig`

CyonAlexRDX:
# Preface
This should have been two PRs, sorry about that. It could have made sense for all the changes to be part of the same PR if I also implemented the [`SecurifyEntitiesManager`](https://radixdlt.atlassian.net/browse/ABW-4016) - which I started doing because I'm bad at not-scope-creeping, but I did not wanna create yet another 6k LOC PR.

So two (somewhat distinct) changes this PR brings is:
1. `TransactionManifest::securify_unsecurified_entity`
2. Adds a `provisional: Option<ProvisionalSecurifiedConfig>` on `UnsecurifiedEntityControl` and `SecurifiedEntityControl`

Additional changes:
* Removed `authentication_signing: Option<HierarchicalDeterministicFactorInstance>` field on `UnsecurifiedEntityControl` 
* Added `authentication_signing_factor_instance: HierarchicalDeterministicFactorInstance ` field to `SecurifiedEntityControl` `SecurityStructureOfFactorInstances`

# `TransactionManifest::securify_unsecurified_entity`
[Resurrected `impl From<Role> for ScryptoAccessRole`](https://github.com/radixdlt/sargon/blob/8d10a02c8fa1497a03ae464148e0a305b89f674f/crates/sargon/src/profile/mfa/security_structures/decl_security_structure_of.rs#L9-L29) (implemented by @0xOmarA ) from before #286 was merged (which temporarily removed this).

# `ProvisionalSecurifiedConfig`
Once a user has selected some accounts and/or personas to securify with a new shield, those accounts should now **retain** the information that user wants to use some shield with them. 

But the `FactorInstancesProvider` might not contain enough factors to securify all those accounts now. And user can always cancel derivation even if we try to derive more factors. **This means we cannot directly save `SecurityStructureOfFactorInstances` per account. **

So we start with saving the unique shield ID - `SecurityStructureID` associated with the accounts (or personas). This is the variant `ShieldSelected(SecurityStructureID)` of the new type `ProvisionalSecurifiedConfig`. 

Then **later** when we were able to use the `FactorInstanceProvider` to provide factor instances for the `SecurityStructureOfFactorSourceIds` - identified by `SecurityStructureID` of `ProvisionalSecurifiedConfig::ShieldSelected` - we update the `provisional` state of all relevant entities in Profile
**from**: `ProvisionalSecurifiedConfig::ShieldSelected`
**to**: `ProvisionalSecurifiedConfig::FactorInstancesDerived(SecurityStructureOfFactorInstances)`.

Later we might ask user to sign a batch of transaction. **User might cancel, so we must support reminding user to batch sign these transaction later**.

Later, once signed we update the `provisional` state of all relevant entities in Profile 
**from**: `ProvisionalSecurifiedConfig::FactorInstancesDerived` 
**to**: `ProvisionalSecurifiedConfig::TransactionQueued(SecurityStructureOfFactorInstances, TXID)`.

I've put assertions on what is destructive / unsupported transitions:
`FactorInstancesDerived(SecurityStructureOfFactorInstances) -> ShieldSelected` is invalid for now, since those `FactorInstance` already have been deleted from the `FactorInstancesCache` so those factor instances would be lost, and we would have "gaps" in the derivation entity index - not the woooorst thing which can happen, but it makes Account Recovery Scan tricker in the future, since we would need to scan a broader range. So we should try to avoid creating such gaps. By "invalid for now" I mean this PR does not support it. But we can probably quite easily support this as a feature, by allowing us to "put instance back into the cache"!

Furthermore transitions from `TransactionQueued` to `FactorInstancesDerived` is not directly supported by `set_provisional`, I've exposed a dedicated method for it called `cancel_queued_transaction`, making it clear that we should first actually cancel the transaction and removed it from the queue, and **then** call `account.entity_security_state.cancel_queued_transaction(..)`.


README.md
CyonAlexRDX, line 1: bah, just pressed saved... and made some minor change

crates/sargon-uniffi/src/profile/mfa/secured_entity_control/secured_entity_control.rs
```diff
@@ -17,4 +23,8 @@ pub struct SecuredEntityControl {
     /// The believed-to-be-current security structure of FactorInstances which
     /// secures this entity.
     pub security_structure: SecurityStructureOfFactorInstances,
+
+    /// A provisional new security structure configuration which user
+    /// is about to change to
+    pub provisional: Option<ProvisionalSecurifiedConfig>,
```
CyonAlexRDX, line 29: Alternatives to `provisional`:
* `draft` - discarded since it sounds like it is not "saved" or otherwise incomplete. 
* `pending` - make it sound like it has been submitted to the network and waiting for a tx complete
* `user_intention` - some version of this might be ok.

Definition of `provisional`: `"arranged or existing for the present, possibly to be changed later."`

crates/sargon-uniffi/src/profile/v100/entity_security_state/provisional_securified_config.rs
```diff
@@ -0,0 +1,30 @@
...
+};
+
+#[derive(Clone, PartialEq, Eq, Hash, InternalConversion, uniffi::Record)]
+pub struct ProvisionalSecurifiedTransactionQueued {
+    pub factor_instances: SecurityStructureOfFactorInstances,
+    pub txid: TransactionIntentHash,
+}
+
+#[derive(Clone, PartialEq, Eq, Hash, InternalConversion, uniffi::Enum)]
+pub enum ProvisionalSecurifiedConfig {
```
CyonAlexRDX, line 14: doc me!

crates/sargon-uniffi/src/profile/v100/entity_security_state/provisional_securified_config.rs
```diff
@@ -0,0 +1,30 @@
+use crate::prelude::*;
+use sargon::{
+    ProvisionalSecurifiedConfig as InternalProvisionalSecurifiedConfig,
+    ProvisionalSecurifiedTransactionQueued as InternalProvisionalSecurifiedTransactionQueued,
+};
+
+#[derive(Clone, PartialEq, Eq, Hash, InternalConversion, uniffi::Record)]
+pub struct ProvisionalSecurifiedTransactionQueued {
```
CyonAlexRDX, line 8: doc me!

crates/sargon/src/profile/mfa/security_structures/security_structure_of_factors/security_structure_of_factor_instances.rs
CyonAlexRDX, line 1: Moved, and added `authentication_signing_factor_instance:
        HierarchicalDeterministicFactorInstance,`

crates/sargon/src/profile/mfa/security_structures/security_structure_of_factors/security_structure_of_factor_source_ids.rs
```diff
@@ -6,56 +6,6 @@ pub type SecurityStructureOfFactorSourceIds =
...
-    /// and Confirmation role.
-    pub matrix_of_factors: MatrixOfFactorInstances,
-}
-
-impl SecurityStructureOfFactorInstances {
-    pub fn new(
-        security_structure_id: SecurityStructureID,
-        matrix_of_factors: MatrixOfFactorInstances,
-    ) -> Self {
-        Self {
```
CyonAlexRDX, line 28: was not throwing, is now!

crates/sargon/src/profile/v100/entity_security_state/entity_security_state.rs
```diff
@@ -7,19 +7,218 @@ use crate::prelude::*;
...
+            }
+            Self::Securified { value } => {
+                value.set_provisional_unchecked(provisional)
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod has_provisional_securified_config_tests {
```
CyonAlexRDX, line 51: move down to `mod tests`

crates/sargon/src/profile/v100/entity_security_state/provisional_securified_config.rs
```diff
@@ -0,0 +1,108 @@
+use crate::prelude::*;
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, Hash)]
+pub struct ProvisionalSecurifiedTransactionQueued {
```
CyonAlexRDX, line 4: doc!

crates/sargon/src/profile/v100/entity_security_state/provisional_securified_config.rs
```diff
@@ -0,0 +1,108 @@
...
+    fn inequality() {
+        assert_ne!(SUT::sample(), SUT::sample_other());
+    }
+}
+
+#[derive(
+    Serialize, Deserialize, Clone, Debug, PartialEq, Eq, Hash, EnumAsInner,
+)]
+#[serde(tag = "discriminator")]
+pub enum ProvisionalSecurifiedConfig {
```
CyonAlexRDX, line 57: doc type!

crates/sargon/fixtures/transaction/create_access_controller_for_account.rtm
```diff
@@ -0,0 +1,98 @@
...
+;
+CREATE_ACCESS_CONTROLLER
+    Bucket("bucket1")
+    Tuple(
+        Enum<2u8>(
+            Enum<1u8>(
+                Array<Enum>(
+                    Enum<0u8>(
+                        Enum<2u8>(
+                            2u8,
```
CyonAlexRDX, line 18: this `2` is threshold

crates/sargon/fixtures/transaction/create_access_controller_for_account.rtm
```diff
@@ -0,0 +1,98 @@
...
+    "securify"
+;
+TAKE_FROM_WORKTOP
+    Address("resource_rdx1nfxxxxxxxxxxaccwnrxxxxxxxxx006664022062xxxxxxxxxaccwnr")
+    Decimal("1")
+    Bucket("bucket1")
+;
+CREATE_ACCESS_CONTROLLER
+    Bucket("bucket1")
+    Tuple(
```
CyonAlexRDX, line 12: @0xOmarA the `RuleSet` uses **position** (index) to determine which `AccessRule` is `primary`, `recovery` and `confirmation`? Because The values of the `Enum` seems to be the same?

```
        Enum<2u8>(
            Enum<1u8>(
                Array<Enum>(
```
for all three?

crates/sargon/src/profile/mfa/security_structures/matrices/matrix_of_factor_instances.rs
```diff
@@ -105,12 +173,17 @@ impl MnemonicWithPassphrase {
 impl MatrixOfFactorInstances {
     fn sample_from_matrix_of_sources(
         matrix_of_sources: MatrixOfFactorSources,
+        entity_kind: CAP26EntityKind,
```
CyonAlexRDX, line 176: we now have stricter validation so that we do not accidentally use Account factorinstances for Persona or vice versa.

crates/sargon/src/profile/mfa/security_structures/roles/factor_levels/factor_instance_level/role_into_scrypto_access_rule.rs
```diff
@@ -0,0 +1,69 @@
...
+            primary_role: primary_role.into(),
+            recovery_role: recovery_role.into(),
+            confirmation_role: confirmation_role.into(),
+        }
+    }
+}
+
+pub const MINUTES_PER_DAY: u32 = 24 * 60;
+
+impl From<SecurityStructureOfFactorInstances> for ScryptoRecoveryProposal {
```
CyonAlexRDX, line 48: this PR does not make use of this `From for ScryptoRecoveryProposal`

crates/sargon/src/profile/mfa/security_structures/roles/factor_levels/factor_instance_level/role_into_scrypto_access_rule.rs
```diff
@@ -0,0 +1,69 @@
...
+            ScryptoCompositeRequirement::BasicRequirement(
+                ScryptoBasicRequirement::AnyOf(from_factors(
+                    value.get_override_factors(),
+                )),
+            ),
+        ]))
+    }
+}
+
+impl From<MatrixOfFactorInstances> for ScryptoRuleSet {
```
CyonAlexRDX, line 29: This PR does in fact not make use of this `From for ScryptoRuleSet` since the ManifestBuilding method in Scrypto takes each role's AccessRule individiaully, see: https://github.com/radixdlt/sargon/pull/293/files#diff-466895b6bcee7ef733f42aa6a1077061bda36ee10b372bda4f68a3004a86f3d7R68-R74

crates/sargon/fixtures/transaction/create_access_controller_for_account.rtm
0xOmarA: It does indeed use the position to determine that since these are just arguments to the function.

crates/sargon-uniffi/src/profile/mfa/secured_entity_control/secured_entity_control.rs
giannis-rdx: `Temporal`?

crates/sargon/src/profile/mfa/secured_entity_control/secured_entity_control.rs
```diff
@@ -26,13 +26,103 @@ pub struct SecuredEntityControl {
...
+                        return self.set_provisional_unchecked_ok(provisional);
+                    }
+                }
+                // We have already consumed FactorInstances from the FactorInstancesCache.
+                // We currently do not have a way of putting these back. We don't want to 
+                // create gaps if we can. Wallet Features might change, to REQUIRE us to
+                // allow this, if so we can change this - allowing transition from 
+                // `FactorInstancesDerived` to Non-TransactionQueued state.
+                Err(
+                CommonError::SecurityEntityControlCannotChangeProvisionalAlreadyDerivedInstances,
```
GhenadieVP, line 79: This can happen quite easily if user updates a shield a second time before applying it. So most likely we will need to handle this, let's discuss in design session how to do it.

crates/sargon/src/profile/mfa/security_structures/roles/factor_levels/factor_instance_level/role_into_scrypto_access_rule.rs
```diff
@@ -0,0 +1,56 @@
...
+    ) -> Self {
+        Self {
+            primary_role: primary_role.into(),
+            recovery_role: recovery_role.into(),
+            confirmation_role: confirmation_role.into(),
+        }
+    }
+}
+
+pub const MINUTES_PER_DAY: u32 = 24 * 60;
```
GhenadieVP, line 46: We do have a constants file

crates/sargon/src/profile/mfa/security_structures/security_structure_of_factors/security_structure_of_factor_instances.rs
```diff
@@ -0,0 +1,500 @@
...
+impl SecurityStructureOfFactorInstances {
+    pub fn new(
+        security_structure_id: SecurityStructureID,
+        matrix_of_factors: MatrixOfFactorInstances,
+        authentication_signing: HierarchicalDeterministicFactorInstance,
+    ) -> Result<Self> {
+        let index_agnostic_path = matrix_of_factors
+            .index_agnostic_path_of_all_tx_signing_factor_instances()?;
+        let entity_kind = index_agnostic_path.entity_kind;
+        matrix_of_factors.assert_has_entity_kind(entity_kind)?;
```
GhenadieVP, line 53: Is this assert needed here?

crates/sargon/src/profile/v100/profile.rs
```diff
@@ -294,7 +294,7 @@ impl Profile {
             .items()
             .into_iter()
             .map(Into::<AccountOrPersona>::into)
-            .map(|e| (e.clone(), e.unique_factor_instances()))
+            .map(|e| (e.clone(), e.unique_tx_signing_factor_instances()))
```
GhenadieVP, line 297: Why not all instances including authentication_signing?

GhenadieVP (commented):
Nice work! Left some minor-ish comments.

crates/sargon-uniffi/src/profile/mfa/secured_entity_control/secured_entity_control.rs
Sajjon: You mean `temporary` I think? Hmm `temporary` relates more to time, whereas I'm going for a term which relates more to step in a sequence of steps, which I feel is a little bit more precise. And "temporary" also has a meaning that after some time it is no longer relevant, whereas what is actually happening here is that it holds the state which will be the future state, so not at all temporary, just that we are note quite there yet.

crates/sargon/src/profile/logic/profile_network/profile_network_details.rs
```diff
@@ -118,12 +118,7 @@ impl AuthorizedPersonaSimple {
 
         let shared_accounts = self.accounts_for_display(non_hidden_accounts)?;
 
-        let has_auth_signing_key = match &persona.security_state {
-            EntitySecurityState::Unsecured { value: uec } => {
-                uec.authentication_signing.is_some()
-            }
-            _ => false, // TODO change that
-        };
+        let has_auth_signing_key = persona.is_securified();
```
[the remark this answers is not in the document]
Sajjon: `AuthorizedPersonaDetailed` struct contains that field. Not sure we use it anywhere in hosts, but I did not wanna change that. And the assignment ` let has_auth_signing_key = persona.is_securified();` is correct. ALL Securified Entties has an auth signing factor. No Unsecurified has one.

crates/sargon/src/profile/mfa/security_structures/roles/factor_levels/factor_instance_level/role_into_scrypto_access_rule.rs
Sajjon: constants which have a "natural home" - should be put in a natural home I think. That is most Rusty I think. The constants file seems to contain mostly constants relating to GW?

crates/sargon/src/profile/mfa/security_structures/security_structure_of_factors/security_structure_of_factor_instances.rs
Sajjon: I should perhaps call it `validate` it does panic, it throws. 

And yes it is good to be quite... cautious when it comes to FactorInstance assignment - as we have seen with e.g. the unfortunate Android bug. I made the `assert_has_entity_kind`  validation easy reusable so that we can be cautious. 

Mostly ensuring we don't do anything wrong "internally" - since UI should not allow for this.

crates/sargon/src/profile/v100/profile.rs
Sajjon: good catch, this should have been `unique_all_factor_instances`, thx, will change

crates/sargon/src/profile/mfa/security_structures/security_structure_of_factors/security_structure_of_factor_instances.rs
GhenadieVP: But it seems redundant here, assert_has_entity will perform the exact same steps as the two line above and then compare the entitiy_kind. Or I am missing something?

crates/sargon/src/profile/mfa/security_structures/security_structure_of_factors/security_structure_of_factor_instances.rs
micbakos-rdx: Thank you

crates/sargon/src/profile/mfa/security_structures/security_structure_of_factors/security_structure_of_factor_instances.rs
Sajjon: Thanks, no you are right, thanks! calling `index_agnostic_path_of_all_tx_signing_factor_instances()?` is enough!

crates/sargon/src/profile/logic/profile_network/profile_network_details.rs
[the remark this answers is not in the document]
micbakos-rdx: @CyonAlexRDX We discussed using `transaction_signing` factor instance when an entity is not securified. Doesn't that mean that `has_auth_signing_key` will always be true? Or does `has_auth_signing_key` mean like a true ROLA key?

crates/sargon/src/profile/mfa/security_structures/roles/factor_levels/factor_instance_level/role_into_scrypto_access_rule.rs
GhenadieVP: well I agree, at the same time `MINUTES_PER_DAY` is very generic. Anyway, this is minor, we'll migrate in case there will be other components using this.

Sajjon: right, ok agreed, this should be moved, since it is common relating to time!

micbakos-rdx (approved):

crates/sargon/src/profile/logic/profile_network/profile_network_details.rs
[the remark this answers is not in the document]
Sajjon: Rather `has_auth_signing_key` will always be `false` for unsecurified entities. But unsecurified entities can ofc proof-ownership (which it does using `transaction_signing`.)

I'm not sure this field is at all relevant. I think it might have been used by iOS behind `DEBUG` flag. So we should not put too much importance on this field.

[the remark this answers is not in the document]
micbakos-rdx: Thanks Alex, indeed it is used to disable the button that created the rola key for unsecurified entities. Indeed unneeded.

PR #294, opened 2024-12-06
Deriving keys purpose

micbakos-rdx:
Informs the host about the purpose of the incoming keys derivation interaction.

The new type is called `DerivationPurpose`, stored in `KeysCollectorDependencies` and passed down to the interactor. Each `FactorInstancesProvider` usage, provides a different purpose according to the operation it is about to derive keys for.

### DIFF
--- a/crates/sargon-uniffi/src/keys_collector/key_derivation_request.rs
+++ b/crates/sargon-uniffi/src/keys_collector/key_derivation_request.rs
@@ -5,7 +5,7 @@ use sargon::{IndexSet, KeyDerivationRequest as InternalKeyDerivationRequest};
 /// A collection of derivation paths, on a per-factor-source basis.
 #[derive(Clone, PartialEq, Eq, uniffi::Record)]
 pub struct KeyDerivationRequest {
-    pub keys_collection_reason: KeysCollectionReason,
+    pub derivation_purpose: DerivationPurpose,
     pub per_factor_source: Vec<KeyDerivationRequestPerFactorSource>,
 }
 
@@ -36,7 +36,7 @@ impl KeyDerivationRequest {
 impl From<InternalKeyDerivationRequest> for KeyDerivationRequest {
     fn from(value: InternalKeyDerivationRequest) -> Self {
         Self {
-            keys_collection_reason: value.keys_collection_reason.into(),
+            derivation_purpose: value.derivation_purpose.into(),
             per_factor_source: value
                 .per_factor_source
                 .into_iter()
@@ -54,7 +54,7 @@ impl From<InternalKeyDerivationRequest> for KeyDerivationRequest {
 impl From<KeyDerivationRequest> for InternalKeyDerivationRequest {
     fn from(value: KeyDerivationRequest) -> Self {
         Self::new(
-            value.keys_collection_reason.into_internal(),
+            value.derivation_purpose.into_internal(),
             IndexMap::from_iter(value.per_factor_source.into_iter().map(|f| {
                 (
                     f.factor_source_id.into_internal(),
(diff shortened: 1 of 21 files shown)


crates/sargon/src/keys_collector/collector/derivation_purpose.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    pub fn new_for_securifying(entity_kind: CAP26EntityKind) -> Self {
+        match entity_kind {
+            CAP26EntityKind::Account => Self::SecurifyingAccount,
+            CAP26EntityKind::Identity => Self::SecurifyingPersona,
+        }
+    }
+
+    pub fn new_for_pre_derivation() -> Self {
```
Sajjon, line 43: this function has a strange name IMO. I think it should be called `cache_filling` or `pre_derivation_for_cache` or simply just `pre_deriving_keys` (not a super necessary function)

crates/sargon/src/keys_collector/collector/keys_collector_dependencies.rs
```diff
@@ -14,16 +14,21 @@ pub(crate) struct KeysCollectorDependencies {
     /// computer and thus unable to make a connection between the Radix Wallet
     /// and a Ledger device.
     pub(super) factors_of_kind: IndexSet<FactorSourcesOfKind>,
+
+    /// The reason that the collection of keys was initiated
+    pub(super) collection_reason: DerivationPurpose,
```
Sajjon, line 19: I would call this field `derivation_purpose` as well I think

crates/sargon/src/keys_collector/collector/derivation_purpose.rs
```diff
@@ -0,0 +1,91 @@
...
+
+impl DerivationPurpose {
+    pub fn new_for_creating(entity_kind: CAP26EntityKind) -> Self {
+        match entity_kind {
+            CAP26EntityKind::Account => Self::CreatingNewAccount,
+            CAP26EntityKind::Identity => Self::CreatingNewPersona,
+        }
+    }
+
+    pub fn new_for_securifying(entity_kind: CAP26EntityKind) -> Self {
```
Sajjon, line 36: `new_for_securifying` -> `for_securifying_or_updating` (drop `new` and include `updating` ) I think

crates/sargon/src/keys_collector/collector/derivation_purpose.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// for identity MFA
+    SecurifyingPersona,
+
+    /// When adding a new factor source, initiates keys collection
+    /// for collecting various factor instances.
+    PreDerivingKeys,
+}
+
+impl DerivationPurpose {
+    pub fn new_for_creating(entity_kind: CAP26EntityKind) -> Self {
```
Sajjon, line 29: `new_for_creating` -> `creation_of_new_virtual_entity` ?

crates/sargon-uniffi/src/keys_collector/key_derivation_request.rs
```diff
@@ -1,11 +1,30 @@
 use crate::prelude::*;
-use sargon::indexmap::IndexMap;
+use sargon::IndexMap;
 use sargon::{IndexSet, KeyDerivationRequest as InternalKeyDerivationRequest};
 
 /// A collection of derivation paths, on a per-factor-source basis.
 #[derive(Clone, PartialEq, Eq, uniffi::Record)]
 pub struct KeyDerivationRequest {
-    pub per_factor_source: HashMap<FactorSourceIDFromHash, Vec<DerivationPath>>,
+    pub derivation_purpose: DerivationPurpose,
```
Sajjon, line 8: maybe include one line doc as to how we will use `derivation_purpose` field? something like:

```
/// We include this DerivationPurpose in dispatched use FactorSource requests to host so that UI 
/// can display contextual information as to why the user is prompted to authenticate FactorSource access.
```

crates/sargon/src/keys_collector/collector/derivation_purpose.rs
micbakos-rdx: We use this `new_*` for whenever we instantiate any type. How can someone know when new is suitable or not?

micbakos-rdx: Personally I would always go with a descriptive name like `for_securifying_or_updating`, but I tried to be consistent with the rest of the codebase.

PR #297, opened 2024-12-09
[ABW-3969] Shield Setup Select Factors - utility functions

danvleju-rdx:
## Changelog
- Added `sorted_factor_sources_for_selection function`, which handles the logic for sorting a list of factor sources on the `Select Factors` screen.
- Added `selected_factor_sources_status` function, which returns a newly added enum, `SelectedFactorSourcesStatus`, representing the status of selected factor sources in the Security Shield building process.
- [Swift] Fixes the `allCases` for `SecurityProblemKind`

### DIFF
--- a/apple/Sources/Sargon/Extensions/Methods/Profile/MFA/SecurityShieldBuilder+Wrap+Functions.swift
+++ b/apple/Sources/Sargon/Extensions/Methods/Profile/MFA/SecurityShieldBuilder+Wrap+Functions.swift
@@ -1,0 +1,12 @@
+import Foundation
+import SargonUniFFI
+
+extension SecurityShieldBuilder {
+	public static func sortFactorSourcesForSelection(factorSources: [FactorSource]) -> [FactorSource] {
+		securityShieldBuilderSortFactorSourcesForSelection(factorSources: factorSources)
+	}
+
+	public static func selectedFactorSourcesStatus(factorSources: [FactorSource]) -> SelectedFactorSourcesStatus {
+		securityShieldBuilderSelectedFactorSourcesStatus(factorSources: factorSources)
+	}
+}
--- a/apple/Sources/Sargon/Extensions/Methods/Profile/MFA/SecurityShieldBuilder+Wrap+Functions.swift
+++ b/apple/Sources/Sargon/Extensions/Methods/Profile/MFA/SecurityShieldBuilder+Wrap+Functions.swift
@@ -5,8 +5,4 @@ extension SecurityShieldBuilder {
 	public static func sortedFactorSourcesForSelection(factorSources: [FactorSource]) -> [FactorSource] {
 		securityShieldBuilderSortedFactorSourcesForSelection(factorSources: factorSources)
 	}
-
-	public static func selectedFactorSourcesStatus(factorSources: [FactorSource]) -> SelectedFactorSourcesStatus {
-		securityShieldBuilderSelectedFactorSourcesStatus(factorSources: factorSources)
-	}
 }
--- a/apple/Sources/Sargon/Extensions/Methods/Profile/MFA/SecurityShieldBuilder+Wrap+Functions.swift
+++ b/apple/Sources/Sargon/Extensions/Methods/Profile/MFA/SecurityShieldBuilder+Wrap+Functions.swift
@@ -2,8 +2,8 @@ import Foundation
 import SargonUniFFI
 
 extension SecurityShieldBuilder {
-	public static func sortFactorSourcesForSelection(factorSources: [FactorSource]) -> [FactorSource] {
-		securityShieldBuilderSortFactorSourcesForSelection(factorSources: factorSources)
+	public static func sortedFactorSourcesForSelection(factorSources: [FactorSource]) -> [FactorSource] {
+		securityShieldBuilderSortedFactorSourcesForSelection(factorSources: factorSources)
 	}
 
 	public static func selectedFactorSourcesStatus(factorSources: [FactorSource]) -> SelectedFactorSourcesStatus {
--- a/apple/Sources/Sargon/Extensions/Methods/Profile/MFA/SecurityShieldBuilder+Wrap+Functions.swift
+++ b/apple/Sources/Sargon/Extensions/Methods/Profile/MFA/SecurityShieldBuilder+Wrap+Functions.swift
@@ -1,8 +1,0 @@
-import Foundation
-import SargonUniFFI
-
-extension SecurityShieldBuilder {
-	public static func sortedFactorSourcesForSelection(factorSources: [FactorSource]) -> [FactorSource] {
-		securityShieldBuilderSortedFactorSourcesForSelection(factorSources: factorSources)
-	}
-}
--- a/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
+++ b/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
@@ -411,7 +411,7 @@ pub fn security_shield_builder_sort_factor_sources_for_selection(
 }
 
 #[uniffi::export]
-pub fn selected_factor_sources_status(
+pub fn security_shield_builder_selected_factor_sources_status(
     factor_sources: Vec<FactorSource>,
 ) -> SelectedFactorSourcesStatus {
     SelectedFactorSourcesStatus::from(
--- a/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
+++ b/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
@@ -400,11 +400,11 @@ impl SecurityShieldBuilder {
 // ==== STATIC ====
 // ================
 #[uniffi::export]
-pub fn security_shield_builder_sort_factor_sources_for_selection(
+pub fn security_shield_builder_sorted_factor_sources_for_selection(
     factor_sources: Vec<FactorSource>,
 ) -> Vec<FactorSource> {
     let factors =
-        InternalSecurityShieldBuilder::sort_factor_sources_for_selection(
+        InternalSecurityShieldBuilder::sorted_factor_sources_for_selection(
             factor_sources.into_internal(),
         );
     factors.into_iter().map(FactorSource::from).collect()
--- a/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
+++ b/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
@@ -406,6 +406,19 @@ impl SecurityShieldBuilder {
         })
     }
 
+    pub fn sorted_factor_sources_for_primary_threshold_selection(
+        &self,
+        factor_sources: Vec<FactorSource>,
+    ) -> Vec<FactorSource> {
+        self.get(|builder| {
+            builder
+                .sorted_factor_sources_for_primary_threshold_selection(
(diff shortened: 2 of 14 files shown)


crates/sargon/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -459,6 +459,31 @@ impl SecurityShieldBuilder {
...
+        let mut sorted = factor_sources;
+        sorted.sort_by_key(|fs| fs.factor_source_kind().display_order());
+        sorted
+    }
+
+    pub fn selected_factor_sources_status(
+        factor_sources: Vec<FactorSource>,
+    ) -> SelectedFactorSourcesStatus {
+        if factor_sources.is_empty() {
+            SelectedFactorSourcesStatus::Insufficient
```
Sajjon, line 478: Hmm `1` is never enough for a whole shield. So this must be PER ROLE and which role is it? `Primary` ("Access")?

If so the enum and this method should make it clear that it is for `Primary`. 

(Or if it is dynamic it should be `for_role: RoleKind` as input, and then we should not return just an enum but rather a struct with the status and the role)

crates/sargon/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -459,6 +459,31 @@ impl SecurityShieldBuilder {
...
 
+// ================
+// ==== STATIC ====
+// ================
+impl SecurityShieldBuilder {
+    pub fn sorted_factor_sources_for_selection(
+        factor_sources: Vec<FactorSource>,
+    ) -> Vec<FactorSource> {
+        let mut sorted = factor_sources;
+        sorted.sort_by_key(|fs| fs.factor_source_kind().display_order());
```
Sajjon, line 470: the perpetual naming challenge for these kind of things... :) I think `let mut sorted = factor_sources` is misleading, since it is in fact not sorted yet!

Luckily In Rust we have ownership!

so:
```rust
let mut unsorted = factor_sources;
unsorted.sort....
let sorted = unsorted;
// no `unsorted` can never be referenced again! it has been consumed! Rust owns!
sorted
```

crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -394,6 +396,31 @@ impl SecurityShieldBuilder {
...
+// ================
+#[uniffi::export]
+pub fn security_shield_builder_sorted_factor_sources_for_selection(
+    factor_sources: Vec<FactorSource>,
+) -> Vec<FactorSource> {
+    let factors =
+        InternalSecurityShieldBuilder::sorted_factor_sources_for_selection(
+            factor_sources.into_internal(),
+        );
+    factors.into_iter().map(FactorSource::from).collect()
```
Sajjon, line 410: I think:

```rust
    factors.into_iter().map(FactorSource::from).collect()
```

can be changed to **just**:

```rust
    factors.into_type()
```

apple/Sources/Sargon/Extensions/Methods/Profile/MFA/SecurityShieldBuilder+Wrap+Functions.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+import SargonUniFFI
+
+extension SecurityShieldBuilder {
```
Sajjon, line 4: 👍

crates/sargon-uniffi/src/profile/mfa/security_structures/selected_factor_sources_status.rs
```diff
@@ -0,0 +1,17 @@
+use crate::prelude::*;
+use sargon::SelectedFactorSourcesStatus as InternalSelectedFactorSourcesStatus;
+
+/// Represents the status of selected factor sources in the Security Shield building process.
+#[derive(
+    Clone, Copy, Debug, PartialEq, Eq, InternalConversion, uniffi::Enum,
+)]
+pub enum SelectedFactorSourcesStatus {
```
CyonAlexRDX, line 8: `SelectedFactorSourcesForRoleStatus` ?

crates/sargon/src/profile/mfa/security_structures/security_shield_builder.rs
CyonAlexRDX: Post-our sync:
**Delegate this to `MatrixBuilder`**

In MatrixBuilder:
We should probably NOT let this be a STATIC, even if we wont use the internal state of the `MatrixBuilder` **initially** we still might want to in the future. So I would do these changes:
* change it to be on `&self` (i.e. not "static")
* change it to take a `for_role: RoleKind`

CyonAlexRDX (changes requested):

apple/Sources/Sargon/Extensions/Methods/Profile/MFA/SecurityShieldBuilder+Wrap+Functions.swift
danvleju-rdx: Kept `sortedFactorSourcesForSelection` as a static function since it’s just a utility function and doesn’t need to involve the shield builder or imply that it can be modified.

crates/sargon-uniffi/src/profile/mfa/security_structures/selected_factor_sources_status.rs
```diff
@@ -0,0 +1,29 @@
...
+    Suboptimal,
+
+    /// The selected factor sources are insufficient for the specified role
+    /// in the Security Shield building process.
+    Insufficient,
+
+    /// The selected factor sources form an invalid combination for the specified role
+    /// in the Security Shield building process.
+    ///
+    /// Example: A Password factor source cannot be used alone for the Primary role.
```
GhenadieVP, line 27: This example seems to fall more under the `Insufficient` category

crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -382,6 +384,28 @@ impl SecurityShieldBuilder {
...
+        role: RoleKind,
+    ) -> Option<SecurityShieldBuilderInvalidReason> {
+        self.get(|builder| {
+            builder
+                .validate_role_in_isolation(role.into_internal())
+                .map(|x| x.into())
+        })
+    }
+
+    pub fn selected_factor_sources_for_role_status(
```
GhenadieVP, line 398: So this is meant to be used in Select Factors for Transactions screen? If so, do we need a an instance of a builder to validate the selected combination of the factors?

Additionally, the fact that it returns `SelectedFactorSourcesForRoleStatus`, implies that the builder might contain invalid factors, but AFAIK we are actively trying to prevent this.

crates/sargon/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -459,6 +506,19 @@ impl SecurityShieldBuilder {
     }
 }
 
+// ================
+// ==== STATIC ====
+// ================
+impl SecurityShieldBuilder {
+    pub fn sorted_factor_sources_for_selection(
```
GhenadieVP, line 513: Isn't this specific for Select Factors for Transactions, aka for Access role? If so, then it should be specific. Also not all factor sources can be used for Access role, thus some filtering need to happen as well AFAIK.

crates/sargon/src/profile/v100/factors/factor_source_kind.rs
```diff
@@ -118,6 +118,20 @@ impl FactorSourceKind {
     }
 }
 
+impl FactorSourceKind {
+    pub fn display_order(&self) -> u8 {
```
GhenadieVP, line 122: IMO in the context of this PR, it should be specific display order for Select Factor for Access Role, not generic order.

crates/sargon-uniffi/src/profile/mfa/security_structures/selected_factor_sources_status.rs
danvleju-rdx: From the host’s perspective, this should fall under the `Invalid` category and display a generic error message. If the Primary role has `threshold` set to `1`, then adding a Password factor source satisfies the threshold condition, but it’s still invalid. It could fall under the `Insufficient` category if we have a dynamic threshold that updates based on the available factor sources. Do we?

crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
danvleju-rdx: > do we need an instance of a builder

Not necessarily, @CyonAlexRDX suggested this. Also, it seems easier since all the current validation functions are on a builder instance.

>  implies that the builder might contain invalid factors

Not the factors themselves, but in the context of a specific role.

crates/sargon-uniffi/src/profile/mfa/security_structures/selected_factor_sources_status.rs
GhenadieVP: Why is threshold important in the context of selecting the factors before going in the Access setup build screen?

Also,[ if there are no roles selected the status is Insufficient](https://github.com/radixdlt/sargon/blob/a37ad774dd6fafcaaa2ee6800cf1b92101630fe0/crates/sargon/src/profile/mfa/security_structures/security_shield_builder.rs#L455), but if you have password it is considered Invalid, when IMO it is a case of insufficient - you need one more factor.

CyonAlexRDX (approved):
LGTM .

We can always iterate if we need adjust it

crates/sargon-uniffi/src/profile/mfa/security_structures/selected_factor_sources_status.rs
danvleju-rdx: This is a very UI specific type and we have in the UI 3 error/warning states:
1. User need to select at least 1 factor -> `Insufficient`
2. It is recommended to select at least 2 factors -> `Suboptimal`
3. Any other invalid combinations for a specific role (Access) -> `Invalid`

Based on that, this case falls under `Invalid`, even if it doesn't sound correct

crates/sargon/src/profile/mfa/security_structures/security_shield_builder.rs
danvleju-rdx: Security questions and trusted contact (not supported in MVP) cannot be used for Access role, we can filter them out

crates/sargon-uniffi/src/profile/mfa/security_structures/selected_factor_sources_status.rs
GhenadieVP: I am confused here - what is any other combinations? Shouldn't we display only the factors that are allowed to be selected in treshold Access role?
So then the only "Invalid" scenario is that the password cannot be alone, some other might be added in the future sure.

> 1. User need to select at least 1 factor -> Insufficient
> 2. It is recommended to select at least 2 factors -> Suboptimal
> 3. Any other invalid combinations for a specific role (Access) -> Invalid

So the Wallet will display red for 1 and 3?

crates/sargon/src/profile/mfa/security_structures/security_shield_builder.rs
GhenadieVP: correct, these should be filtered out accordingly.

GhenadieVP (approved):
Needs to include the filtering also

PR #298, opened 2024-12-09
Expose sign auth

micbakos-rdx:
Signing auth is used in login flows by the host. In similar fashion with `sign_transaction` and `sign_subintent`, `sign_auth` is exposed by sargon os. The `AuthenticationSigningInteractor` is added in `UseFactorSourcesInteractor`. Notable fact is that `RolaChallenge` also exposes the `hash` method used by host in order to sign it.

### DIFF
--- a/crates/sargon-uniffi/src/signing/authentication_signing_input.rs
+++ b/crates/sargon-uniffi/src/signing/authentication_signing_input.rs
@@ -8,6 +8,29 @@ pub struct AuthenticationSigningInput {
     /// with.
     pub owned_factor_instance: OwnedFactorInstance,
 
-    /// The challenge that will be signed by `owned_factor_instance`
-    pub challenge: RolaChallenge,
+    /// The challenge nonce that with some `metadata` values are generating the `RolaChallenge`
+    /// needed to be signed
+    pub challenge_nonce: DappToWalletInteractionAuthChallengeNonce,
+
+    /// The metadata that together with the `challenge_nonce` are generating the `RolaChallenge`
+    /// needed to be signed
+    pub metadata: DappToWalletInteractionMetadata,
+}
+
+#[uniffi::export]
+pub fn authentication_signing_input_get_rola_challenge(
+    input: &AuthenticationSigningInput,
+) -> Result<RolaChallenge> {
+    input.into_internal().rola_challenge().into_result()
+}
+
+#[uniffi::export]
+pub fn new_authentication_signing_input_sample() -> AuthenticationSigningInput {
+    InternalAuthenticationSigningInput::sample().into()
+}
+
+#[uniffi::export]
+pub fn new_authentication_signing_input_sample_other(
+) -> AuthenticationSigningInput {
+    InternalAuthenticationSigningInput::sample_other().into()
 }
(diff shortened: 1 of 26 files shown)


crates/sargon-uniffi/src/signing/authentication_signing_response.rs
```diff
@@ -0,0 +1,7 @@
+use crate::prelude::*;
+use sargon::AuthenticationSigningResponse as InternalAuthenticationSigningResponse;
+
+#[derive(Clone, PartialEq, InternalConversion, uniffi::Record)]
+pub struct AuthenticationSigningResponse {
+    pub signature_with_public_key: SignatureWithPublicKey,
```
CyonAlexRDX, line 6: Perhaps include the challenge, RolaChallenge, here?

crates/sargon-uniffi/src/signing/rola_challenge.rs
```diff
@@ -0,0 +1,22 @@
+use crate::prelude::*;
+use sargon::RolaChallenge as InternalRolaChallenge;
+
+#[derive(Debug, Clone, PartialEq, InternalConversion, uniffi::Record)]
+pub struct RolaChallenge {
+    pub payload: BagOfBytes,
```
CyonAlexRDX, line 6: Need
To include the whole ROLAChallege, including host and origin for Ledger HW

micbakos-rdx: Why? Those data are provided in `AuthenticationSigner`!

CyonAlexRDX: Sure, depends on much back and forth one wants to do. But I think we want all sign input to be self contained.

On iOS we might be in global scope
When we impl the host
Interactor.
Thus we must send the full rolachallenge as sign input, since
ledger app requires it

CyonAlexRDX: I'm trying to say that it might be two disjoint app contexts: where we from host call Sargon to ROLA Sign and where host is told by Sargon to sign ROLA

micbakos-rdx: I see what you mean, will do that

micbakos-rdx: Changed the `AuthenticationSigningInput` to contain the `nonce` and the `metadata`. We can get the `RolaChallenge` by doing:

```
input.rola_challenge()
```

With that way we can
1. Pass down to the host, the input that was requested
2. Deterministically derive the challenge using sargon os.
3. Validity checks still in place. For example checks the validity of the metadata's input url.

crates/sargon-uniffi/src/signing/authentication_signing_input.rs
```diff
@@ -0,0 +1,13 @@
+use crate::prelude::*;
+use sargon::AuthenticationSigningInput as InternalAuthenticationSigningInput;
+
+#[derive(Clone, PartialEq, InternalConversion, uniffi::Record)]
```
CyonAlexRDX, line 4: Can't remember, does InternalConversion also generate sample
Values? Else maybe we wanna create and UniFFI export sample
Values!?

micbakos-rdx: Did that, will push soon.

CyonAlexRDX: And that goes for all UniFFI exported types :)

CyonAlexRDX (approved):
LGTM

PR #299, opened 2024-12-10
Automatic Assignment of Factors to Recovery & Confirmation Role

CyonAlexRDX:
Implementation of ["Automatic Security Shield Construction"][doc].

> [!TIP]
> The name "Automatic Security Shield Construction" is slightly misleading since it is actually just pre-selection of factors for `Recovery` and `Confirmation` roles **after already having selected factors for `Primary** role

> [!TIP]
> Hosts can call `auto_assign_factors_to_recovery_and_confirmation_based_on_primary` on `SecurityShieldBuilder` after having set the Primary role factors, and providing a list of **ALL** FactorSources in Profile (including those assigned to Primary).
> The `SecurityShieldBuilder` will then internally dispatch to `AutomaticShieldBuilder` which is an internal
> type which will assign the factors to the Recovery and Confirmation roles.

I did not really understand the instruction:
> Distribute to try to get up to 2 START and then 2 CONFIRM factors if possible
>
> Add any Custodian factors in the list, starting with the most recently used, to START until there are exactly 2 factors in START.
> 
> Add any Hardware (Ledger, Arculus, Yubikey) factors in the list, starting with the most recently used, to START until there are exactly 2 factors in START.

Or rather, I think there are two both reasonable, interpretations of this, and I've implemented the first and then the second and current head is using the second:
1. [`interpretation if < limit; add` (hash: `45be07f`)](https://github.com/radixdlt/sargon/pull/299/commits/45be07face748481ad91c87c8a8cbba43c1eefb6). 
2. [`interpretation while < limit; add` (hash: `de17433`)](https://github.com/radixdlt/sargon/pull/299/commits/de17433e6077853e4ebe2a16536423e30043d898)

The difference is the interpretation of "Add any .. until there are exactly 2 factors" - the phrasing "Add any" make it sound like we ought to add *one*. The phrasing "until" make it sound like we **loop**, i.e. add potentially two.

[doc]: https://radixdlt.atlassian.net/wiki/spaces/AT/pages/3758063620/MFA+Rules+for+Factors+and+Security+Shields#Automatic-Security-Shield-Construction

### DIFF
--- a/crates/sargon-uniffi/src/core/types/collections/internal_mapping.rs
+++ b/crates/sargon-uniffi/src/core/types/collections/internal_mapping.rs
@@ -1,13 +1,15 @@
 use crate::prelude::*;
 
-use sargon::IdentifiedVecOf;
+use sargon::{IdentifiedVecOf, IndexSet};
 
-// From InternalType =================================================================================================
-
+// ==========================
+// === From InternalType ====
+// ==========================
 pub trait FromInternal<InternalType, Type> {
     fn into_type(self) -> Type;
 }
 
+// =====  Vec  ======
 impl<InternalElement, Element> FromInternal<Vec<InternalElement>, Vec<Element>>
     for Vec<InternalElement>
 where
@@ -18,6 +20,7 @@ where
     }
 }
 
+// === IdentifiedVec ====
 impl<InternalElement, Element>
     FromInternal<IdentifiedVecOf<InternalElement>, Vec<Element>>
     for IdentifiedVecOf<InternalElement>
@@ -30,12 +33,50 @@ where
     }
 }
 
-// Into InternalType =================================================================================================
+// ====  IndexSet  ======
+impl<InternalElement, Element>
+    FromInternal<IndexSet<InternalElement>, Vec<Element>>
+    for IndexSet<InternalElement>
+where
+    Element: From<InternalElement>,
+{
+    fn into_type(self) -> Vec<Element> {
+        self.into_iter().map(Element::from).collect()
+    }
+}
+
+// ==========================
+// === Into InternalType ====
+// ==========================
 
 pub trait IntoInternal<Type, InternalType> {
     fn into_internal(self) -> InternalType;
 }
 
+// =====    Vec  ========
+impl<InternalElement, Element> IntoInternal<Vec<Element>, Vec<InternalElement>>
+    for Vec<Element>
+where
+    Element: Into<InternalElement>,
+{
+    fn into_internal(self) -> Vec<InternalElement> {
+        self.into_iter().map(Into::into).collect()
+    }
+}
+
+// ====  IndexSet  ======
+impl<InternalElement, Element>
+    IntoInternal<Vec<Element>, IndexSet<InternalElement>> for Vec<Element>
+where
+    Element: Into<InternalElement>,
+    InternalElement: std::hash::Hash + Eq,
+{
+    fn into_internal(self) -> IndexSet<InternalElement> {
+        self.into_iter().map(Into::into).collect::<IndexSet<_>>()
+    }
+}
+
+// === IdentifiedVec ====
 impl<InternalElement, Element>
     IntoInternal<Vec<Element>, IdentifiedVecOf<InternalElement>>
     for Vec<Element>
@@ -44,16 +85,6 @@ where
     Element: Into<InternalElement>,
 {
     fn into_internal(self) -> IdentifiedVecOf<InternalElement> {
-        self.into_iter().map(Into::into).collect()
-    }
-}
-
-impl<InternalElement, Element> IntoInternal<Vec<Element>, Vec<InternalElement>>
-    for Vec<Element>
-where
-    Element: Into<InternalElement>,
-{
-    fn into_internal(self) -> Vec<InternalElement> {
         self.into_iter().map(Into::into).collect()
     }
 }
--- a/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
+++ b/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
@@ -400,7 +400,10 @@ impl SecurityShieldBuilder {
             .map(|x| x.into_internal())
             .collect::<IndexSet<sargon::FactorSource>>();
 
-        self.set_ret(|builder| builder.auto_assign_factors_to_recovery_and_confirmation_based_on_primary(all_factors.clone()))
+        self.set_ret(|builder| {
+            let _ = builder.auto_assign_factors_to_recovery_and_confirmation_based_on_primary(all_factors.clone());
+            Ok(())
+        })
     }
 }
 
--- a/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
+++ b/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
@@ -406,9 +406,9 @@ impl AutoShieldBuilderValidatorOfPickedPrimaryFactors {
     }
 }
 
-impl Into<sargon::ValidatedPrimary> for ValidatedPrimary {
-    fn into(self) -> sargon::ValidatedPrimary {
-        let factors = self
+impl From<ValidatedPrimary> for sargon::ValidatedPrimary {
+    fn from(val: ValidatedPrimary) -> Self {
+        let factors = val
             .secret_magic
             .into_iter()
             .map(|x| x.into())
--- a/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
+++ b/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
@@ -49,6 +49,16 @@ impl SecurityShieldBuilder {
         _ = write(&mut binding);
     }
 
+    fn set_ret<R>(
+        &self,
+        mut write: impl FnMut(
+            &mut sargon::SecurityShieldBuilder,
+        ) -> sargon::Result<R>,
+    ) -> Result<R> {
+        let mut binding = self.wrapped.write().expect("No poison");
+        write(&mut binding).into_result()
+    }
+
     fn validation_for_addition_of_factor_source_by_calling(
         &self,
         factor_sources: Vec<FactorSourceID>,
@@ -377,103 +387,20 @@ impl SecurityShieldBuilder {
     }
 }
 
-#[derive(uniffi::Record)]
-pub struct ValidatedPrimary {
-    secret_magic: Vec<FactorSourceID>,
-}
-
-#[derive(uniffi::Object)]
-pub struct AutoShieldBuilderValidatorOfPickedPrimaryFactors {
-    wrapped: sargon::AutoShieldBuilderValidatorOfPickedPrimaryFactors,
-}
-
-#[uniffi::export]
-impl AutoShieldBuilderValidatorOfPickedPrimaryFactors {
-    pub fn validate_picked(
-        &self,
-        picked: Vec<FactorSourceID>,
-    ) -> Result<ValidatedPrimary> {
-        let valid = self
-            .wrapped
-            .validate_picked(picked.into_iter().map(|x| x.into()).collect())?;
-        Ok(ValidatedPrimary {
-            secret_magic: valid
-                .validated_picked()
-                .into_iter()
-                .map(|x| x.into())
-                .collect(),
-        })
-    }
-}
-
-impl From<ValidatedPrimary> for sargon::ValidatedPrimary {
-    fn from(val: ValidatedPrimary) -> Self {
-        let factors = val
-            .secret_magic
-            .into_iter()
-            .map(|x| x.into())
-            .collect::<IndexSet<sargon::FactorSourceID>>();
-        unsafe { sargon::ValidatedPrimary::new(factors) }
-    }
-}
-
-impl From<sargon::AutoShieldBuilderValidatorOfPickedPrimaryFactors>
-    for AutoShieldBuilderValidatorOfPickedPrimaryFactors
-{
-    fn from(
-        value: sargon::AutoShieldBuilderValidatorOfPickedPrimaryFactors,
-    ) -> Self {
-        Self { wrapped: value }
-    }
-}
-
-// =====================
-// ==== AUTO BUILD =====
-// =====================
-#[uniffi::export(with_foreign)]
-#[async_trait::async_trait]
-pub trait PickFactors: Send + Sync {
-    async fn user_picked_factors(
-        &self,
-        possible: Vec<FactorSource>,
-        validator: Arc<AutoShieldBuilderValidatorOfPickedPrimaryFactors>,
-    ) -> ValidatedPrimary;
-}
-
 use sargon::FactorSource as InternalFactorSource;
 
 #[uniffi::export]
 impl SecurityShieldBuilder {
-    pub async fn auto_build(
+    pub fn auto_assign_factors_to_recovery_and_confirmation_based_on_primary(
         &self,
         all_factors: Vec<FactorSource>,
-        pick_primary_role_factors: Arc<dyn PickFactors>,
-    ) -> Result<SecurityStructureOfFactorSourceIDs> {
+    ) -> Result<()> {
         let all_factors = all_factors
             .into_iter()
             .map(|x| x.into_internal())
             .collect::<IndexSet<sargon::FactorSource>>();
 
-        let shield: sargon::SecurityStructureOfFactorSourceIDs =
-            sargon::AutomaticShieldBuilder::build(
-                all_factors,
-                async |internal_possible: IndexSet<sargon::FactorSource>, internal_validator: sargon::AutoShieldBuilderValidatorOfPickedPrimaryFactors| {
-                    let possible = internal_possible
-                        .into_iter()
-                        .map(|x| crate::FactorSource::from(x.clone()))
-                        .collect_vec();
-
-                    let validator = AutoShieldBuilderValidatorOfPickedPrimaryFactors::from(internal_validator);
-
-                        pick_primary_role_factors
-                            .user_picked_factors(possible, Arc::new(validator))
-                            .await.into()
-                },
-            )
-            .await
-            .into_result()?;
-        let shield = SecurityStructureOfFactorSourceIDs::from(shield);
-        Ok(shield)
+        self.set_ret(|builder| builder.auto_assign_factors_to_recovery_and_confirmation_based_on_primary(all_factors.clone()))
     }
 }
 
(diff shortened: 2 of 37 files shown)


crates/sargon/src/profile/mfa/security_structures/matrices/builder/matrix_builder.rs
```diff
@@ -164,6 +164,14 @@ impl MatrixBuilder {
         Ok(())
     }
 
+    pub fn validate_primary_role_in_isolation(
```
CyonAlexRDX, line 167: @danvleju-rdx do this in a PR, but for each role

CyonAlexRDX: you need to put it in `sargon-uniffi` crate too:

```rust
#[uniffi::export]
impl SecurityShieldBuilder {
   pub fn validate_primary_role_in_isolation(&self) ...
   pub fn validate_recovery_role_in_isolatoin(&self) ...
}
```

or maybe you want:

```rust
#[uniffi::export]
impl SecurityShieldBuilder {
   pub fn validate_role_in_isolation(&self, role: RoleKind) {
        self.get(|builder| {
              match role {
                 RoleKind::Primary => builder.validate_primary_role_in_isolation
             }
       })
   }
}
```
.......

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder.rs
```diff
@@ -0,0 +1,528 @@
...
+        category: FactorSourceCategory,
+        quantity_to_add: Quantity,
+    ) {
+        self.assign_factors_of_category(Confirmation, category, quantity_to_add)
+    }
+
+    /// Automatic assignment of factors to roles according to [this heuristics][doc].
+    ///
+    /// [doc]: https://radixdlt.atlassian.net/wiki/spaces/AT/pages/3758063620/MFA+Rules+for+Factors+and+Security+Shields#Automatic-Security-Shield-Construction
+    fn assign(&mut self) -> Result<ProtoMatrix> {
```
CyonAlexRDX, line 263: @ymatto Would be great if you could take a look at my impl of the "Automatic Shield Construction" here.

We call `assign` on for a SecurityShieldBuilder which has already valid `primary_role threshold factors` set.

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder.rs
```diff
@@ -0,0 +1,528 @@
...
+
+            // 📒 "Add any Hardware factors in the list, starting with the most recently used, to RECOVERY
+            // until there is at least _ factor source in RECOVERY."
+            self.assign_one_hardware_factor(Recovery);
+
+            // 📒 "Add any Hardware (Ledger, Arculus, Yubikey) factors in the list, starting with the most recently used, to CONFIRMATION until there is at least _ factor factors in CONFIRMATION."
+            self.assign_one_hardware_factor(Confirmation);
+        };
+
+        // 📒 "Distribute to try to get at least 1 RECOVERY and then 1 CONFIRMATION" [if possible]
```
CyonAlexRDX, line 298: @ymatto specifically it is this part I want feedback on. Here I try to first add a Hardware factor to `Recovery` and then one to `Confirmation`, and then same again. **Unconditionally**, and it is OK if non was found.

Did you want to make this conditional? As in, only add if number of factors for that role is equal, less than or greater than some threshold? I think this is correct... just that the text in confluence `until there is at least` confused me...

crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -50,6 +51,16 @@ impl SecurityShieldBuilder {
         _ = write(&mut binding);
     }
 
+    fn set_ret<R>(
```
GhenadieVP, line 54: what does `ret` stand for?

crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -378,6 +389,26 @@ impl SecurityShieldBuilder {
...
+#[uniffi::export]
+impl SecurityShieldBuilder {
+    pub fn auto_assign_factors_to_recovery_and_confirmation_based_on_primary(
+        &self,
+        all_factors: Vec<FactorSource>,
+    ) -> Result<()> {
+        let all_factors = all_factors
+            .into_iter()
+            .map(|x| x.into_internal())
+            .collect::<IndexSet<sargon::FactorSource>>();
```
GhenadieVP, line 403: Let's define this in internal_mappings:
```rust
impl<InternalElement, Element> IntoInternal<Vec<Element>, IndexSet<InternalElement>>
    for Vec<Element>
where
    Element: Into<InternalElement>,
    InternalElement: std::hash::Hash + Eq,
{
    fn into_internal(self) -> IndexSet<InternalElement> {
        self.into_iter().map(Into::into).collect::<IndexSet<_>>()
    }
}
```

also the inverse conversion:
```rust
impl<InternalElement, Element> FromInternal<IndexSet<InternalElement>, Vec<Element>>
    for IndexSet<InternalElement>
where
    Element: From<InternalElement>,
{
    fn into_type(self) -> Vec<Element> {
        self.into_iter().map(Element::from).collect()
    }
}
```

crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -378,6 +389,26 @@ impl SecurityShieldBuilder {
...
+
+        self.set_ret(|builder| {
```
GhenadieVP, line 405: Do we need this defined if it is used only once?

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
```diff
@@ -0,0 +1,916 @@
...
+    /// Returns `true` if any factor was assigned, `false` otherwise.
+    fn assign_factors_of_category_to_confirmation(
+        &mut self,
+        category: FactorSourceCategory,
+        quantity_to_add: Quantity,
+    ) -> bool {
+        self.assign_factors_of_category(Confirmation, category, quantity_to_add)
+    }
+
+    fn assign_factor_of_category_to_role_while_meaningful_and_less_than_limit(
```
CyonAlexRDX, line 259: @ymatto I've updated to `loop` adding Factors to `role` of `category` while `cound_factors_for_role(role)` is `< limit` and this is called twice in `assign_custodian_and_hardware_to_role_while_meaningful_and_less_than_limit` below, once for `Custodian` (nothing happens for now, since unsupported) and and once for `Hardware` 

which is then in turn called twice in `assign_custodian_and_hardware_to_non_primary_roles_while_less_than_limit_for_each_assignment` below, first for `Recovery` and then for `Confirmation`, 

which is then in turn called twice in "main function" `assign` below, once with `limit: 1` and then again with `limit: 2`.

I think this is what you want?

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
```diff
@@ -0,0 +1,916 @@
...
+use super::{
+    proto_matrix::ProtoMatrix, quantity::Quantity,
+    CallsToAssignUnsupportedFactor,
+};
+
+use FactorSourceCategory::*;
+use RoleKind::*;
+
+impl FactorSourceCategory {
+    fn is_supported(&self) -> bool {
```
GhenadieVP, line 12: What does this mean exactly? supported as part of automatic build? so we cannot assign Custodian ever as part of automatic?

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
```diff
@@ -0,0 +1,916 @@
...
+                    "Primary role is not valid: {:?}",
+                    invalid_reason
+                ),
+            });
+        }
+
+        if !self.get_primary_override_factors().is_empty() {
+            return Err(CommonError::AutomaticShieldBuildingFailure {
+                underlying: "Primary override factors not allowed when preselecting factors for Recovery and Confirmation".to_string(),
+            });
```
GhenadieVP, line 72: Can you elaborate why there should be no override factors in primary?

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
```diff
@@ -0,0 +1,916 @@
...
+                    .collect::<IndexSet<_>>(),
+            )
+            .cloned()
+            .collect::<IndexSet<FactorSourceID>>()
+            != primary_factors
+        {
+            return Err(CommonError::AutomaticShieldBuildingFailure {
+                underlying: "Primary factors not in profile".to_string(),
+            });
+        }
```
GhenadieVP, line 94: I guess this is good to have, though we should be able to build the primary with factors not in profile in the first place.

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
```diff
@@ -0,0 +1,916 @@
...
+/* Feel free to replace `AutoBuildOutcomeForTesting` return type if you need anything else, I had Unit, so might as well make testing easier by returning this type. */
+    {
+        if let Some(invalid_reason) = self.validate_primary_role() {
+            return Err(CommonError::AutomaticShieldBuildingFailure {
+                underlying: format!(
+                    "Primary role is not valid: {:?}",
+                    invalid_reason
+                ),
+            });
+        }
```
GhenadieVP, line 67: probably a good assertion, though the primary should be already valid when reaching this step.

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
```diff
@@ -0,0 +1,916 @@
...
+
+        if !Self::prerequisites_status(
+            &all_factors_in_profile.iter().map(|f| f.id()).collect(),
+        )
+        .is_sufficient()
+        {
+            return Err(CommonError::AutomaticShieldBuildingFailure {
+                underlying: "Prerequisites not met".to_string(),
+            });
+        }
```
GhenadieVP, line 104: What does this mean more specifically? If user entered the shield build flow, doesn't it mean that they are able to build the shield?

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
```diff
@@ -0,0 +1,916 @@
...
+        if let Some(invalid_reason) = self.validate() {
+            Err(CommonError::AutomaticShieldBuildingFailure {
+                underlying: invalid_reason.to_string(),
+            })
+        } else {
+            Ok(auto_builder.stats_for_testing)
+        }
+    }
+
+    /// Updates the Primary, Recovery and Confirmation roles with the factors of the given `ProtoMatrix`.
```
GhenadieVP, line 129: Why Primary is updated?

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
```diff
@@ -0,0 +1,916 @@
...
+    fn set_state(&self, proto_matrix: ProtoMatrix) {
+        self.reset_factors_in_roles();
+        self.set_threshold(proto_matrix.primary.len() as u8);
+        proto_matrix.primary.into_iter().for_each(|f| {
+            self.add_factor_source_to_primary_threshold(f);
+        });
```
GhenadieVP, line 135: Can you elaborate on this part?

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
```diff
@@ -0,0 +1,916 @@
...
+                        number_of_factors_for_role: self
+                            .count_factors_for_role(role),
+                        limit,
+                    });
+                return;
+            }
+
+            if !self.assign_factors_of_category(role, category, Quantity::One) {
+                // We did not manage to assign any hardware factor, meaning we
+                // it is meaningless to try to assign more factors of this category.
```
GhenadieVP, line 288: Why hardware? Cannot infer this from the code as the category parameter can be any category.

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
```diff
@@ -0,0 +1,916 @@
...
+            );
+        }
+
+        // 📒 Assign Custodian/Hardware factors to RECOVERY & CONFIRMATION
+        // without exceeding limit of 1 factor in each role.
+        self.assign_custodian_and_hardware_to_non_primary_roles_while_less_than_limit_for_each_assignment(1);
+
+        // 📒 Assign Custodian/Hardware factors to RECOVERY & CONFIRMATION
+        // without exceeding limit of 2 factor in each role.
+        self.assign_custodian_and_hardware_to_non_primary_roles_while_less_than_limit_for_each_assignment(2);
```
GhenadieVP, line 362: A bit hard to understand why the second call will not cover the first one also, can you elaborate?

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
```diff
@@ -0,0 +1,916 @@
...
+}
+
+impl SecurityShieldBuilder {
+    /// Returns the status of the prerequisites for building a Security Shield.
+    ///
+    /// According to [definition][doc], a Security Shield can be built if the user has, asides from
+    /// the Identity factor, "2 or more factors, one of which must be Hardware"
+    ///
+    /// [doc]: https://radixdlt.atlassian.net/wiki/spaces/AT/pages/3758063620/MFA+Rules+for+Factors+and+Security+Shields#Factor-Prerequisites
+    pub fn prerequisites_status(
```
GhenadieVP, line 401: This is moved, right?

crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
CyonAlexRDX: `return`... I'll update it

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
CyonAlexRDX: All of Matts SecurityShield building uses Threshold list for Primary.

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
CyonAlexRDX: I like asserts, as you know :) always helpful!

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
CyonAlexRDX: not if we do programmer error in host.... this protects us from any Radix developer calling this in Hosts without having set the factors on Primary.

GhenadieVP (approved):

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
CyonAlexRDX: `ProtoMatrix` is just a named tuple of the three lists of factors, for the three roles.

The `auto_assign_factors_to_recovery_and_confirmation_based_on_primary` either is fully successful, or fails - without any mutation. So the assignment of the factors to role need to happen NOT to `self` (`SecurityShieldBuilder`) so they happen to `ProtoMatrix`.

Once assignment is successful `SecurityShieldBuilder` instance calls `set_state` to set the factors.

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
CyonAlexRDX: let me know if you have concrete suggestions, now that we have walked this through in a call.

crates/sargon/src/profile/mfa/security_structures/automatic_shield_builder/automatic_shield_builder.rs
CyonAlexRDX: yes just moved

crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
CyonAlexRDX: @GhenadieVP thx! I've added it here, **and made use of it in other files too**, very nice:

https://github.com/radixdlt/sargon/pull/299/commits/d4335a14a3e4f31781db1fe6f9f10e09178561dd

We should add one for `IndexMap` too

crates/sargon-uniffi/src/core/types/collections/internal_mapping.rs
```diff
@@ -30,30 +33,58 @@ where
...
+// === IdentifiedVec ====
+impl<InternalElement, Element>
+    IntoInternal<Vec<Element>, IdentifiedVecOf<InternalElement>>
     for Vec<Element>
 where
+    InternalElement: Debug + PartialEq + Eq + Clone + sargon::Identifiable,
     Element: Into<InternalElement>,
 {
-    fn into_internal(self) -> Vec<InternalElement> {
+    fn into_internal(self) -> IdentifiedVecOf<InternalElement> {
```
CyonAlexRDX, line 87: unfortunate git diff by Github. I moved around the methods a bit....

crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
GhenadieVP: it should be possible to achieve through some trait trickery, but it is a separate task on its own :), aka converting automatically from IndexMap<Key,Value> to some struct X { key, value }

crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
GhenadieVP: no its fine, just asked :)

danvleju-rdx (approved):

sergiupuhalschi-rdx (approved):

PR #302, opened 2024-12-11
Entities linked to Factor Source

matiasbzurovski:
## Changelog
Exposes a function on `SargonOS` to allow checking entities related to a `FactorSource`. 

This allow us to determine, for example, which entities were created under a given `DeviceFactorSource` or a `LedgerFactorSource`. Also, it will work for Securified Entities to see if the given factor source is in any way involved in the matrix that securifies it.

## Output
Asides from returning the corresponding [entities and their visibility](https://github.com/radixdlt/sargon/pull/302/files#diff-9abe56d56197a9e444ecbf54de9ba0c9305300788e044c1a3358959fa647df11R11-R20), the output includes also an `integrity` field which will become helpful for host apps to determine the integrity of the mentioned `FactorSource`.

In the case of `DeviceFactorSource`, its [integrity](https://github.com/radixdlt/sargon/pull/302/files#diff-41dfa590587f4c38b2819ed9f559fcc96a3d839c6ee7aedb76b59ad012ff2cffR6) will detail whether its mnemonic is present in keychain, and if it has been backed up by the user.

## Storage clients 
To check if the mnemonic is present in Keychain, I make use of the `SecureStorageClient` using the previously supported key `DeviceFactorSourceMnemonic`.

However, in order to check if the mnemonic is backed up I need to make usage of the `UnsafeStorageClient`. In iOS, when booting SargonOS, we implement the `UnsafeStorageDriver` directly from native `UserDefaults`. In other words, every time Sargon needs to read/write some values on unsafe storage, it will directly use the `UserDefaults`.

Then, the only pending thing I needed to add was a way to map the keys previously defined by iOS Wallet, into the ones expected by Sargon. Right now we only have one `UnsafeStorageKey` which is `FactorSourceUserHasWrittenDown`, which I map on [iOS](https://github.com/radixdlt/babylon-wallet-ios/blob/ABW-3920-device-factor-sources-list/RadixWallet/Clients/UserDefaults%2BDependency%2BExtension/UserDefaults%2BDependency%2BExtension.swift#L27-L33) into its pre-Sargon-defined-key named `"mnemonicsUserClaimsToHaveBackedUp"`.

> [!IMPORTANT]
> This approach only works for both platforms if someone from Android team can validate that:
> - they are also saving in Unsafe Storage a set of `FactorSourceIDFromHash` of device factor sources whose mnemonic has been written down.
> - they are either using the SargonOS key, or they can also map it to the expected ones as iOS does.

If this is not the case, then we may not be able to have the straightforward solution of interacting directly with `UserDefaults` on iOS, and instead we would need to treat unsafe storage similarly to secure storage: just send the key to the host and let it read/write the value however it decides to.


crates/sargon/src/profile/mfa/secured_entity_control/secured_entity_control.rs
```diff
@@ -65,6 +65,23 @@ impl SecuredEntityControl {
...
+        factor_source: FactorSource,
+    ) -> bool {
+        self.security_structure
+            .matrix_of_factors
+            .primary_role
+            .get_threshold_factors()
+            .iter()
+            .any(|factor| {
+                factor.factor_source_id == factor_source.factor_source_id()
+            })
```
matiasbzurovski, line 81: Note: initial implementation, will be revisited

crates/sargon/src/system/drivers/unsafe_storage_driver/support/unsafe_storage_key.rs
```diff
@@ -7,25 +7,30 @@ pub enum UnsafeStorageKey {
...
+        // format!(
+        //     "unsafe_storage_key_{}",
+        //     match self {
+        //         UnsafeStorageKey::FactorSourceUserHasWrittenDown =>
+        //             "factor_source_user_has_written_down".to_owned(),
+        //     }
+        // )
+        match self {
+            UnsafeStorageKey::FactorSourceUserHasWrittenDown => {
+                "mnemonicsUserClaimsToHaveBackedUp".to_owned()
```
matiasbzurovski, line 19: Note: working on solution for this

crates/sargon-uniffi/src/system/sargon_os/entities_controlled_by_factor_source/support/entities_controlled_by_factor_source.rs
```diff
@@ -0,0 +1,27 @@
+use crate::prelude::*;
+use sargon::EntitiesControlledByFactorSource as InternalEntitiesControlledByFactorSource;
+
+/// This is the result of checking what entities are controlled by a given `FactorSource`.
+#[derive(Clone, PartialEq, Eq, InternalConversion, uniffi::Record)]
+pub struct EntitiesControlledByFactorSource {
+    /// The factor source that controls the entities.
```
Sajjon, line 7: `/// The factor source that is used by the entities` ?

crates/sargon-uniffi/src/system/sargon_os/entities_controlled_by_factor_source/support/entities_controlled_by_factor_source.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub factor_source: FactorSource,
+
+    /// Whether the mnemonic of the factor source is present in keychain.
+    pub is_mnemonic_present_in_keychain: bool,
```
Sajjon, line 11: many of the fields here are `DeviceFactorSource` specific. If the WHOLE query is that, we should name it `EntitiesUsingDeviceFactorSource`? Or we should generalize it, probably with enums with associated data, so that `is_mnemonic_present_in_keychain: bool` only exists when the factor source is of kind `device`

crates/sargon-uniffi/src/system/sargon_os/entities_controlled_by_factor_source/sargon_os_entities_controlled_by_factor_source.rs
```diff
@@ -0,0 +1,19 @@
+use crate::prelude::*;
+
+#[uniffi::export]
+impl SargonOS {
+    /// Returns the entities controlled by a given `FactorSource`, either on the current `Profile` or a specific one.
+    pub async fn entities_controlled_by_factor_source(
```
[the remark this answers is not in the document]
matiasbzurovski: 100% that `controlled` is not the correct word now we have MFA. Not sure if `using` is the appropriate one though.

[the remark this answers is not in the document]
Sajjon: "Security state References"?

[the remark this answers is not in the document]
Sajjon: `factor_sources_with_referenced_entities` as name of function

And type
`FactorSourceWithReferencedEntities`?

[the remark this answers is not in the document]
matiasbzurovski: my last attempt: `entities_linked_to_factor_source` & `EntitiesLinkedToFactorSource` ?

Not that I think we should name things based on UI, but this is how we call it on UI and I like it better than `referenced`

crates/sargon-uniffi/src/system/drivers/secure_storage_driver/secure_storage_driver.rs
```diff
@@ -19,6 +19,11 @@ pub trait SecureStorageDriver: Send + Sync + std::fmt::Debug {
     ) -> Result<()>;
 
     async fn delete_data_for_key(&self, key: SecureStorageKey) -> Result<()>;
+
+    async fn contains_data_for_key(
```
Sajjon, line 23: maybe a *batch* query would make sense? `contains_data_for_keys` ? Check with @micbakos-rdx but on iOS we can load the entire contents of Keychain in one go without prompting for biometrics and check the list of key-values. So we can actually impl `check_existence_of_mnemonics_for_device_factor_sources(ids: IndexSet<FactorSourceIDFromHash>) -> IndexSet<FactorSourceIDFromHash>` where if return value == ids.len() then **all** was found.

crates/sargon-uniffi/src/system/sargon_os/entities_controlled_by_factor_source/support/accessibility/device.rs
```diff
@@ -0,0 +1,15 @@
+use crate::prelude::*;
+use sargon::DeviceFactorSourceAccessibility as InternalDeviceFactorSourceAccessibility;
+
+/// A struct representing the accessibility of a device factor source.
```
Sajjon, line 4: I think `accessibility` is the wrong term. I think you mean `existence` ? Accessibility make it sound like [keychain *accessibility*](https://developer.apple.com/documentation/security/restricting-keychain-item-accessibility), which is a specific keychain **configuration**.

Sajjon: Right we have `is_mnemonic_marked_as_backed_up`, so what about: `ClaimedAndActualExistenceOfDeviceFactorSource`

crates/sargon-uniffi/src/system/drivers/secure_storage_driver/secure_storage_driver.rs
matiasbzurovski: for the current implementation we will only make the request for a given Factor Source, but I can extend it to take a set of keys instead.

crates/sargon/src/profile/mfa/secured_entity_control/secured_entity_control.rs
```diff
@@ -65,6 +65,23 @@ impl SecuredEntityControl {
     }
 }
 
+impl SecuredEntityControl {
+    /// Returns whether the entity is linked to the given factor source.
+    pub fn is_linked_to_factor_source(
```
Sajjon, line 70: Ah! You can actually `impl HasFactorInstances for EntitySecurityState` and then you can let `is_linked_to_factor_source` be a trait method on `HasFactorInstances` with **default** implementation, so in the body you will call `self.unique_factor_instances()` and then `.iter().any(...)`.

crates/sargon-uniffi/src/system/sargon_os/entities_controlled_by_factor_source/support/accessibility/device.rs
matiasbzurovski: I didn't name based on keychain accesibility, but based on the definition of the word itself:
> the quality of being able to be reached or entered.

crates/sargon/src/system/sargon_os/entities_linked_to_factor_source/sargon_os_entities_linked_to_factor_source.rs
```diff
@@ -0,0 +1,133 @@
...
+        EntitiesLinkedToFactorSource {
+            accessibility,
+            accounts,
+            hidden_accounts,
+            personas,
+            hidden_personas,
+        }
+    }
+
+    async fn accessibility(
```
Sajjon, line 99: `claimed_and_actual_existence_of_factor_source`

crates/sargon/src/system/sargon_os/entities_linked_to_factor_source/sargon_os_entities_linked_to_factor_source.rs
```diff
@@ -0,0 +1,133 @@
+use crate::prelude::*;
+
+impl SargonOS {
+    /// Returns the entities linked to a given `FactorSource`, either on the current `Profile` or a specific one.
+    pub async fn entities_linked_to_factor_source(
+        &self,
+        factor_source: FactorSource,
+        profile_to_check: ProfileToCheck,
+    ) -> Result<EntitiesLinkedToFactorSource> {
+        let accessibility = self.accessibility(factor_source.clone()).await?;
```
Sajjon, line 10: Same review comment as in some PRs ago... :) this should not be on `SargonOS` :) `SargonOS` ought to only "forward" calls to `Profile` (via `ProfileHolder`.). since you here support `ProfileToCheck::Specific` so maybe something like?:


```rust
impl SargonOS {
    /// Returns the entities linked to a given `FactorSource`, either on the current `Profile` or a specific one.
    pub async fn entities_linked_to_factor_source(
        &self,
        factor_source: FactorSource,
        profile_to_check: ProfileToCheck,
    ) -> Result<EntitiesLinkedToFactorSource> {

        let existence = self.claimed_and_actual_existence_of_factor_source(factor_source.clone()).await?;

        match profile_to_check {
            ProfileToCheck::Specific(profile) => profile.entities_linked_to_factor_source(factor_source, existence),
            ProfileToCheck::Current => self.profile().entities_linked_to_factor_source(factor_source, existence),
        }
    }
}
```

crates/sargon-uniffi/src/system/sargon_os/entities_controlled_by_factor_source/support/accessibility/device.rs
Sajjon: Yeah I understood that :) my point is it is a loaded term, especially since we are touching something very Keychain related

crates/sargon-uniffi/src/system/drivers/secure_storage_driver/secure_storage_driver.rs
GhenadieVP: @micbakos-rdx mentioned that we can check for key existence without any biometrics.

crates/sargon/src/profile/logic/profile_network/profile_network_entities_linked_to_factor_source.rs
```diff
@@ -0,0 +1,111 @@
...
+            .accounts_non_hidden()
+            .iter()
+            .filter(|a| {
+                a.security_state
+                    .is_linked_to_factor_source(factor_source.clone())
+            })
+            .collect::<Accounts>();
+        let hidden_accounts = self
+            .accounts
+            .hidden()
```
GhenadieVP, line 20: Minor: I guess we should have similar function to `accounts_non_hidden` on ProfileNetwork - `accounts_hidden`

crates/sargon/src/profile/logic/profile_network/profile_network_entities_linked_to_factor_source.rs
```diff
@@ -0,0 +1,111 @@
...
+        &self,
+        factor_source: FactorSource,
+        integrity: FactorSourceIntegrity,
+    ) -> Result<EntitiesLinkedToFactorSource> {
+        let accounts = self
+            .accounts_non_hidden()
+            .iter()
+            .filter(|a| {
+                a.security_state
+                    .is_linked_to_factor_source(factor_source.clone())
```
GhenadieVP, line 15: This could be in a common function, then passed to filter.

GhenadieVP (approved):
Looks good overall, very nice improvement.

crates/sargon/src/profile/logic/profile_network/profile_network_entities_linked_to_factor_source.rs
matiasbzurovski: that what was my original idea: to define the `filter` predicate once and then use it on every collection:
```Rust
let filter = |e: &dyn HasSecurityState| {
    e.security_state().is_linked_to_factor_source(factor_source.clone())
};
```

I wasn't able to make it work with a closure, but using a function that sends the `factor_source` each time works, thanks!

crates/sargon/src/system/drivers/secure_storage_driver/support/test/mock_secure_storage.rs
```diff
@@ -0,0 +1,52 @@
+#![cfg(test)]
+
+use crate::prelude::*;
+
+#[derive(Debug)]
+pub struct MockSecureStorage {
+    mock_contains_data_for_key: IndexMap<SecureStorageKey, bool>,
```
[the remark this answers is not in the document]
matiasbzurovski: deleted

giannis-rdx:
> This approach only works for both platforms if someone from Android team can validate that:
> - they are also saving in Unsafe Storage a set of `FactorSourceIDFromHash` of device factor sources whose mnemonic has been written down.
> - they are either using the SargonOS key, or they can also map it to the expected ones as iOS does.

[yes for both](https://github.com/radixdlt/babylon-wallet-android/blob/25f817e99815a987d5ffdf9f6aa7a74045e017d9/core/src/main/java/rdx/works/core/preferences/PreferencesManager.kt#L215).

crates/sargon/src/system/sargon_os/sargon_os_entities_linked_to_factor_source.rs
```diff
@@ -0,0 +1,317 @@
...
+
+impl SargonOS {
+    /// Returns the entities linked to a given `FactorSource`, either on the current `Profile` or a specific one.
+    pub async fn entities_linked_to_factor_source(
+        &self,
+        factor_source: FactorSource,
+        profile_to_check: ProfileToCheck,
+    ) -> Result<EntitiesLinkedToFactorSource> {
+        let integrity = self.integrity(factor_source.clone()).await?;
+        match profile_to_check {
```
Sajjon, line 11: prefer expressions:

```rust
        let profile = match profile_to_check {
            ProfileToCheck::Current => self.profile()?
            ProfileToCheck::Specific(specific_profile) => specific_profile
        };

profile
                .current_network()?
                .entities_linked_to_factor_source(factor_source, integrity),
```

crates/sargon-uniffi/src/profile/v100/entities_linked_to_factor_source/integrity/device.rs
```diff
@@ -0,0 +1,15 @@
...
+pub struct DeviceFactorSourceIntegrity {
+    /// The factor source that is linked to the entities.
+    pub factor_source: DeviceFactorSource,
+
+    /// Whether the mnemonic of the factor source is present in keychain.
+    pub is_mnemonic_present_in_keychain: bool,
+
+    /// Whether the mnemonic of the factor source is marked as backed up.
+    pub is_mnemonic_marked_as_backed_up: bool,
+}
```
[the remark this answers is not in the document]
giannis-rdx: why the name has to be related to iOS though?
shouldn't be `is_mnemonic_present` or something else?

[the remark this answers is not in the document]
giannis-rdx: we have similar thing in Android [mnemonicExist](https://github.com/radixdlt/babylon-wallet-android/blob/main/profile/src/main/java/rdx/works/profile/data/repository/MnemonicRepository.kt)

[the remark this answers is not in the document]
Sajjon: it should probably be `_in_secure_storage`. Not "keychain", good catch @giannis-rdx 

And @matiasbzurovski we can have a computed property in Swift Sargon called ...inKeychain, to make it clear

crates/sargon/src/system/sargon_os/sargon_os_entities_linked_to_factor_source.rs
matiasbzurovski: this doesn't compile because of multiple errors about borrowed values not living long enough, and in order to fix it I need to add more variables which I think complicates the code more.

crates/sargon-uniffi/src/profile/v100/entities_linked_to_factor_source/integrity/device.rs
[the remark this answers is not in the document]
matiasbzurovski: Updated it to replace every reference to `_in_keychain` to `_in_secure_stroage`.

giannis-rdx (approved):

PR #303, opened 2024-12-12
Keep HostId in memory

GhenadieVP:
This is part of fixing the issue with user seeing Claim Profile prompt on iOS.

The root cause of the issue is that the iOS Wallet fails to resolve the host id and [then we do consider that profile needs to be claimed](https://github.com/radixdlt/babylon-wallet-ios/blob/c209586b2753539384be656f291a4cafb0be8bce/RadixWallet/Clients/ProfileStore/ProfileStore.swift#L66). While this might be wrong assumption, just changing that to true is not the most correct fix.

The iOS wallet will fail to resolve the host id when the app is in background, as the access to the keychain is not allowed while the app is in background. We will try to resolve the host id every 15 minutes, which is the interval we do check for profile claim.
So this bug happens in a very specific scenario - when user moves the app to background somewhere at 13-14 minute - and when at 15 minute mark the wallet tries to resolve the host id, the following error is thrown by the secure storage:

> OSStatus error:[-25308] User interaction is not allowed.

The fix, is to resolve the host id at boot time and then just read it from memory throughout the app lifecycle. Additionally, failing to resolve the host id is not considered a breaking error - even more, it is highly unlikely that the app will fail to perform secure storage operations at boot time.

### DIFF
--- a/crates/sargon-uniffi/src/system/sargon_os/sargon_os.rs
+++ b/crates/sargon-uniffi/src/system/sargon_os/sargon_os.rs
@@ -75,8 +75,8 @@ impl SargonOS {
         self.wrapped.delete_wallet().await.into_result()
     }
 
-    pub async fn resolve_host_id(&self) -> Result<HostId> {
-        self.wrapped.resolve_host_id().await.into_result()
+    pub fn resolve_host_id(&self) -> HostId {
+        self.wrapped.host_id().into()
     }
 
     pub async fn resolve_host_info(&self) -> HostInfo {
--- a/crates/sargon/src/system/sargon_os/sargon_os.rs
+++ b/crates/sargon/src/system/sargon_os/sargon_os.rs
@@ -345,12 +345,13 @@ impl SargonOS {
             Ok(Some(loaded_host_id)) => {
                 debug!("Found saved host id: {:?}", &loaded_host_id);
                 loaded_host_id
-            },
+            }
             Ok(None) => {
                 debug!("Found no saved host id, creating new.");
                 let new_host_id = HostId::generate_new();
                 debug!("Created new host id: {:?}", &new_host_id);
-                let save_result = secure_storage.save_host_id(&new_host_id).await;
+                let save_result =
+                    secure_storage.save_host_id(&new_host_id).await;
                 if let Err(error) = save_result {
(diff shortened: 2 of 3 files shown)


crates/sargon/src/system/sargon_os/sargon_os.rs
```diff
@@ -334,26 +335,33 @@ impl SargonOS {
...
-    pub(crate) async fn host_id(&self) -> Result<HostId> {
-        Self::get_host_id(&self.clients).await
-    }
-
-    pub(crate) async fn get_host_id(clients: &Clients) -> Result<HostId> {
+    pub(crate) async fn get_host_id(clients: &Clients) -> HostId {
         debug!("Get Host ID");
         let secure_storage = &clients.secure_storage;
+        let stored_host_id = secure_storage.load_host_id().await;
+        let new_host_id = HostId::generate_new();
```
Sajjon, line 342: you are calling `HostId::generate_new();` twice, once here, and then once in `Ok(None)` branch. I think best remove this one and put one in the `Err` branch, which means that we wont call `HostId::generate_new();` at all for the `Ok(Some(_))` branch, which is good. because if we e.g. add logging inside `HostId::generate_new();`, it might confuse us to see a new Id being generated even though `match stored_host_id` was `Ok(Some(_))`

CyonAlexRDX (approved):
Excellent job tracking this one down! 👏

crates/sargon-uniffi/src/system/sargon_os/sargon_os.rs
```diff
@@ -75,8 +75,8 @@ impl SargonOS {
         self.wrapped.delete_wallet().await.into_result()
     }
 
-    pub async fn resolve_host_id(&self) -> Result<HostId> {
-        self.wrapped.resolve_host_id().await.into_result()
+    pub fn resolve_host_id(&self) -> HostId {
+        self.wrapped.host_id().into()
     }
 
     pub async fn resolve_host_info(&self) -> HostInfo {
```
giannis-rdx, line 82: should we do the same for `host_info` or this is safe?

GhenadieVP: safe

PR #305, opened 2024-12-16
Allow setting name of FactorSources

matiasbzurovski:
Exports a UniFFI function that allows setting name of any FactorSource.

### DIFF
--- a/crates/sargon-uniffi/src/profile/v100/factors/factor_source.rs
+++ b/crates/sargon-uniffi/src/profile/v100/factors/factor_source.rs
@@ -59,6 +59,16 @@ pub fn factor_source_name(factor_source: &FactorSource) -> String {
 }
 
 #[uniffi::export]
+pub fn factor_source_set_name(
+    factor_source: FactorSource,
+    updated: String,
+) -> FactorSource {
+    let mut factor_source = factor_source.into_internal();
+    factor_source.set_name(updated);
+    factor_source.into()
+}
+
+#[uniffi::export]
 pub fn new_factor_source_sample() -> FactorSource {
     InternalFactorSource::sample().into()
 }
--- a/crates/sargon/src/profile/mfa/mfa_factor_sources/off_device_mnemonic_factor_source/off_device_mnemonic_factor_source.rs
+++ b/crates/sargon/src/profile/mfa/mfa_factor_sources/off_device_mnemonic_factor_source/off_device_mnemonic_factor_source.rs
@@ -112,6 +112,10 @@ impl BaseBaseIsFactorSource for OffDeviceMnemonicFactorSource {
     fn name(&self) -> String {
         self.hint.label.value.clone()
     }
+
+    fn set_name(&mut self, updated: String) {
+        self.hint.label.value = updated;
+    }
 }
 
 #[cfg(test)]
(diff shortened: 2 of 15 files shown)


crates/sargon/src/profile/mfa/mfa_factor_sources/off_device_mnemonic_factor_source/off_device_mnemonic_factor_source.rs
```diff
@@ -112,6 +112,10 @@ impl BaseBaseIsFactorSource for OffDeviceMnemonicFactorSource {
     fn name(&self) -> String {
         self.hint.label.value.clone()
     }
+
+    fn set_name(&mut self, updated: String) {
```
Sajjon, line 116: minor improvement would be to change `String` -> `impl AsRef<str>` making it a bit easier to write unit tests, allowing you to pass string literals (&str) or `String` :)

Sajjon: and inside this function you do `updated.as_ref().to_owned()`

Sajjon (commented):
LGTM - will approve on work phone tomorrow. anyone on work phone now can approve it for me :)

CyonAlexRDX (approved):

crates/sargon-uniffi/src/profile/v100/factors/factor_source.rs
```diff
@@ -58,6 +58,16 @@ pub fn factor_source_name(factor_source: &FactorSource) -> String {
     factor_source.into_internal().name()
 }
 
+#[uniffi::export]
+pub fn factor_source_set_name(
```
GhenadieVP, line 62: This is nice, but I think we can go a step further and also make the update in the Profile itself, so maybe this API should have the factor source id to be updated.

Or your intention is different?

matiasbzurovski: I was thinking of handling it the same way we handle Account name updates, but guess we can directly do altogether. Will update

GhenadieVP: That is more legacy :), we should migrate how we rename accounts sometime to be handled in Sargon.

matiasbzurovski: actually I am not sure if it is better to have a function that updates the `name: String` of a given `factor_source_id: FactorSourceId`. Considering Hosts we will already have access to the actual `FactorSource` and that we already have a [function](https://github.com/radixdlt/sargon/blob/main/crates/sargon/src/system/sargon_os/sargon_os_factors.rs#L32) that updates a `FactorSource`, seems pointless to send the `id` just to have Sargon fetch it from the Profile.

Maybe instead the function should be like this
```Rust
pub async fn update_factor_source_name(&self, factor_source: FactorSource, name: String) -> Result<()>
```

GhenadieVP: Sure that also works, the main idea is to have the Profile mutation in Sargon. You might want to return the update FactorSource as response of this function.

GhenadieVP (approved):

PR #306, opened 2024-12-17
FIP: Support mix of Accounts & Personas

CyonAlexRDX:
`FactorInstancesProvider` - FIP - does not support providing instances for a mix of DerivationPresets - meaning we cannot ask it for instance for accounts and personas mixed. 

Today one would need to do "two passes" to the FIP => bad UX since if the cache could not fully satisfied any requests => two passes to the KeysCollector (once per pass to the FIP). 

What we want is a single pass to the KeysCollector in case of not enough instances in cache. 

Thus what we need is to support multiple derivation presets queried. 

# Changes
- FIP now supports providing instances for a mix of accounts and personas
- FIP now derives ROLA keys for accounts and personas for new Factors and for securifying accounts If cache could not satisfy request. Also upgraded the `NextDerivationEntityIndexProfileAnalysingAssigner` to analyse profile and return next derivation index for
ROLA factors.
- SecurityStructureOfFactor{Source/SourceIds} now contain an `authentication_signing_factor` field 
- SecurityShieldBuilders (Sargon and Sargon-UniFFI) now exposes methods for setting Auth Signing factor
- Auto Shield Builder now supports ROLA key: auto assign a `Device` FactorSource as `authentication_signing` factor.
- Split single const `CACHE_FILLING_QUANTITY` into one const per preset (they still have the same value: 30)

> [!IMPORTANT]
> This PR does NOT validate the FactorSourceKind used for Authentication Signing Factor. So the SecurityShieldBuilder need to validate that. A tiny PR by anyone fixing this post getting this merged is welcome! See `owner` row in table https://radixdlt.atlassian.net/wiki/spaces/AT/pages/3758063620/MFA+Rules+for+Factors+and+Security+Shields#Factor-Type-Usage-Rules


crates/sargon/src/factor_instances_provider/provider/outcome/internal_factor_instances_provider_outcome.rs
```diff
@@ -2,13 +2,21 @@ use crate::prelude::*;
 
 #[derive(Clone, Debug)]
 pub struct InternalFactorInstancesProviderOutcome {
+    pub per_derivation_preset: IndexMap<
```
Sajjon, line 5: Hmm... should it be:
* PER FactorSource PER DerivationPreset or
* PER DerivationPreset PER FactorSource
?
I think the latter is better, because it would allow to query the FIP with different FactorSources per DerivationPreset, which would allow e.g. `ShieldA` to be used for personas `X` & `Y` but `ShieldB` (where `ShieldA != ShieldB`) for account `A` & `B` **to be queried for in the same single pass to FIP**. So that is one argument in favour of `PER DerivationPreset PER FactorSource`. Are there any arguments in favour of `PER FactorSource PER DerivationPreset` or?

@matiasbzurovski @GhenadieVP I would appriciated any early input here!

Sajjon: (I've begun upgrading the `FactorInstancesCache` first, which I think is near complete)

The `InternalFactorInstancesProviderOutcome` is ofc used by `FactorInstancesProviderOutcome` (which I have not updated yet).

So how are there outcomes used? Well they are used by `VirtualEntityCreatingInstanceProvider` and then in turn by `Profile` method [`create_unsaved_entities_with_factor_source_with_derivation_outcome `](https://github.com/radixdlt/sargon/blob/7f1368319f29e93024635d7572f9fd6133e2cad2/crates/sargon/src/profile/logic/create_entity.rs#L35-L67) which will essentially remain the same, since it will NOT query the FIP (via the`VirtualEntityCreatingInstanceProvider` ) with multiple DerivationPresets, but rather just the one (either `DerivationPreset::AccountVeci` or `DerivationPresets::IdentityVeci`).

So lets look at the use case where we will query the FIP with multiple derivation presets:

Sajjon: **Gated behind `cfg(test)`: `make_security_structure_of_factor_instances_for_entities_without_consuming_cache_with_derivation_outcome`**
which you find here:
https://github.com/radixdlt/sargon/blob/7f1368319f29e93024635d7572f9fd6133e2cad2/crates/sargon/src/system/sargon_os/sargon_os_accounts.rs#L800-L837

It is, as mentioned above gated behind `cfg(test)` so not at all final... and we note that the signature is now:

```rust
    pub(crate) async fn make_security_structure_of_factor_instances_for_entities_without_consuming_cache_with_derivation_outcome<
        A: IsEntityAddress,
    >(
        &self,
        addresses_of_entities: IndexSet<A>,
        security_structure_of_factor_sources: SecurityStructureOfFactorSources, // Aka "shield"
    ) -> Result<(
```

**But this PR WILL change the signature to NOT be generic over `E`**, so `addresses_of_entities: IndexSet<A>` is going to be changed to `IndexSet<AddressOfAccountOrPersona>`.

[So instead of](https://github.com/radixdlt/sargon/blob/7f1368319f29e93024635d7572f9fd6133e2cad2/crates/sargon/src/system/sargon_os/sargon_os_accounts.rs#L852-L854):
```rust
addresses_of_entities.clone().into_iter().map(|entity_address|
```

we probably wanna split the `addresses_of_entities` into Persona addresses and query `outcome.get(DerivationPreset::IdentityMfa)` and then iterate... and then the same for accounts?

Sajjon: So the more I write about this the more convinced I become that we want `PER DerivationPreset PER FactorSource`. or rather NEED.

GhenadieVP: Yes `PER DerivationPreset PER FactorSource` makes more sense, as in different contexts we need to use the different preset, and it is nice to be able to slice all of the factor instances per preset.

Sajjon: This is new - introduce by this PR:
```rust
#[derive(Debug, PartialEq, Eq)]
pub struct CacheInstancesAndRemainingQuantityToDerive {
    pub instances_to_use_from_cache: FactorInstances, // if empty then this was not a requested derivation preset, but we are cache filling and found `quantity_to_derive` needed to fill cache.
    pub quantity_to_derive: usize,
}
```

So I should probably change this - which is also new in this PR:
```diff
#[derive(Debug, PartialEq, Eq)]
pub struct CacheNotSatisfied { // NEW introduced in this PR
-   /// PER FactorSourceID => PER DerivationPreset => CacheInstancesAndRemainingQuantityToDerive
+   /// PER DerivationPreset => PER FactorSourceID => CacheInstancesAndRemainingQuantityToDerive
    pub cached_and_quantities_to_derive: IndexMap<
-        FactorSourceIDFromHash,
+       DerivationPreset
        IndexMap<
-             DerivationPreset, 
+             FactorSourceIDFromHash, 
             CacheInstancesAndRemainingQuantityToDerive
        >,
    >,
}
```

GhenadieVP: At the cache level we do store per factor source id, right?

Sajjon: Yes the cache is:

```rust
pub(super) type FICStorage = IndexMap<
    FactorSourceIDFromHash,
    IndexMap<IndexAgnosticPath, FactorInstances>,
>;
```

But I think that makes sense to keep as is, since, it allows for easy add/delete instances based on new FactorSource.

Sajjon: So I think it makes sense to keep that, but ""transposing"" it to:

```rust
#[derive(Debug, PartialEq, Eq)]
pub struct CacheNotSatisfied {
    /// PER DerivationPreset => PER FactorSourceID => CacheInstancesAndRemainingQuantityToDerive
    pub cached_and_quantities_to_derive: IndexMap<
        DerivationPreset,
        IndexMap<
            FactorSourceIDFromHash,
            CacheInstancesAndRemainingQuantityToDerive,
        >,
    >,
}
```

which then quite easily and naturally maps to;

```rust
#[derive(Clone, Debug)]
pub struct InternalFactorInstancesProviderOutcome {
    pub per_derivation_preset: IndexMap<
        DerivationPreset,
        InternalFactorInstancesProviderOutcomePerFactor,
    >,
}
```

which will then be quite easy to use in `make_security_structure_of_factor_instances_for_entities_without_consuming_cache_with_derivation_outcome` which will "group" entities of kind Persona and entities of kind Account together and use the `DerivationPreset::AccountMfa` and `DerivationPreset::IdentityMfa` respectively as "key" to get their `InternalFactorInstancesProviderOutcomePerFactor`, from the outcome of the FIP.

I think that will turn out quite nicely

crates/sargon/src/factor_instances_provider/provider/factor_instances_provider.rs
```diff
@@ -68,12 +88,9 @@ impl FactorInstancesProvider {
 impl FactorInstancesProvider {
     fn make_instances_in_cache_consumer(
         &self,
-        instances_per_factor_sources_to_delete: IndexMap<
-            FactorSourceIDFromHash,
-            FactorInstances,
-        >,
+        instances_to_delete: CachedInstancesToUse,
```
Sajjon, line 91: Change type
To less confusing type alias

crates/sargon/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(warnings)]
```
CyonAlexRDX, line 1: ‼️❌‼️⛔️ REVERT THIS!!!!!!

crates/sargon/src/factor_instances_provider/agnostic_paths/quantified_derivation_preset.rs
```diff
@@ -7,11 +7,61 @@ pub struct QuantifiedDerivationPreset {
...
+
 impl QuantifiedDerivationPreset {
     pub fn new(derivation_preset: DerivationPreset, quantity: usize) -> Self {
         Self {
             derivation_preset,
             quantity,
         }
     }
+
+    pub fn mfa_for_entities(
```
Sajjon, line 25: Doc

crates/sargon/src/factor_instances_provider/provider/outcome/factor_instances_provider_outcome.rs
```diff
@@ -4,6 +4,30 @@ use crate::prelude::*;
 /// renamed field values to make it clear that `to_cache` instances  already have been cached.
 #[derive(Clone, Debug)]
 pub struct FactorInstancesProviderOutcome {
```
Sajjon, line 6: Check doc

crates/sargon/src/factor_instances_provider/provider/outcome/internal_factor_instances_provider_outcome.rs
```diff
@@ -2,13 +2,107 @@ use crate::prelude::*;
...
+    }
+
+    pub fn get_for_derivation_preset(
+        &self,
+        preset: DerivationPreset,
+    ) -> Option<&InternalFactorInstancesProviderOutcomePerFactor> {
+        self.per_derivation_preset.get(&preset)
+    }
+
+    pub fn get_for_derivation_preset_for_factor(
```
Sajjon, line 32: Used?

crates/sargon/src/factor_instances_provider/provider/outcome/internal_factor_instances_provider_outcome.rs
```diff
@@ -2,13 +2,107 @@ use crate::prelude::*;
...
+            DerivationPreset,
+            InternalFactorInstancesProviderOutcomePerFactor,
+        >,
+    ) -> Self {
+        Self {
+            per_derivation_preset,
+        }
+    }
+
+    /// "Transposes"
```
Sajjon, line 62: Doc

crates/sargon/src/factor_instances_provider/provider/outcome/internal_factor_instances_provider_outcome.rs
```diff
@@ -23,19 +117,19 @@ impl InternalFactorInstancesProviderOutcome {
...
-                .into_iter()
-                .map(|(k, v)| {
-                    (
-                        k,
-                        InternalFactorInstancesProviderOutcomeForFactor::satisfied_by_cache(k, v),
-                    )
-                })
-                .collect(),
-        )
+                pf_found_in_cache
```
Sajjon, line 120: Fmt?

crates/sargon/src/factor_instances_provider/provider/provider_adopters/cache_filler.rs
```diff
@@ -3,6 +3,18 @@ use crate::prelude::*;
 /// Uses a `FactorInstancesProvider` to fill the cache with instances for a new FactorSource.
 pub struct CacheFiller;
 
+pub struct CacheFillingQuantities;
+impl CacheFillingQuantities {
+    pub fn for_factor_source(id: FactorSourceIDFromHash) -> QuantitiesToDerive {
```
Sajjon, line 8: Refactor / share? Doc!

crates/sargon/src/factor_instances_provider/provider/provider_adopters/securify_entity_factor_instances_provider.rs
```diff
@@ -129,13 +127,24 @@ impl SecurifyEntityFactorInstancesProvider {
...
         let (instances_in_cache_consumer, outcome) = provider
-            .provide(
-                QuantifiedDerivationPreset::new(
-                    DerivationPreset::mfa_entity_kind(entity_kind),
-                    addresses_of_entities.len(),
-                ),
-                DerivationPurpose::for_securifying_or_updating(entity_kind),
+            .provide_for_presets(
+                // QuantifiedDerivationPreset::new(
+                //     DerivationPreset::mfa_entity_kind(entity_kind),
```
Sajjon, line 143: Dead code

Sajjon, line 143: Dead code

crates/sargon/src/factor_instances_provider/provider/provider_adopters/securify_entity_factor_instances_provider.rs
```diff
@@ -310,35 +319,51 @@ mod tests {
...
+                DerivationPreset::AccountMfa,
+                &bdfs.id_from_hash(),
+            )
+            .unwrap();
+        assert_eq!(account_outcome.to_use_directly.len(), 1);
+
+        // let profile = Arc::new(os.profile().unwrap());
+        // let (instances_in_cache_consumer, outcome) = SUT::for_persona_mfa(
+        //     cache_client.clone(),
+        //     profile,
```
Sajjon, line 350: Dead code , remove

crates/sargon/src/types/samples/account_samples.rs
```diff
@@ -17,6 +17,7 @@ static ALL_ACCOUNT_SAMPLES: Lazy<[Account; 10]> = Lazy::new(|| {
         // Carla | 2 | Securified { Single Threshold only }
         Account::sample_securified_mainnet(
             "Carla",
+2,
```
Sajjon, line 20: Fmt

crates/sargon/src/types/samples/persona_samples.rs
```diff
@@ -67,6 +71,7 @@ static ALL_PERSONA_SAMPLES: Lazy<[Persona; 8]> = Lazy::new(|| {
         // Kasparov | 6 | Securified { Threshold #3 and Override factors #2  }
         Persona::sample_securified_mainnet(
             "Kasparov",
+        6,
```
Sajjon, line 74: Ident

crates/sargon/src/keys_collector/collector/derivation_purpose.rs
```diff
@@ -12,11 +12,15 @@ pub enum DerivationPurpose {
     /// for identity VECIs
     CreatingNewPersona,
 
-    /// When applying a security shield to an account, initiates keys collection
+    /// When applying a security shield to accounts and personas mixed, initiates keys collection
+    /// for account MFA
+    SecurifyingAccountsAndPersonas,
```
CyonAlexRDX, line 17: @micbakos-rdx we will automatically select this as `DerivationPurpose` if we are securifying both account **and** personas mixed - but as we have talked about in the past, host probably maps all keys to the same copy.

crates/sargon/src/factor_instances_provider/provider/factor_instances_provider.rs
```diff
@@ -137,136 +154,220 @@ impl FactorInstancesProvider {
 
     async fn derive_more_and_cache(
```
CyonAlexRDX, line 155: upgraded to be PER DerivationPreset and then PER FactorSource

crates/sargon/src/factor_instances_provider/factor_instances_cache/factor_instances_cache.rs
```diff
@@ -220,108 +225,284 @@ impl FactorInstancesCache {
...
-                quantities_to_derive: pf_pdp_qty_to_derive,
-            }
+            CachedInstancesWithQuantitiesOutcome::Satisfied(CacheSatisfied {
+                cached: per_derivation_preset
+                    .into_iter()
+                    // Satisfied, but `per_derivation_preset` contains ALL Presets (in case of `NotSatisfied` - we are cache filling),
+                    // so we filter out only the originally requested ones.
+                    .filter(|(preset, _)| {
+                        originally_request_presets.contains(preset)
+                    })
```
GhenadieVP, line 373: This filter could be done before `is_quantity_unsatisfied_for_any_requested` and then `originally_request_presets.contains(preset)` could be removed.

crates/sargon/src/factor_instances_provider/factor_instances_cache/factor_instances_cache.rs
```diff
@@ -220,108 +225,284 @@ impl FactorInstancesCache {
...
+                if per_factor.is_empty() {
+                    None
+                } else {
+                    Some((preset, per_factor))
+                }
+            })
+            .collect::<IndexMap<
+            DerivationPreset,
+            IndexMap<FactorSourceIDFromHash, R>,
+        >>()
```
GhenadieVP, line 459: Can't Rust infer this type?

crates/sargon/src/factor_instances_provider/provider/provider_adopters/virtual_entity_creating_instance_provider.rs
```diff
@@ -176,23 +162,55 @@ mod tests {
...
+                .values()
+                .all(|y| y.debug_was_cached.len()
+                    == preset.cache_filling_quantity())
+        ));
+
+        for (k, v) in outcome.per_derivation_preset.iter() {
+            for (_, y) in v.per_factor.iter() {
+                let derivation_based_offset =
+                    if *k == DerivationPreset::IdentityVeci {
+                        1 /* One account created */
```
GhenadieVP, line 191: account or identity? the key is IdentityVeci

GhenadieVP (approved):
LGTM! Nice work!

Left some minor comments.

crates/sargon-uniffi/src/keys_collector/derivation_purpose.rs
```diff
@@ -23,6 +23,10 @@ pub enum DerivationPurpose {
     /// for identity MFA
     SecurifyingPersona,
 
+    /// When applying a security shield to accounts and personas mixed, initiates keys collection
+    /// for account MFA
+    SecurifyingAccountsAndPersonas,
```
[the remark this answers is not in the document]
CyonAlexRDX: Yes if you only security account or only security personas those other are used. 

But host will most likely map all variants to same Crowdin key anyway...

crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -135,6 +140,17 @@ impl SecurityShieldBuilder {
         self.set(|builder| builder.set_name(&name));
     }
 
+    pub fn set_authentication_signing_factor(
```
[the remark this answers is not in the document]
CyonAlexRDX: Add / remove makes sense when we support multiple factors. This is always one single

crates/sargon/src/factor_instances_provider/agnostic_paths/quantified_derivation_preset.rs
```diff
@@ -7,11 +7,103 @@ pub struct QuantifiedDerivationPreset {
...
         }
     }
+
+    /// Returns a the QuantifiedDerivationPresets needed to securify the `addresses_of_entities`, including
+    /// a new Authentication Signing factor instance for each entity. Will return
+    /// the `Account` variant of each DerivationPreset for each Account in `addresses_of_entities`
+    /// and the `Identity` variant of each DerivationPreset for each Persona in `addresses_of_entities`.
+    pub fn securifying_unsecurified_entities(
+        addresses_of_entities: &IndexSet<AddressOfAccountOrPersona>,
+    ) -> IdentifiedVecOf<Self> {
```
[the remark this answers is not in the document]
CyonAlexRDX: IdentifiedVec is more safe to use. 

It ensures that we not allow entities of the same Address, even though their other state might differ.

If fact, there are many many places were we currently are using IndexSet but ought to use IdentifiedVecOf

crates/sargon/src/factor_instances_provider/factor_instances_cache/factor_instances_cache.rs
Sajjon: you mean you can:

```rust
let per_derivation_preset = per_derivation_preset
                    .into_iter()
                    // Satisfied, but `per_derivation_preset` contains ALL Presets (in case of `NotSatisfied` - we are cache filling),
                    // so we filter out only the originally requested ones.
                    .filter(|(preset, _)| {
                        originally_request_presets.contains(preset)
                    }).collect();
```
?

That we cannot, since we need to *not* filter for case `CachedInstancesWithQuantitiesOutcome::NotSatisfied`.

So filtering should only happen for `CachedInstancesWithQuantitiesOutcome::Satisfied`.

Or maybe I misunderstood you..?

GhenadieVP: sure, but you can have it stored in a separate variable, aka filtered_presets. This is a minor suggestion anyway.

crates/sargon-uniffi/src/keys_collector/derivation_purpose.rs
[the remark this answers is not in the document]
GhenadieVP: In the host API, it will send a list of AccountOrPersona right? and then in Sargon you map to specific purpose based on the input, I think I've seen this logic in this PR.

PR #307, opened 2024-12-17
Transaction Analysis allow unknown accounts

GhenadieVP:
Allow transaction manifests which contain unknown accounts which to require signatures. The Wallet should not reject such transactions.

References:
- https://rdxworks.slack.com/archives/C031A0V1A1W/p1734036514557019?thread_ts=1733976138.385129&cid=C031A0V1A1W
- https://rdxworks.slack.com/archives/C031A0V1A1W/p1734356845160559

The new behaviour was introduced with Sargon, and we reject the manifest at the moment we try to extract the public keys to be sent to the GW. Old implementation, did allow for unknown accounts, [see iOS for reference](https://github.com/radixdlt/babylon-wallet-ios/blob/4b0738caf583521751a5c03f3c6410ab956c68a3/RadixWallet/Clients/TransactionClient/TransactionClient%2BLive.swift#L50).

### DIFF
--- a/crates/sargon/src/signing/collector/extractor_of_instances_required_to_sign_transactions.rs
+++ b/crates/sargon/src/signing/collector/extractor_of_instances_required_to_sign_transactions.rs
@@ -47,13 +47,19 @@ mod tests {
 
     #[test]
     fn preprocessor_init_fail() {
+        let intent_with_invalid_persona =
+            TransactionIntent::sample_entity_addresses_requiring_auth(
+                vec![],
+                vec![Persona::sample_mainnet().address],
+            );
+
         let result = ExtractorOfInstancesRequiredToSignTransactions::extract(
             &Profile::sample_other(),
-            vec![TransactionIntent::sample()],
+            vec![intent_with_invalid_persona],
             RoleKind::Primary,
         );
 
-        assert!(matches!(result, Err(CommonError::UnknownAccount)));
+        assert!(matches!(result, Err(CommonError::UnknownPersona)));
     }
 
     #[test]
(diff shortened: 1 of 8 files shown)


crates/sargon/src/signing/collector/extractor_of_instances_required_to_sign_transactions.rs
```diff
@@ -47,13 +47,19 @@ mod tests {
...
+
         let result = ExtractorOfInstancesRequiredToSignTransactions::extract(
             &Profile::sample_other(),
-            vec![TransactionIntent::sample()],
+            vec![intent_with_invalid_persona],
             RoleKind::Primary,
         );
 
-        assert!(matches!(result, Err(CommonError::UnknownAccount)));
+        assert!(matches!(result, Err(CommonError::UnknownPersona)));
```
GhenadieVP, line 62: Still invalid for unknown persona. We might change it in the future, but feels strange for a manifest to contain an unknown persona.

crates/sargon/src/signing/extractor_of_entities_requiring_auth.rs
```diff
@@ -17,7 +17,8 @@ impl ExtractorOfEntitiesRequiringAuth {
             .addresses_of_accounts_requiring_auth
             .iter()
             .map(|a| profile.account_by_address(*a))
-            .collect::<Result<Vec<_>>>()?;
+            .filter_map(|a| a.ok())
+            .collect::<Vec<_>>();
```
GhenadieVP, line 21: The actual fix.

micbakos-rdx (approved):

sergiupuhalschi-rdx (approved):

danvleju-rdx (approved):

CyonAlexRDX (approved):

crates/sargon/src/system/sargon_os/transactions/sargon_os_transaction_analysis.rs
```diff
@@ -582,7 +582,11 @@ mod transaction_preview_analysis_tests {
             )
             .await;
 
-        assert_eq!(result, Err(CommonError::UnknownAccount))
+        // Just asserts that the execution path reached GW preview call
+        assert!(matches!(
+            result,
+            Err(CommonError::ExecutionSummaryFail { .. })
+        ))
```
matiasbzurovski, line 589: This test is still located among the "tests that fail for some reason" group. I would move it under `success()` one, and probably add the needed stubs so that we don't check for an error (which may lead to think this should fail).

matiasbzurovski (approved):

PR #312, opened 2024-12-20
Transform UniFFI SecurityShieldBuilder to value type semantics

danvleju-rdx:
To improve the handling of TCA state in the iOS host, we need to transform the UniFFI `SecurityShieldBuilder` to use value type semantics, similar to how the [TransactionBuilder is implemented in radix-engine-toolkit](https://github.com/radixdlt/radix-engine-toolkit/blob/2ab5b35a6cb96c9faeb98ed8df67b870336e6873/crates/radix-engine-toolkit-uniffi/src/builder/manifest_builder/builder_v1.rs#L24).

This change means that every mutating function will return a new instance of `SecurityShieldBuilder`.

### DIFF
--- a/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
+++ b/crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
@@ -18,32 +18,10 @@ use crate::prelude::*;
 /// A builder of `SecurityStructureOfFactorSourceIds` a.k.a. `SecurityShield`,
 /// which contains a MatrixOfFactorSourceIds - with primary, recovery, and
 /// confirmation roles.
-#[derive(Debug, uniffi::Object)]
+#[derive(Debug, Hash, PartialEq, Clone, uniffi::Object)]
 #[uniffi::export(Hash, Eq)]
 pub struct SecurityShieldBuilder {
-    wrapped: RwLock<sargon::SecurityShieldBuilder>,
-}
-
-impl std::hash::Hash for SecurityShieldBuilder {
-    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        self.wrapped.read().unwrap().hash(state);
-    }
-}
-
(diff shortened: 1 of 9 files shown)


crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -19,10 +19,33 @@ use crate::prelude::*;
 /// which contains a MatrixOfFactorSourceIds - with primary, recovery, and
 /// confirmation roles.
 #[derive(Debug, uniffi::Object)]
+#[uniffi::export(Hash, Eq)]
 pub struct SecurityShieldBuilder {
     wrapped: RwLock<sargon::SecurityShieldBuilder>,
```
GhenadieVP, line 24: Do we still need RwLock?

danvleju-rdx: No, removed

crates/sargon/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -53,6 +53,81 @@ impl Default for SecurityShieldBuilder {
     }
 }
 
+impl PartialEq for SecurityShieldBuilder {
+    fn eq(&self, other: &Self) -> bool {
+        let (matrix, name, authentication_signing_factor) = (
+            self.matrix_builder
+                .read()
+                .expect("Failed to read matrix_builder"),
+            self.name.read().expect("Failed to read name"),
```
CyonAlexRDX, line 62: we might wanna use `try_read` here? but we can merge without. Neither Eq nor Clone impl should blocking wait

CyonAlexRDX (approved):
Maybe not blocking wait in clone and eq.

sergiupuhalschi-rdx (approved):

GhenadieVP (approved):

PR #314, opened 2024-12-20
Upgrade `ShieldBuilder` to validate `FactorSourceKind` for ROLA factors

CyonAlexRDX:
Upgrade ShieldBuilder to validate `FactorSourceKind` for ROLA factor.

Also expose functions to query ShieldBuilder for which FactorSourceKinds are valid for ROLA.


crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -252,6 +252,29 @@ impl SecurityShieldBuilder {
 
 #[uniffi::export]
 impl SecurityShieldBuilder {
+    /// "Statically" queries which FactorSourceKinds are disallowed for authentication signing.
```
CyonAlexRDX, line 255: "statically" put in quoutes since UniFFI does not allow static methods to be exported.

Documented here so that one understands that this does not mutate any state.

danvleju-rdx (approved):
LGTM, just bump version :)

PR #315, opened 2024-12-20
Lenient shield builder

CyonAlexRDX:
Allows for more lenient `SecurityShieldBuilder`, by passing `SecurityShieldBuilderMode::Lenient` as `mode` to `SecurityShieldBuilder`.

# Changes
The `RoleBuilder` and `MatrixBuilder` have got a new variant of almost all their methods, having the suffix `_with_mode` which takes a `mode: SecurityShieldBuilderMode`. The reason why this is not a field on the `MatrixBuilder` and `role_builder` is because it requires bigger changes to `AbstractRoleBuilderOrBuilt` and `AbstractMatrixBuilderOrBuilt` - which we can do but a bit more intrusive change. We could potentially add a serde_skipped extra field which we set to `()` for `Built` but to `SecurityShieldBuilderMode` (which we might rename to `FactorBuilderMode` to fit more into `RoleBuilder`) for **non-Built** (i.e. builders).

The SecurityShield builder than pass `self.mode` to all required methods.

The RoleBuilder has been updated to make use of `mode` passed as method argument to be more lenient - when applicable.


crates/sargon/src/profile/mfa/security_structures/roles/builder/roles_builder.rs
```diff
@@ -539,14 +608,40 @@ impl<const ROLE: u8> RoleBuilder<ROLE> {
...
+            }
+            SecurityShieldBuilderMode::Lenient => {
+                match factor_list_kind {
+                    Override => {
+                        if self
+                            .override_contains_factor_source(factor_source_id)
+                        {
+                            return duplicates_err;
+                        }
+                        // but if threshold contains it, we're good (since mode is lenient)
```
CyonAlexRDX, line 631: With Lenient, we allow a factor which **is** in `override` to be added to `threshold` (and vice versa)

crates/sargon/src/profile/mfa/security_structures/roles/builder/roles_builder.rs
```diff
@@ -629,14 +735,18 @@ impl<const ROLE: u8> RoleBuilder<ROLE> {
...
+                    if self.contains_factor_source_of_kind(
+                        FactorSourceKind::Device,
+                    ) {
+                        return RoleBuilderMutateResult::forever_invalid(
+                            PrimaryCannotHaveMultipleDevices,
+                        );
+                    }
                 }
-            }
+                SecurityShieldBuilderMode::Lenient => {}
```
CyonAlexRDX, line 748: for `Lenient` we do not count the number of `Device` FactorSource, we allow many.

crates/sargon/src/profile/mfa/security_structures/roles/builder/roles_builder.rs
```diff
@@ -707,6 +829,10 @@ impl<const ROLE: u8> RoleBuilder<ROLE> {
             );
         }
 
+        if mode == SecurityShieldBuilderMode::Lenient {
+            return Ok(());
```
CyonAlexRDX, line 833: For `Lenient` we allow `password` by itself.

danvleju-rdx (approved):
Very nice!
You can merge it, and I'll handle the conflicts on my branch when I’m back.

CyonAlexRDX:
pushing a fix for swift tests now

sergiupuhalschi-rdx (approved):
Perfect!

PR #321, opened 2024-12-23
Split out module `core` into separate crate

CyonAlexRDX:
Split out the module `core` from Sargon.

Notable changes:
- I had to remove the associated values of the `CommonError` Enum


Sajjon:
@GhenadieVP or @micbakos-rdx if you are working tomorrow, this is ready for review. We might perhaps in fact merge it to main - I need to test hosts though! Most likely not a problem since this does not change UniFFI (or should not!)

I had to do something silly with `into_iter_result` - maybe it can be cleaned up. Ghenadie might have ideas? The problem is that before this PR all our `Accounts` / `Personas` / `FactorSources` etc were all `pub type Accounts = IdentifiedVecOf<Account>` or `pub type Personas = IdentifiedVecOf<Persona>` etc, meaning they were all really `IdentifiedVecOf`. And we can easily impl trait for one concrete type. 

but with this PR they are all distinct newtypes around `IdentifiedVecOf` - so that we can further split the crates, i.e.:
have `IdentifiedVecOf` in one crate, have `HasSampleaValues` in another crate, have `Account` in a third crate and have `Accounts` in a fourth crate. This PR makes it work, but it is a bit ugly in UniFFI crate - which I care the least about since it should just work, need not be too pretty.

crates/sargon-uniffi/src/core/result/result.rs
```diff
@@ -33,6 +33,11 @@ pub trait FromInternalResult<InternalType, Type> {
     fn into_result(self) -> Result<Type>;
 }
 
+/// Utility trait to map `InternalResult` to `Result`
+pub trait FromInternalIterResult<InternalType, Type> {
```
Sajjon, line 37: Uh this is not so great, but not too bad. Instead of `into_result()`  on `Result<Vec<T>>` we need to call `into_iter_result()`....

crates/sargon-uniffi/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -18,17 +19,18 @@ use crate::prelude::*;
 /// A builder of `SecurityStructureOfFactorSourceIds` a.k.a. `SecurityShield`,
 /// which contains a MatrixOfFactorSourceIds - with primary, recovery, and
 /// confirmation roles.
-#[derive(Debug, uniffi::Object)]
+#[derive(Debug, Hash, PartialEq, Clone, uniffi::Object)]
+#[uniffi::export(Hash, Eq)]
 pub struct SecurityShieldBuilder {
-    wrapped: RwLock<sargon::SecurityShieldBuilder>,
+    wrapped: Arc<sargon::SecurityShieldBuilder>,
```
Sajjon, line 25: Hmm weird that Git think this is new... I've merged main into my branch. hmm....

Sajjon, line 25: strange that Git thinks this has changed. I've merged `main` into both this branch (`ac/split-core`) and target branch `ac/split`. Hmm... 🤷‍♂️

Sajjon:
> [!NOTE]
> Not DRAFT - Ready for review - but I cannot change status of PR with Private Github user

PR #338, opened 2025-01-09
Update role threshold model

sergiupuhalschi-rdx:
[Ticket](https://radixdlt.atlassian.net/browse/ABW-4047)

Expressed the role threshold of a security structure with an `enum` instead of a value. This way we can reflect the user preference regarding the number of factors required to sign with.  By default, the threshold should be set to `All`.

### DIFF
--- a/crates/profile/logic/logic_SPLIT_ME/src/tests/matrix_of_factor_instances_index_agnostic.rs
+++ b/crates/profile/logic/logic_SPLIT_ME/src/tests/matrix_of_factor_instances_index_agnostic.rs
@@ -9,7 +9,7 @@ type SUT = MatrixOfFactorInstances;
 fn wrong_entity_kind() {
     let invalid = unsafe {
         SUT::unbuilt_with_roles_and_days(
-            PrimaryRoleWithFactorInstances::unbuilt_with_factors(0, [
+            PrimaryRoleWithFactorInstances::unbuilt_with_factors(Threshold::All, [
                 HierarchicalDeterministicFactorInstance::sample_mainnet_entity_device_factor_fs_0_securified_at_index(
                 CAP26EntityKind::Account,
                 0,
@@ -18,12 +18,12 @@ fn wrong_entity_kind() {
                 1,
             ).into()], []),
             RecoveryRoleWithFactorInstances::unbuilt_with_factors(
-                0,
+                Threshold::Specific(0),
                 [],
                 [],
             ),
             ConfirmationRoleWithFactorInstances::unbuilt_with_factors(
-                0,
+                Threshold::Specific(0),
                 [],
                 [],
             ),
@@ -41,7 +41,7 @@ fn wrong_entity_kind() {
 fn wrong_key_kind() {
     let invalid = unsafe {
         SUT::unbuilt_with_roles_and_days(
-            PrimaryRoleWithFactorInstances::unbuilt_with_factors(0, [
+            PrimaryRoleWithFactorInstances::unbuilt_with_factors(Threshold::All, [
                 HierarchicalDeterministicFactorInstance::sample_mainnet_entity_device_factor_fs_0_securified_at_index(
                 CAP26EntityKind::Account,
                 0,
@@ -53,12 +53,12 @@ fn wrong_key_kind() {
                 SecurifiedU30::ZERO
             ).into()], []),
             RecoveryRoleWithFactorInstances::unbuilt_with_factors(
-                0,
+                Threshold::Specific(0),
                 [],
                 [],
             ),
             ConfirmationRoleWithFactorInstances::unbuilt_with_factors(
-                0,
+                Threshold::Specific(0),
                 [],
                 [],
             ),
--- a/crates/profile/logic/logic_SPLIT_ME/src/tests/matrix_of_factor_instances_index_agnostic.rs
+++ b/crates/profile/logic/logic_SPLIT_ME/src/tests/matrix_of_factor_instances_index_agnostic.rs
@@ -18,12 +18,12 @@ fn wrong_entity_kind() {
                 1,
             ).into()], []),
             RecoveryRoleWithFactorInstances::unbuilt_with_factors(
-                Threshold::Specific(0),
+                Threshold::zero(),
                 [],
                 [],
             ),
             ConfirmationRoleWithFactorInstances::unbuilt_with_factors(
-                Threshold::Specific(0),
+                Threshold::zero(),
                 [],
                 [],
             ),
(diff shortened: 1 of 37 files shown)


crates/uniffi/uniffi_SPLIT_ME/src/profile/mfa/security_structures/models/threshold.rs
```diff
@@ -0,0 +1,14 @@
...
+use sargon::Threshold as InternalThreshold;
+
+/// A kind of threshold, either All or a specific number of factors
+/// must be used to perform some function with.
+#[derive(
+    Clone, Copy, Debug, PartialEq, Eq, Hash, InternalConversion, uniffi::Enum,
+)]
+pub enum Threshold {
+    /// All factors in the threshold factors list must be used to perform some function with
+    All,
```
CyonAlexRDX, line 11: any unit test testing JSON of `all` ?

CyonAlexRDX: but only in isolation
no test when the object is put inside a role?

sergiupuhalschi-rdx: Not yet, for now, I just updated the existing tests. I'll add tests for the new code as well.

crates/profile/models/security-structures/src/roles_matrices_structures/roles/builder/roles_builder.rs
```diff
@@ -716,11 +717,12 @@ impl<const ROLE: u8> RoleBuilder<ROLE> {
...
                 let threshold_factors_len =
                     self.get_threshold_factors().len() as u8;
-                if self.get_threshold() > threshold_factors_len {
-                    // N.B. we don't use `set_threshold` since this might be a
-                    // temporary invalid state, if e.g. primary role does not have
-                    // any factors.
-                    self.unchecked_set_threshold(threshold_factors_len);
+                if threshold_factors_len == 0 {
+                    self.unchecked_set_threshold(Threshold::All);
+                } else if self.get_threshold() > threshold_factors_len {
```
GhenadieVP, line 722: This will indirectly account for the case of threshold being All, since `get_threshold` will return `threshold_factors_len` when threshold is All, right?

crates/uniffi/uniffi_SPLIT_ME/src/profile/mfa/security_structures/security_shield_builder.rs
```diff
@@ -652,6 +652,32 @@ mod tests {
...
+
+        sut = sut
+            .remove_factor_from_primary(
+                FactorSourceID::sample_device(),
+                FactorListKind::Threshold,
+            )
+            .remove_factor_from_primary(
+                FactorSourceID::sample_password_other(),
+                FactorListKind::Threshold,
+            );
```
GhenadieVP, line 676: Assertions to validate that removing a factor decreases the threshold, or these are added somewhere else?
Also, an assertion for the case of going from specific to All, that is when removing all of the factor sources.

GhenadieVP (approved):
Nice work!

crates/profile/models/security-structures/src/roles_matrices_structures/roles/builder/roles_builder.rs
sergiupuhalschi-rdx: Good point. I think we want to preserve the user selection whenever we can. For example, if the user chooses `All` and removes a factor the threshold must still be `All`. I've updated the code to reflect this.

crates/uniffi/uniffi_SPLIT_ME/src/profile/mfa/security_structures/security_shield_builder.rs
sergiupuhalschi-rdx: Added [here](https://github.com/radixdlt/sargon/pull/338/files#diff-a861497d9b12a9be7a86d1a4823853bf17b0ebb252456c985c32fb9177f5814d)

crates/profile/models/security-structures/src/roles_matrices_structures/roles/builder/roles_builder.rs
```diff
@@ -714,13 +723,17 @@ impl<const ROLE: u8> RoleBuilder<ROLE> {
...
+
+                if let Threshold::Specific(_) = self.get_threshold_kind() {
+                    let threshold_factors_len =
+                        self.get_threshold_factors().len() as u8;
+                    if threshold_factors_len == 0 {
+                        self.unchecked_set_threshold(Threshold::All);
+                    } else {
+                        self.unchecked_set_threshold(Threshold::Specific(
+                            threshold_factors_len,
+                        ));
```
danvleju-rdx, line 735: Question: Should the threshold always be set to `threshold_factors_len`? For example, suppose the user has _5_ threshold factors and has explicitly set the threshold to `Specific(3)`. If they want to remove a threshold factor, should the threshold update to `Specific(4)` (i.e., `threshold_factors_len`) or remain `Specific(3)` (the value explicitly set by the user)?

Also,`fn remove_factor_source`'s docs should be updated. This part:

> /// Lowers the threshold if the deleted factor source is in the threshold list
    /// and if after removal of `factor_source_id` `self.threshold > self.threshold_factors.len()`

sergiupuhalschi-rdx: You're right! we shouldn't change the threshold value.

crates/profile/logic/logic_SPLIT_ME/src/tests/matrix_of_factor_instances_index_agnostic.rs
```diff
@@ -18,12 +18,12 @@ fn wrong_entity_kind() {
                 1,
             ).into()], []),
             RecoveryRoleWithFactorInstances::unbuilt_with_factors(
-                0,
+                Threshold::Specific(0),
```
CyonAlexRDX, line 21: consider adding `pub fn zero() -> Self` ctor to `Threshold` and replace `Threshold::Specific(0)` with `Threshold::zero()`.

crates/profile/models/security-structures/src/roles_matrices_structures/roles/abstract_role_builder_or_built.rs
```diff
@@ -146,6 +160,12 @@ impl<const ROLE: u8, const MODE: u8, FACTOR>
     /// How many threshold factors that must be used to perform some function with
     /// this role.
     pub fn get_threshold(&self) -> u8 {
+        self.threshold.value(self.threshold_factors.len())
+    }
+
+    /// How many threshold factors that must be used to perform some function with
+    /// this role.
+    pub fn get_threshold_kind(&self) -> Threshold {
```
CyonAlexRDX, line 168: hmm I think this should be called `get_threshold` and the method above -> `get_threshold_value`. not sure though. this looks a bit wrong. and since we call `.value` in the function above I think that would make sense.

CyonAlexRDX (approved):
LGTM, good job!

danvleju-rdx (approved):

PR #340, opened 2025-01-10
Apply security shield API

CyonAlexRDX:
# New API
Adds this UniFFI export API:
```rust
#[uniffi::export]
impl SargonOS {
    async fn apply_security_shield_with_id_to_entities(
        &self,
        security_shield_id: SecurityStructureID,
        addresses: IndexSet<AddressOfAccountOrPersona>,
    ) -> Result<()> { ... }
}
```

Which does not change the `security_state` of entities matching, instead it changes the `provisional_security_config`. If any of the entities already has a `provisional_security_config` set, then we fail. If we fail to load the security shield with `security_shield_id` we fail. 

When success, we have used the FactorInstancesProvider to provide instances for each entity for the shield and saved those instances into the `provisional_security_config`.

# Implementation
Here is the highlight of this implementation (delta):
* The POC `cfg(test)`-only implementations I had which was called "securified" changed the `security_state: EntitySecurityState` on entities from `::Unsecured` to `::Secured` on entities, with the `SecurityStructureOfFactorInstances`, that has change to instead set the `provisional_security_config`. If one is already set => fail.
* Upgrade `NextDerivationEntityIndexProfileAnalyzingAssigner` to be aware of `provisional_security_config` (when it is `Some`) on entities - this is needed now since I no longer change the `security_state: EntitySecurityState` on entities, so the `NextDerivationEntityIndexProfileAnalyzingAssigner` must analyze `provisional_security_config` to not try to reuse those Derivation Entity Indices (note that the FactorInstance have been consumed and deleted from the cache, but the `NextDerivationEntityIndexProfileAnalyzingAssigner` runs before to determine the derivation path and would cause the FactorInstancesProvider to re-derive FactorInstances with those derivation paths...).
* 

# Other changes:
* the ctor `SecurityStructureOfFactorInstances ::fulfilling_structure_of_factor_sources_with_instances` now reuses the ROLA key if the factor source for ROLA has not changed
* Removed two variants `::ShieldSelected` and `::TransactionQueued` of `ProvisionalSecurityConfig`
* Removed all logic related to state transitions for `provisional_security_config`
* Removed `ProvisionalSecurifiedTransactionQueued` type (was associated value of the `::TransactionQueued` variant.
* Started using `"prelude"` crate, using it to publically re-export `std::sync::{Arc,RwLock}` and thus remove those imports from all other crates
* Same but for `std::str::FromStr`
* moved `test_instances_provider` from `"sargon-os"` crate to `"os-factors"` crate => which forced me to use `#[cfg(debug_assertions)]` inside SargonOS in `"sargon-os"` crate, instead of `#[cfg(test)]` so that the code snippet to eaglery pre-derive factor instances for FactorSources is run, since `#[cfg(debug_assertions)]` **is set between crates**.
* Tarpaulin config was updated, the ignored files had wrong (forgot to update) paths, also ensured that the default output format is `Html` which we want to use when we run locally. **I've removed `force-clean` by default, I've updated CI to manually override output format with `Xml` and with `force-clean`**. If you have any issues locally just manually clean.


crates/system/os/factors/src/sargon_os_apply_shield.rs
```diff
@@ -0,0 +1,331 @@
...
+        addresses: IndexSet<AddressOfAccountOrPersona>,
+    ) -> Result<()> {
+        let shield = self
+            .security_structure_of_factor_sources_from_security_structure_id(
+                security_shield_id,
+            )?;
+        self.apply_shield_to_entities(&shield, addresses).await
+    }
+
+    async fn _apply_shield_to_entities_with_diagnostics(
```
CyonAlexRDX, line 69: this method is mostly migrated as is from `sargon-os` method `__OFFLINE_ONLY_securify_entities`, here https://github.com/radixdlt/sargon/blob/main/crates/system/os/os/src/os_testing_support.rs#L148-L208

But with changes I highlight below (inline)

crates/system/os/factors/src/sargon_os_apply_shield.rs
```diff
@@ -0,0 +1,331 @@
...
+        let securified_entities = entity_addresses
+            .into_iter()
+            .map(|entity_address| {
+                let security_structure_of_factor_instances =
+                    security_structures_of_factor_instances
+                        .shift_remove(&entity_address)
+                        .unwrap();
+
+                let mut entity = self.entity_by_address(entity_address)?;
+                entity.set_provisional(
```
CyonAlexRDX, line 115: here we used to call a helper method called `__OFFLINE_ONLY_securify_entity_without_saving` here https://github.com/radixdlt/sargon/blob/main/crates/system/os/os/src/os_testing_support.rs#L186C22-L186C67
and declaration https://github.com/radixdlt/sargon/blob/main/crates/system/os/os/src/os_testing_support.rs#L89-L123
which changed the `entity_security_state` (we called `set_security_state`) now instead we call `set_provisional`.

CyonAlexRDX, line 115: here we used to call a helper method called `__OFFLINE_ONLY_securify_entity_without_saving` here https://github.com/radixdlt/sargon/blob/main/crates/system/os/os/src/os_testing_support.rs#L186C22-L186C67
and declaration https://github.com/radixdlt/sargon/blob/main/crates/system/os/os/src/os_testing_support.rs#L89-L123
which changed the `entity_security_state` (we called `set_security_state`) now instead we call `set_provisional`.

crates/system/os/os/src/os_testing_support.rs
```diff
@@ -1,9 +1,8 @@
 #![allow(non_snake_case)]
 
-#[cfg(test)]
 use crate::prelude::*;
 
-#[cfg(test)]
+#[cfg(debug_assertions)]
```
CyonAlexRDX, line 5: need to use `debug_assertions` so that this code is available from other crates and only for DEBUG

crates/system/os/os/src/sargon_os_accounts.rs
```diff
@@ -859,137 +859,6 @@ impl SargonOS {
     }
 }
 
-// # Securify
-impl SargonOS {
-    #[allow(dead_code)]
-    #[cfg(test)]
-    pub(crate) async fn make_security_structure_of_factor_instances_for_entities_without_consuming_cache_with_derivation_outcome(
```
CyonAlexRDX, line 866: moved to `"sargon-os-factors"` crate, and renamed to:

`_provide_instances_for_shield_for_entities_without_consuming_cache`

crates/uniffi/uniffi_SPLIT_ME/src/profile/v100/entity_security_state/provisional_securified_config.rs
```diff
@@ -1,46 +1,15 @@
...
+/// derived the factor instances yet.
 #[derive(Clone, PartialEq, Eq, Hash, InternalConversion, uniffi::Enum)]
 pub enum ProvisionalSecurifiedConfig {
-    /// User has selected which security shield to use for some entity,
-    /// but no FactorInstances has been provided yet.
-    ShieldSelected { value: SecurityStructureID },
-
     /// User has fully prepared a `SecurityStructureOfFactorInstances` but
     /// not made a transaction to apply it to the entity yet.
     FactorInstancesDerived {
```
CyonAlexRDX, line 12: Now this enum has a single variant `FactorInstancesDerived`, in future we might reintroduce the other two, maybe... We keep this as an enum so that the JSON structure of Profile changed if we were to do so.

crates/uniffi/uniffi_SPLIT_ME/src/system/drivers/unsafe_storage_driver/support/test/ephemeral_unsafe_storage.rs
CyonAlexRDX, line 1: this was unused

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/entities/apply_shield/sargon_os_apply_shield.rs
```diff
@@ -0,0 +1,23 @@
+use crate::prelude::*;
+
+use sargon::OsShieldApplying;
+
+// ==================
+// Apply Shield to entities
+// ==================
+#[uniffi::export]
+impl SargonOS {
+    pub async fn apply_security_shield_to_entities(
```
CyonAlexRDX, line 10: The goal/purpose of this PR.... called from hosts when user has saved the shield and have when they have selected which accounts/personas to apply the shield to.

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/entities/mod.rs
CyonAlexRDX, line 1: just grouped entity related modules together.

crates/factors/instances-provider/src/next_index_assigner/next_derivation_entity_index_profile_analyzing_assigner.rs
CyonAlexRDX, line 1: this file is probably easieast to **NOT** review in split mode in github

crates/factors/instances-provider/src/next_index_assigner/next_derivation_entity_index_profile_analyzing_assigner.rs
```diff
@@ -142,79 +142,52 @@ impl NextDerivationEntityIndexProfileAnalyzingAssigner {
...
-                    entity_kind: CAP26EntityKind::Account,
-                    key_space: KeySpace::Securified,
-                }
-                .matches(&fi.derivation_path())
-                .index()
-            })
-            .max()
+            .max();
+
+        max_securified.max(max_provisional_unsecurified)
```
CyonAlexRDX, line 173: For `KeySpace::Secure` (`DerivationPreset::AccountMfa` or `DerivationPreset::IdentityMfa` ) we must check `max` of provisional and non-provisional (probably if `provisional` if set will always be hight, but safest to just use `max`).

CyonAlexRDX, line 173: For `KeySpace::Secure` (`DerivationPreset::AccountMfa` or `DerivationPreset::IdentityMfa` ) we must check `max` of provisional and non-provisional (probably if `provisional` if set will always be hight, but safest to just use `max`).

crates/system/os/factors/src/sargon_os_apply_shield.rs
```diff
@@ -0,0 +1,434 @@
...
+
+#[async_trait::async_trait]
+pub trait OsShieldApplying {
+    async fn apply_security_shield_to_entities(
+        &self,
+        security_shield_id: SecurityStructureID,
+        addresses: IndexSet<AddressOfAccountOrPersona>,
+    ) -> Result<()>;
+
+    async fn _apply_shield_to_entities_with_diagnostics(
```
GhenadieVP, line 11: why `_` prefix? These are used for testing in other crates?

crates/system/os/factors/src/sargon_os_apply_shield.rs
```diff
@@ -0,0 +1,434 @@
...
+    /// Without consuming cache, providers instances for the shield for the entities.
+    async fn _provide_instances_for_shield_for_entities_without_consuming_cache(
+        &self,
+        security_structure_of_factor_sources: SecurityStructureOfFactorSources, // Aka "shield"
+        entities: IndexSet<AccountOrPersona>,
+    ) -> Result<(
+        IndexMap<AddressOfAccountOrPersona, SecurityStructureOfFactorInstances>,
+        InstancesInCacheConsumer,
+        FactorInstancesProviderOutcome,
+    )>;
```
GhenadieVP, line 52: Should these be in the trait?

crates/system/os/factors/src/sargon_os_apply_shield.rs
```diff
@@ -0,0 +1,434 @@
...
+        // consume!
+        instances_in_cache_consumer.consume().await?;
+
+        let securified_entities = entity_addresses
+            .into_iter()
+            .map(|entity_address| {
+                let security_structure_of_factor_instances =
+                    security_structures_of_factor_instances
+                        .shift_remove(&entity_address)
+                        .unwrap();
```
GhenadieVP, line 112: `expect` instead?

crates/system/os/factors/src/sargon_os_apply_shield.rs
```diff
@@ -0,0 +1,434 @@
...
+            .collect::<Result<IdentifiedVecOf<AccountOrPersona>>>()?;
+
+        assert!(security_structures_of_factor_instances.is_empty());
+
+        // Assert that none of the NEW FactorInstances collide with the existing ones
+        self.profile()
+            .unwrap()
+            .assert_new_factor_instances_not_already_used_erased(
+                securified_entities.clone(),
+            )?;
```
GhenadieVP, line 131: How can we recover from this error? And should it be throwing? It seems to be similar to profile integrity check that you introduce, no?

crates/system/os/factors/src/sargon_os_apply_shield.rs
```diff
@@ -0,0 +1,434 @@
...
+            .iter()
+            .map(|e| e.address())
+            .collect::<IndexSet<AddressOfAccountOrPersona>>(
+        );
+
+        let profile_snapshot = self.profile()?;
+        let key_derivation_interactors = self.keys_derivation_interactor();
+
+        // We only consume ROLA factors for:
+        // * Unsecured entities (because they do not yet have nay ROLA key)
```
GhenadieVP, line 187: `nay` -> `any`

crates/system/os/factors/src/sargon_os_apply_shield.rs
```diff
@@ -0,0 +1,434 @@
...
+
+                let mut instances_per_factor_source = {
+                    let tx_preset =
+                        DerivationPreset::mfa_entity_kind(entity_kind);
+                    let rola_preset =
+                        DerivationPreset::rola_entity_kind(entity_kind);
+
+                    let instances_per_factor_source_mfa = instances_per_preset_per_factor_source
+                    .swap_remove(&tx_preset)
+                    .unwrap_or_else(|| panic!("Expected to find instances for derivation preset: {:?}", tx_preset));
```
GhenadieVP, line 287: Isn't this the same as `expect`?

crates/system/os/factors/src/sargon_os_apply_shield.rs
```diff
@@ -0,0 +1,434 @@
...
+
+                    let instances_per_factor_source_rola = instances_per_preset_per_factor_source
+                    .swap_remove(&rola_preset)
+                    .unwrap_or_else(|| panic!("Expected to find instances for derivation preset: {:?}", rola_preset));
```
GhenadieVP, line 291: same here as above

crates/system/os/factors/src/sargon_os_apply_shield.rs
```diff
@@ -0,0 +1,434 @@
...
+
+                    // Merge `instances_per_factor_source_mfa` and `instances_per_factor_source_rola` together
+                    let mut instances_per_factor_source =
+                        instances_per_factor_source_mfa;
+                    for (k, v) in instances_per_factor_source_rola {
+                        instances_per_factor_source.append_or_insert_to(k, v);
+                    }
```
GhenadieVP, line 298: Could `extend` do the same?

crates/system/os/factors/src/sargon_os_security_structures.rs
```diff
@@ -10,11 +10,36 @@ pub trait OsSecurityStructuresQuerying {
...
         structure_of_ids: &SecurityStructureOfFactorSourceIDs,
     ) -> Result<SecurityStructureOfFactorSources>;
 
+    async fn add_security_structure_of_factor_source_ids(
+        &self,
+        structure_of_ids: &SecurityStructureOfFactorSourceIDs,
+    ) -> Result<bool> {
+        let shield_of_sources = self.security_structure_of_factor_sources_from_security_structure_of_factor_source_ids(structure_of_ids)?;
+        self.add_security_structure_of_factor_sources(&shield_of_sources)
+            .await
```
GhenadieVP, line 40: This will probably not be needed anymore with Sergiu's work. Otherwise redundant work is done where -> convert from IDs to factor sources just o convert back to ids in `add_security_structure_of_factor_sources`

crates/system/os/factors/src/sargon_os_security_structures.rs
```diff
@@ -59,6 +84,22 @@ impl OsSecurityStructuresQuerying for SargonOS {
         })
     }
 
+    fn security_structure_of_factor_source_ids_from_security_structure_id(
```
GhenadieVP, line 87: `from_security_structure_id`, it looks like this more like `for_security_structure_id`

GhenadieVP (commented):
Looks very good, left some comments.

crates/system/os/factors/src/sargon_os_apply_shield.rs
CyonAlexRDX: yes exactly, used by tests and should not really be regarded part of the "public API"

crates/system/os/factors/src/sargon_os_apply_shield.rs
CyonAlexRDX: 💯

crates/system/os/factors/src/sargon_os_apply_shield.rs
CyonAlexRDX: No, `expect` does not allow formatting, so we cannot pass variables (here: `tx_preset`).

So this is the same as: `.expect(&format!(..))` but when running clippy fix, it changes to this.

crates/system/os/factors/src/sargon_os_apply_shield.rs
CyonAlexRDX: ditto

crates/system/os/factors/src/sargon_os_apply_shield.rs
GhenadieVP: Yes this needs to be changed to not be throwing. 
Asserts are good, but we will never fail because in development we work with release build. I think we ought to make also a debug version of sargon that we can use in development to catch these bad scenarios, otherwise we will encounter these only in release mode.
Should not be hard to build both flavours of Sargon on CD.

crates/system/os/factors/src/sargon_os_apply_shield.rs
GhenadieVP: interesting, good to know about clippy 👍.

crates/system/os/factors/src/sargon_os_apply_shield.rs
CyonAlexRDX: nope, `extend` does not merge when the **values** themselves are maps. this does that.

CyonAlexRDX: nope, `extend` does not merge when the **values** themselves are maps. this does that.

CyonAlexRDX: but we could try add this particular functionality to our `AppendableMap` trait perhaps https://github.com/radixdlt/sargon/blob/main/crates/core/collections/src/appendable_collection.rs#L17

crates/system/os/factors/src/sargon_os_security_structures.rs
CyonAlexRDX: exactly, I will update this post Sergiu merges his PR

crates/system/os/factors/src/sargon_os_security_structures.rs
CyonAlexRDX: I will change to `_by_` since it is a lookup and we usually use `by` then

CyonAlexRDX: I will change to `_by_` since it is a lookup and we usually use `by` then

crates/system/os/factors/src/sargon_os_apply_shield.rs
CyonAlexRDX: could be inside a different trait, but I think it is just more complex? then this trait needs to inherit from that trait. and this is outside of `sargon-os` crate, so must be inside some trait.

CyonAlexRDX: these traits are... "virtual" in the sense that we dont really care about those traits to much, just need them to circumvent the orphan rule.

crates/system/os/factors/src/sargon_os_apply_shield.rs
GhenadieVP: It's fine as is, this is minor.

GhenadieVP (approved):
LGTM!

crates/factors/instances-provider/src/next_index_assigner/next_derivation_entity_index_profile_analyzing_assigner.rs
```diff
@@ -226,21 +199,109 @@ impl NextDerivationEntityIndexProfileAnalyzingAssigner {
...
+            key_kind: CAP26KeyKind::AuthenticationSigning,
+            entity_kind,
+            key_space: KeySpace::Securified,
+        };
+
+        let max_unsecurified = unsecurified_entities
+        .iter()
+        .filter(|e| e.network_id() == self.network_id)
+        .flat_map(|e| {
+
```
CyonAlexRDX, line 228: TODO: remove two empty newlines

matiasbzurovski (approved):

danvleju-rdx (approved):

sergiupuhalschi-rdx (approved):

PR #341, opened 2025-01-10
[ABW-4131] Set main FactorSource

matiasbzurovski:
Allows setting a `FactorSource` as main. 
Although it will only be used by hosts for `Device` factor sources, implementation supports one main factor source for any given `FactorSourceKind`.

### DIFF
--- a/crates/system/os/os/src/sargon_os_factors.rs
+++ b/crates/system/os/os/src/sargon_os_factors.rs
@@ -364,7 +364,7 @@ impl SargonOS {
         Ok(())
     }
 
-    /// Set the given `factor_source_id` as the main factor source for the given `kind`.
+    /// Set the given `factor_source` as the main factor source of its kind.
     /// Throws `UpdateFactorSourceMutateFailed` error if the factor source is not found.
     ///
     /// # Emits Event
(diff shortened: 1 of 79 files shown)


crates/system/os/os/src/sargon_os_factors.rs
```diff
@@ -363,6 +363,50 @@ impl SargonOS {
...
+    /// Throws `UpdateFactorSourceMutateFailed` error if the factor source is not found.
+    ///
+    /// # Emits Event
+    /// Emits `Event::ProfileSaved` after having successfully written the JSON
+    /// of the active profile to secure storage.
+    ///
+    /// Also emits `EventNotification::ProfileModified { change: EventProfileModified::FactorSourceUpdated { id } }`
+    ///
+    /// If there is any main `FactorSource` of the given `FactorSourceKind`, such events are emitted also when
+    /// removing the flag from the old main factor source.
```
matiasbzurovski, line 377: note: consider doing it all in one round, adding a new `FactorSourcesUpdated { ids }` event

crates/system/os/os/src/sargon_os_factors.rs
```diff
@@ -363,6 +363,48 @@ impl SargonOS {
...
+        &self,
+        factor_source_id: impl Into<FactorSourceID>,
+        kind: FactorSourceKind,
+    ) -> Result<()> {
+        // If there is an existent main factor source of the same kind, remove the main flag
+        if let Some(current_main) = self
+            .profile_state_holder
+            .access_profile_with(|p| p.main_factor_source_of_kind(kind))??
+        {
+            self.update_factor_source_remove_flag_main(current_main)
```
GhenadieVP, line 388: I would opt to call this inside the `update_profile_with` below, to have both changes on profile saved together in one write operation to secure storage instead of two.

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/sargon_os_factors.rs
```diff
@@ -195,4 +195,29 @@ impl SargonOS {
...
+    /// of the active profile to secure storage.
+    ///
+    /// Also emits `EventNotification::ProfileModified { change: EventProfileModified::FactorSourceUpdated { id } }`
+    ///
+    /// If there is any main `FactorSource` of the given `FactorSourceKind`, such events are emitted also when
+    /// removing the flag from the old main factor source.
+    pub async fn set_main_factor_source_of_kind(
+        &self,
+        factor_source_id: FactorSourceID,
+        kind: FactorSourceKind,
```
GhenadieVP: As mentioned above, should not be needed.

CyonAlexRDX (approved):
LGTM

crates/system/os/os/src/sargon_os_factors.rs
```diff
@@ -363,6 +363,48 @@ impl SargonOS {
...
+    /// of the active profile to secure storage.
+    ///
+    /// Also emits `EventNotification::ProfileModified { change: EventProfileModified::FactorSourceUpdated { id } }`
+    ///
+    /// If there is any main `FactorSource` of the given `FactorSourceKind`, such events are emitted also when
+    /// removing the flag from the old main factor source.
+    pub async fn set_main_factor_source_of_kind(
+        &self,
+        factor_source_id: impl Into<FactorSourceID>,
+        kind: FactorSourceKind,
```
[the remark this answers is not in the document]
matiasbzurovski: not sure why you saw this version, since I was sending one `FactorSource` param instead. Still, updated it to send just `FactorSourceID` since it will be lighter, and we can get its kind with the function you mention.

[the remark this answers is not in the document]
GhenadieVP: aah, I reviewed the first commit, to filter out all of the files that were modified by bump version. I see that there are follow up changes.

GhenadieVP (approved):

PR #342, opened 2025-01-14
Finish early mono sign when skipping factor

micbakos-rdx:
Updated mono sign, to return no request when there are no remaining per transaction requests. When a user skips a factor source, the signature collector will iterate to the next factor source. There if there are no more remaining still to be signed as valid transactions, the iterator continues. This ends up ending the collector's signing process, with an unsuccessful outcome.

* Updated the tests to incorporate skipping of a factor source instead of just failing one.
* Added some missing kotlin extensions needed in host to sign, with a mnemonic, multiple per transaction signables at once and also an extension to derive factor source ids. Both were required for the integration with the host

### DIFF
--- a/crates/app/signing-traits/src/testing/simulated_user.rs
+++ b/crates/app/signing-traits/src/testing/simulated_user.rs
@@ -19,6 +19,8 @@ pub struct SimulatedUser<S: Signable> {
     mode: SimulatedUserMode,
     /// `None` means never failures
     failures: Option<SimulatedFailures>,
+    /// `None` means no skips
+    skips: Option<SimulatedSkips>,
 }
 
 impl<S: Signable> SimulatedUser<S> {
@@ -27,19 +29,22 @@ impl<S: Signable> SimulatedUser<S> {
             + 'static,
         mode: SimulatedUserMode,
         failures: impl Into<Option<SimulatedFailures>>,
+        skips: impl Into<Option<SimulatedSkips>>,
     ) -> Self {
         Self {
             spy_on_request: Arc::new(spy_on_request),
             mode,
             failures: failures.into(),
+            skips: skips.into(),
         }
     }
 
     pub fn new(
         mode: SimulatedUserMode,
         failures: impl Into<Option<SimulatedFailures>>,
+        skips: impl Into<Option<SimulatedSkips>>,
     ) -> Self {
-        Self::with_spy(|_, _| {}, mode, failures)
+        Self::with_spy(|_, _| {}, mode, failures, skips)
     }
 }
 
@@ -48,6 +53,7 @@ pub struct SimulatedFailures {
     /// Set of FactorSources which should always fail.
     simulated_failures: IndexSet<FactorSourceIDFromHash>,
 }
+
 impl SimulatedFailures {
     pub fn with_details(
         simulated_failures: IndexSet<FactorSourceIDFromHash>,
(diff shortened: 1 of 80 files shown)


crates/app/signing/src/collector/signatures_collector.rs
```diff
@@ -253,30 +253,33 @@ impl<S: Signable> SignaturesCollector<S> {
...
 
-            // Report the results back to the collector
-            self.process_batch_response(response);
+                // Report the results back to the collector
+                self.process_batch_response(response);
 
-            if self.continuation() == FinishEarly {
-                break;
+                if self.continuation() == FinishEarly {
+                    break;
```
matiasbzurovski, line 281: shouldn't we also check this when there is no request for the given `factor_source_id`. This, when there are no transactions to sign for the given factor source

crates/app/signing-traits/src/testing/simulated_user.rs
```diff
@@ -99,31 +131,37 @@ impl SimulatedUserMode {
...
 
     pub fn prudent_with_failures(
         simulated_failures: SimulatedFailures,
     ) -> Self {
-        Self::new(SimulatedUserMode::Prudent, simulated_failures)
+        Self::new(SimulatedUserMode::Prudent, simulated_failures, None)
+    }
+
+    pub fn prudent_with_skips(simulated_skips: SimulatedSkips) -> Self {
+        Self::new(SimulatedUserMode::Prudent, None, simulated_skips)
```
CyonAlexRDX, line 148: `Prudent` means "never skips". So this looks wrong. Better let this be ctor of `Lazy`. And update definition of Lazy:

```diff
-    /// Emulation of a "lazy" user, that skips signing with as many factor
-    /// sources as possible.
+    /// Emulation of a "lazy" user, that skips signing with as some factor
+    /// sources, depending on `Laziness`.
    Lazy(Laziness),
```

crates/system/os/signing/src/sargon_os_signing.rs
```diff
@@ -418,6 +504,7 @@ mod test {
 
     enum SigningFailure {
         NeglectedFactorSources(Vec<FactorSourceIDFromHash>),
+        SkippingFactorSources(Vec<FactorSourceIDFromHash>),
```
Sajjon, line 507: We used to define "Skip(ped)" a subset of "Neglected" - so is this needed?

Sajjon: You can probably use Neglected variants

crates/app/signing-traits/src/testing/simulated_user.rs
micbakos-rdx: Yes, but `Laziness` has no option to sign all the rest of the factor sources except the ones defined in `simulated_skips`. This is what I wanted to achieve. 

Lazyness describes a user that either skips all or signs the minimum.

crates/system/os/signing/src/sargon_os_signing.rs
micbakos-rdx: I need to rename `NeglectedFactorSources` to `FailingFactorSources`

crates/app/signing-traits/src/testing/simulated_user.rs
Sajjon: Yes so we update Laziness enum with a third variant..?

SkipsSpecific(IndexSet<_>)

micbakos-rdx:
@matiasbzurovski @CyonAlexRDX 

Latest changes:

1. Added to user mode a user that skips specific factors.
2. Introduced a new error for when the user fails to sign since too many factors neglected.
3. Added `--locked` to cargo ndk.

GhenadieVP (approved):

matiasbzurovski (approved):

PR #343, opened 2025-01-15
Security Shield builder API updates

sergiupuhalschi-rdx:
Updated the API of `SecurityShieldBuilder` as follows:

- Added convenience function to remove all primary override factors
- Added `TimePeriod` and `TimePeriodUnit` to represent in the UI the time until recovery auto-confirmation
- Updated the order of the checks in the `validate` function to better accommodate the UI requirements. For example, in the first step in the UI, after selecting enough factor sources to build a good shield and auto-assign them to recovery and confirmation roles, and removing all but one primary threshold factor sources and the authentication signing factor, the UI needs to reflect that an authentication signing factor source has to be added before continuing to the recovery / confirmation step.


crates/core/utils/src/constants.rs
```diff
@@ -31,6 +31,18 @@ pub const POLLING_DELAY_INCREMENT_IN_SECONDS: u64 = 1;
...
 pub const MINUTES_PER_DAY: u32 = 24 * 60;
 
+/// Number of weeks per year.
+pub const WEEKS_PER_YEAR: u16 = 52;
+
+/// Number of days per week.
+pub const DAYS_PER_WEEK: u16 = 7;
+
+/// Number of days per year.
+pub const DAYS_PER_YEAR: u16 = DAYS_PER_WEEK * WEEKS_PER_YEAR;
```
Sajjon, line 41: Do we want to skip 1 day? Since this is 364.

sergiupuhalschi-rdx: Right, I'm not sure. Maybe it's best to make it 365 with the downside of being unable to express a year in a fixed number of weeks, but I guess it's fine since we don't need that now anyway.

crates/profile/models/security-structures/src/threshold.rs
```diff
@@ -23,6 +23,13 @@ impl Threshold {
             Threshold::Specific(value) => *value,
         }
     }
+
+    /// Returns the selectable values for a threshold.
+    pub fn values(max_threshold: u8) -> Vec<Threshold> {
```
CyonAlexRDX, line 28: we *can* do this, but *should we*? This genarates `9999` elements and then is UniFFI exported. probably not a huge cost but still, it is 10k elements which are uniffi converted, when you could use 2 elements instead by use of a range:

```rust
#[derive(uniffi::Record)]
pub struct Range16 {
   pub lower: u16,
   pub upper: u16
}
```

And then in Swift Sargon or Kotlin Sargon map the range to `Vec<u16>.

Futhermore, current impl *allows for* 10000 elements of `Threshold::All` which is weird...

crates/core/utils/src/constants.rs
```diff
@@ -31,6 +31,15 @@ pub const POLLING_DELAY_INCREMENT_IN_SECONDS: u64 = 1;
...
 pub const MINUTES_PER_DAY: u32 = 24 * 60;
 
+/// Number of days per week.
+pub const DAYS_PER_WEEK: u16 = 7;
+
+/// Number of days per year.
+pub const DAYS_PER_YEAR: u16 = 365;
+
+/// Maximum number of units (days, weeks, years) for the security structure recovery confirmation fallback period.
+pub const MAX_RECOVERY_CONFIRMATION_FALLBACK_PERIOD_UNITS: u16 = 9999;
```
CyonAlexRDX, line 41: hmm this seems too specific to be on `core` crate, ought to into `profile-security-structures` crate ?

crates/profile/models/security-structures/src/threshold.rs
sergiupuhalschi-rdx: Are you confusing this with the values for the emergency fallback where we have the upper bound of 9999? For the threshold, the array length would be the number of primary threshold factors the user has chosen. Also, we could use `IndexSet` instead of `Vec` to ensure the values are unique, but we'll still have to convert to Vec when exporting it through uniffi, so I'm not sure if it's worth making this change since we've covered this with a unit test.

crates/profile/models/security-structures/src/time_period_unit.rs
```diff
@@ -0,0 +1,75 @@
...
+        TimePeriodUnit::Days
+    }
+
+    fn sample_other() -> Self {
+        TimePeriodUnit::Weeks
+    }
+}
+
+impl TimePeriodUnit {
+    pub fn values(&self) -> Vec<u16> {
```
sergiupuhalschi-rdx, line 24: @CyonAlexRDX Here we have an array with 9999 `u16` elements which crosses the uniffi boundary, do you think the performance cost of this is too high compared to the benefit of having this logic in a single place?

CyonAlexRDX: right, so here I would use a Range instead.

sergiupuhalschi-rdx: Right, currently it brings value only in terms of display readiness (there's nothing else to do in UI, just render the values) and in terms of the lower and upper bounds, and if we ever were to display non-consecutive values. But it's ok to have this defined in hosts as it's very UI specific.

danvleju-rdx (approved):

GhenadieVP (approved):

PR #344, opened 2025-01-16
Use `--locked` for `cargo` commands && change `Cargo.toml` to use `crates.io` 

CyonAlexRDX:
Also rename github worksflows **file names** which were confusing.

For more context [see Slack](https://rdxworks.slack.com/archives/C031A0V1A1W/p1736850099864229).

I think I've updated **all** `cargo` commands, but please review that, especially @micbakos-rdx for the gradle files.


.github/workflows/test.yml
```diff
@@ -137,7 +137,7 @@ jobs:
           components: rustfmt
 
       - name: Check formatting
-        run: cargo fmt --check
+        run: cargo fmt --locked --check --all
```
CyonAlexRDX, line 140: (not sure `--all` flag is needed, but I like it for symmetry)

Cargo.toml
CyonAlexRDX, line 1: I was unable to use version for Scrypto crates since `radix-engine-toolkit-common` was not on crates.io

crates/crypto/hd/src/bip39/mnemonic.rs
```diff
@@ -53,7 +53,7 @@ impl Mnemonic {
         let language = internal.language();
 
         let words = internal
-            .word_iter()
+            .words()
```
CyonAlexRDX, line 56: `word_iter` method was deprecated.

scripts/ios/build-sargon.sh
CyonAlexRDX, line 1: Witness:
- [x] I've added `--locked` to all three `cargo build` commands (iOS device, iOS sim, macOS Apple Silicon)
- [x] I've added `--locked` to `cargo run sargon-bindgen` for iOS

jvm/buildSrc/src/main/java/com/radixdlt/cargo/desktop/CargoDesktopPlugin.kt
CyonAlexRDX, line 1: Witness:
- [x] I've added `--locked` to the `cargo build` command for Kotlin
- [x] I've added `--locked` to `cargo run sargon-bindgen` for Kotlin (in file `jvm/sargon-android/build.gradle.kts` below)

.github/workflows/test.yml
CyonAlexRDX, line 1: Witness - I've added `--locked` to:
- [x] `cargo check`
- [x] `cargo fmt`
- [x] `cargo clippy`
- [x] `cargo run nextest`

.github/workflows/release-swift-ios.yml
CyonAlexRDX, line 1: Witness - iOS workflow does not perform any `cargo` command, it only calls `scripts/ios/build-sargon.sh` which **has been modified**.

.github/workflows/release-kotlin-android.yml
CyonAlexRDX, line 1: Witness - Android workflow does not perform any `cargo` command, it only calls **gradle** which **has been modified**. (both `CargoDesktopPlugkin.kt`). Hmm @micbakos-rdx there is no corresponding "CargoAndroidPlugin.kt" right? We use `CargoDesktopPlugkin.kt` for Android too right?

.github/workflows/test.yml
```diff
@@ -137,7 +137,7 @@ jobs:
           components: rustfmt
 
       - name: Check formatting
-        run: cargo fmt --check
+        run: cargo fmt --check --all
```
CyonAlexRDX, line 140: `--locked` did not work for `cargo fmt`, so skipped that, but added `--all` for symmetry (I think I've previously asserted that it was not needed?)

giannis-rdx (approved):

sergiupuhalschi-rdx (approved):

GhenadieVP (approved):

PR #345, opened 2025-01-16
Apply Shield Wallet Interaction Part 1

CyonAlexRDX:
[Jira ABW-4056](https://radixdlt.atlassian.net/browse/ABW-4056)

> [!NOTE]
> Part 1 of N

The first of two (or more) PRs for producing manifests which updates the security config of entities - i.e. applying a security shield (factor instances) to an entity.

This PR specifically only adds support for securifying an unsecurified account.

# Other Changes
## `UnsecurifiedPersona` && `UnsecurifiedAccount` types
Refactored `UnsecurifiedEntity` to be generic over entity => `AbstractUnsecurifiedEntity` with type aliases `AnyUnsecurifiedEntity`, `UnsecurifiedAccount` and `UnsecurifiedPersona`-- acting just like their "securified cousins" (`AbstractSecurifiedEntity`, `AnySecurifiedEntity`, `SecurifiedAccount` and `SecurifiedPersona`)

## `EntitiesOnNetwork`
Refactor the five fields of `NextDerivationEntityIndexProfileAnalyzingAssigner ` into the type `EntitiesOnNetwork` and let `NextDerivationEntityIndexProfileAnalyzingAssigner` iml `Defef<Target = EntitiesOnNetwork>`. The `EntitiesOnNetwork` is now also returned from `apply_security_shield_with_id_to_entities` which we make use of when we call that function from the new API: `make_interaction_for_applying_security_shield` and immediately get a structured object of which entites are securified and which are not (all having a provisional config).

## `IndexSet` -> `IdentifiedVecOf`
I've changed to use `IdentifiedVecOf<Sec/Unse Entity>` instead of `IndexSet` which is an important change preventing bugs. Why? Because `IndexSet` would allow for two copies of the same account if one single deeply nested value has changed, e.g. a `hidden` flag being present vs non-present - since `IndexSet` is Hash-based. So the same `AccountAddress` would be present twice. This should never be allowed to happen. And could potentially lead to dangerous bugs. And we automatically protect against this by using `IdentifiedVecOf`!

# Future Work
Future PR will add the more complex situation where we must produce 6 different kinds of manifest based on:

```rust
pub enum TransactionManifestApplySecurityShieldKindSelector {
    /// (Primary + Recovery + Confirmation)
    PrimaryAndRecoveryWithExplicitConfirmation,

    /// (Primary + Recovery + Time)
    PrimaryAndRecoveryWithTimedAutoConfirm,

    /// (Primary + Confirmation)
    PrimaryAndExplicitConfirmation,

    /// (Primary + Time) 
    PrimaryWithTimedAutoConfirm,

    /// (Recovery + Confirmation)
    RecoveryAndExplicitConfirmation,

    /// (Recovery + Time)
    RecoveryWithTimedAutoConfirm,
}
```


crates/system/os/factors/src/apply_security_shield/wallet_interaction_apply_security_shield.rs
```diff
@@ -0,0 +1,245 @@
...
+            ///  - `primary_and_confirmation`
+            ///  - `recovery_and_confirmation`
+            #[derive(Debug, Clone, PartialEq, Eq)]
+            pub struct $name {
+                $(
+                    #[doc = $expr]
+                )*
+                $entity_name: $entity_type,
+
+                /// A tripple of manifests for exercising all roles.
```
CyonAlexRDX, line 144: I think we need those four manifests?

CyonAlexRDX, line 144: I think we need those four manifests?

CyonAlexRDX, line 144: I think we need those four manifests?

crates/transaction/manifests/src/modify_manifest.rs
```diff
@@ -176,6 +179,7 @@ impl TransactionManifestModifying for TransactionManifest {
 
 pub enum InstructionPosition {
     First,
+    Last,
```
CyonAlexRDX, line 182: Not used actually, so we might remove it

crates/transaction/manifests/src/manifests_security_shield.rs
```diff
@@ -0,0 +1,620 @@
...
+    fn apply_security_shield_for_securified_entity(
+        securified_entity: AnySecurifiedEntity,
+        input: TransactionManifestApplySecurityShieldSecurifiedInput,
+    ) -> Result<Self>;
+
+    /// If `top_up_amount` is `None` the `XRD_TO_AC_VAULT_FIRST_TOP_UP` will be used.
+    /// We allow to pass amount so that we can top of with more or less based on
+    /// token balance of `payer` and current balance of the access controller (when
+    /// we use this method for securified entities.)
+    fn modify_manifest_add_withdraw_of_xrd_for_access_controller_xrd_vault_top_up_payed_by_account(
```
CyonAlexRDX, line 177: @0xOmarA @GhenadieVP N.B. It is my intention we call this both with manifest which Securified an unsecurified entity **and for manifests updating factors for already Securified entity**.

crates/transaction/manifests/src/manifests_security_shield.rs
```diff
@@ -0,0 +1,620 @@
...
+        payer: Account,
+        // TODO: remove `entity_applying_shield`, this should be read out from the manifest in a throwing function, `manifest.get_address_of_entity_applying_shield()` or similar which Omar need to provide us with, oh well we need the account here, so elsewhere, in SargonOS where we have access to Profile we would call `manifest.get_address_of_entity_applying_shield` and then lookup the entity.
```
GhenadieVP, line 179: May be possible based on manifest static analysis, but it should be completely acceptable to receive account as input, after all modify should happen on Wallet action which has the context of the account.

crates/transaction/manifests/src/manifests_security_shield.rs
```diff
@@ -0,0 +1,620 @@
...
+    ) -> ManifestBuilder {
+        let rola_key_hash = PublicKeyHash::hash(
+            authentication_signing_factor_instance.public_key(),
+        );
+        let owner_key_hashes = vec![rola_key_hash];
+        Self::set_owner_keys_hashes_on_builder(
+            &entity_address,
+            owner_key_hashes,
+            builder,
+        )
```
GhenadieVP, line 293: I believe we should append the rola_key_hash, not replace existing owner_key_hashes with just rola.

CyonAlexRDX: @GhenadieVP Matt said to **set** it: https://rdxworks.slack.com/archives/C031A0V1A1W/p1734097688450489?thread_ts=1733953261.863269&cid=C031A0V1A1W

CyonAlexRDX: @GhenadieVP Matt said to **set** it: https://rdxworks.slack.com/archives/C031A0V1A1W/p1734097688450489?thread_ts=1733953261.863269&cid=C031A0V1A1W

GhenadieVP: Ok, although just reading the message I am not so sure, we can clarify in standup.

crates/transaction/manifests/src/manifests_security_shield.rs
```diff
@@ -0,0 +1,620 @@
...
+        ACCESS_CONTROLLER_CREATE_PROOF_IDENT,
+    },
+    account::AccountSecurifyManifestInput as ScryptoAccountSecurifyManifestInput,
+};
+use radix_transactions::prelude::ManifestBuilder;
+
+use crate::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TransactionManifestApplySecurityShieldKindSelector {
```
CyonAlexRDX, line 22: not used by this PR, but will be used in the follow up PR

GhenadieVP (approved):

PR #347, opened 2025-01-20
Allow building a weak shield

sergiupuhalschi-rdx:
This PR allows the creation of shields with less secure factor combinations. Added `status` function to represent the factors selection state:
`Strong` - meets all security combination rules.
`Weak` - breaks some combination rules but still can be built.
`Invalid` - lacks at least one factor per role and cannot be built.

Breaking changes:
- Renamed `SecurityShieldBuilderInvalidReason` to `SecurityShieldBuilderRuleViolation`.
- Replaced the `validate()` function with the new status() function.
- Updated the return type of `add_security_structure_of_factor_source_ids` from `Result<bool>` to `Result<()>`.


crates/profile/models/security-structures/src/roles_matrices_structures/security_shield_builder_rule_violation_reason.rs
```diff
@@ -3,15 +3,15 @@ use crate::prelude::*;
 pub trait AsShieldBuilderViolation {
     fn as_shield_validation(
         &self,
-    ) -> Option<SecurityShieldBuilderInvalidReason>;
+    ) -> Option<SecurityShieldBuilderRuleViolationReason>;
```
Sajjon, line 6: RuleViolationReason is a bit of a tautology I think, I would drop the `Reason` suffix

danvleju-rdx (approved):
Very nice!

PR #350, opened 2025-01-22
[ABW-4135] Export method to derive public keys

matiasbzurovski:
JIRA ticket: [ABW-4135](https://radixdlt.atlassian.net/browse/ABW-4138)

## Changelog
- Exports API to allow deriving public keys from a given [source](https://github.com/radixdlt/sargon/blob/5877c7518e4e475b75c1bbe0de7074a092beedd9/crates/system/os/derive-public-keys/src/support/source.rs). Such method will be used hosts to perform Account Recovery scan.
- Adds some extensions on Swift Sargon to allow integration `HostInteractor`.
- Updates references of `bdfs` to `main_bdfs`.

## Notes
A lot of the file diff comes from the latest change mentioned, so you may want to review that commit independently. You can review the 16 previous commits [here](https://github.com/radixdlt/sargon/pull/350/files/9a3de4587e4e1aea0cec0be6e8ba2ec161f0d4e9..69a1328346c1ef95a08f195d6164a04d8c045ddc).

[ABW-4135]: https://radixdlt.atlassian.net/browse/ABW-4135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/sargon_os_derive_public_keys.rs
```diff
@@ -0,0 +1,27 @@
...
+                source.into_internal(),
+            )
+            .await
+            .into_iter_result()
+    }
+}
+
+#[derive(Clone, PartialEq, InternalConversion, uniffi::Enum)]
+pub enum DerivePublicKeysSource {
+    MnemonicWithPassphrase(MnemonicWithPassphrase),
```
micbakos-rdx, line 24: Can we rename it to seed phrase? This enum variant when translated to kotlin fails the build:

```
 Type mismatch: inferred type is com.radixdlt.sargon.MnemonicWithPassphrase but com.radixdlt.sargon.DerivePublicKeysSource.MnemonicWithPassphrase was expected
 ```

Comment (matiasbzurovski) [reply crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/sargon_os_derive_public_keys.rs:24]: sure

Comment (micbakos-rdx) [reply crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/sargon_os_derive_public_keys.rs:24]: I think the problem with kotlin in this scenario is that the inner value has the same name with the variant. This is how it compiles into:

```
sealed class DerivePublicKeysSource {
    
    data class MnemonicWithPassphrase(
        val v1: MnemonicWithPassphrase) : DerivePublicKeysSource() {
        companion object
    }
    
    data class FactorSource(
        val v1: FactorSourceIdFromHash) : DerivePublicKeysSource() {
        companion object
    }
    

    
    companion object
}
```
You can see `v1` maybe infers to be not a `MnemonicWithPassphrase` but `DerivePublicKeysSource.MnemonicWithPassphrase` since it has no distinction

File: crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/sargon_os_personas.rs [hunk 2]
```diff
@@ -0,0 +1,46 @@
...
+            )
+            .await
+            .into_result()
+    }
+
+    /// Create a new Persona and adds it to the active Profile.
+    ///
+    /// # Emits Event
+    /// Emits `Event::ProfileModified { change: EventProfileModified::PersonaAdded }`
+    pub async fn create_and_save_new_persona_with_bdfs(
```
micbakos-rdx, line 31: Maybe rename it to `_main_bdfs`

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/entities/sargon_os_accounts.rs
```diff
@@ -115,7 +115,27 @@ impl SargonOS {
...
+            )
+            .await
+            .into_result()
+    }
+
+    /// Create a new Account and adds it to the active Profile.
+    ///
+    /// # Emits Event
+    /// Emits `Event::ProfileModified { change: EventProfileModified::AccountAdded }`
+    pub async fn create_and_save_new_account_with_bdfs(
```
micbakos-rdx, line 138: Maybe rename it to `_main_bdfs`

micbakos-rdx (approved):
Interesting what you did in tests using double `_` to "group" them.

GhenadieVP (approved):
LGTM, nice work!

PR #355, opened 2025-01-23
Apply Shield Wallet Interaction Part 2

CyonAlexRDX:
[Jira ABW-4056](https://radixdlt.atlassian.net/browse/ABW-4056)

> [!CAUTION]
> Sargon breaking change in Profile format - only Wallet team devs affected since models are MFA models end users have not been able to produce yet.
> renamed "auto_confirm" to "time_based_confirmation" instead since "auto" was indeed very misleading! Since we MUST call confirm in a manifest and submit to the network. Is does not happen "automatically"! I've also changed then integer to `TimePeriod`

> [!NOTE]
> Part 2 of N
> [Part 1 merged into main](https://github.com/radixdlt/sargon/pull/345)

# Update Factors of Securified Entities

For securified entities we need many different manifests to sign depending on which roles the user is able to exercise at the moment of applying the shield.

Host will display the variant which is exercising all roles with confirmation rule using quick confirm - but if user slides to sign, Sargon will under the hood try to sign all (maybe less if we can finish early). If she does not have access to some factor belonging to Primary role she can skip signing with that factor and all manifest variants out of those six which requires Primary will be invalid. We can still withdraw XRD from AccessController XRD vault so we can continue trying sign manifests requiring Recovery role. Just that we cannot top up the XRD vault of the access controller. If she manages to sign with the required factors of Recovery role we proceed trying to sign with factors of ConfirmationRole. Which can fail too, then we use time based fallback, if user was unable to exercise Primary Role and Reocery role we fail. 

This PR adds logic for creating all flavours:

```rust
pub enum RolesExercisableInTransactionManifestCombination {
    InitiateWithPrimaryCompleteWithRecovery,
    InitiateWithPrimaryCompleteWithConfirmation,
    InitiateWithRecoveryCompleteWithConfirmation,
    InitiateWithRecoveryDelayedCompletion,
    /// ‼️ REQUIRES "Dugong" ‼️
    InitiateWithPrimaryDelayedCompletion,
}
```

# Other Changes
- Changed from `number_of_days_until_auto_confirmation: u16` -> `time_until_delayed_confirmation_is_callable: TimePeriod` in Profile - two important changes since auto confirm was misleading and since u16 representation in days prevented us from ever changing to hours/minutes if we ever wanted that. The TimePeriod was a perfect type and is made Serde
- moved incorrectly placed JSON files into `/models` from `/vectors`- a vector is a list of small object with input and expected output - often used by cryptographic operations
- Swift: fix deprecation warnings


crates/transaction/manifests/src/manifests_security_shield/access_controller_factors_and_time_input.rs
```diff
@@ -0,0 +1,100 @@
...
+use radix_engine_interface::blueprints::access_controller::{
+    AccessControllerInitiateRecoveryAsPrimaryInput as ScryptoAccessControllerInitiateRecoveryAsPrimaryInput,
+    AccessControllerInitiateRecoveryAsRecoveryInput as ScryptoAccessControllerInitiateRecoveryAsRecoveryInput,
+    AccessControllerQuickConfirmPrimaryRoleRecoveryProposalInput as ScryptoAccessControllerQuickConfirmPrimaryRoleRecoveryProposalInput,
+    AccessControllerQuickConfirmRecoveryRoleRecoveryProposalInput as ScryptoAccessControllerQuickConfirmRecoveryRoleRecoveryProposalInput,
+    AccessControllerTimedConfirmRecoveryInput as ScryptoAccessControllerTimedConfirmRecoveryInput,
+};
+
+#[derive(Debug, Clone)]
+pub struct AccessControllerFactorsAndTimeInput {
```
Sajjon, line 11: shared type to create:
- AccessControllerInitiateRecoveryAsPrimaryInput
- AccessControllerInitiateRecoveryAsRecoveryInput
- AccessControllerQuickConfirmPrimaryRoleRecoveryProposalInput
- AccessControllerQuickConfirmRecoveryRoleRecoveryProposalInput
- AccessControllerTimedConfirmRecoveryInput

Input to `call_method` instruction, using `SecurityStructureOfFactorInstances` - which is `Into<RuleSet>` (and time u32).

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
```diff
@@ -0,0 +1,95 @@
...
+    ) -> Result<Self> {
+        let TransactionManifestApplySecurityShieldSecurifiedInput {
+            security_structure_of_factor_instances,
+            apply_shield_manifest_kind: kind,
+        } = input.clone();
+
+        let entity_address = securified_entity.entity.address();
+
+        // ACCESS_CONTROLLER_CREATE_PROOF_IDENT
+        let mut builder = TransactionManifest::produce_owner_badge(
```
Sajjon, line 29: This does:
```
builder.call_method(
    access_controller_address.scrypto(),
    ACCESS_CONTROLLER_CREATE_PROOF_IDENT,
    (),
);
```

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
```diff
@@ -0,0 +1,95 @@
...
+            .securified_entity_control
+            .access_controller_address;
+
+        let factors_and_time_input = &AccessControllerFactorsAndTimeInput::new(
+            &security_structure_of_factor_instances,
+        );
+
+        // INITIATE RECOVERY
+        let (init_method, init_input) =
+            kind.input_for_initialization(factors_and_time_input);
```
Sajjon, line 44: Depending on selector of which of the six `TransactionManifest` flavours as selected by `kind: TransactionManifestApplySecurityShieldKind` we use different roles for **initialization of recovery** (and the correct concrete scrypto "Input"-type (Box dyn-ed))

For details see `transaction_manifest_apply_security_shield_kind.rs` which was added by this PR

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
```diff
@@ -0,0 +1,95 @@
...
+        );
+
+        // CONFIRM RECOVERY
+        // TODO: for timed, should we really call it here, now? Should
+        // we not call it AFTER the time has elapsed???
+        let (confirm_method, confirm_input) =
+            kind.input_for_confirm(factors_and_time_input);
+        builder = builder.call_method(
+            access_controller_address.scrypto(),
+            confirm_method,
```
Sajjon, line 58: Depending on selector of which of the six `TransactionManifest` flavours as selected by `kind: TransactionManifestApplySecurityShieldKind` we use different roles for **confirmation of recovery** (and the correct concrete scrypto "Input"-type (Box dyn-ed))

For details see `transaction_manifest_apply_security_shield_kind.rs` which was added by this PR.

> [!IMPORTANT]
> Im almost certain this is wrong... I'm calling `ACCESS_CONTROLLER_TIMED_CONFIRM_RECOVERY_IDENT` if we cannot quick confirm. That should not be done NOW right? bot rather after the time has elapsed?

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
```diff
@@ -0,0 +1,95 @@
...
+            (confirm_input.deref(),),
+        );
+
+        // Set Rola Key
+        let should_set_rola_key = security_structure_of_factor_instances
+            .authentication_signing_factor_instance
+            != securified_entity
```
Sajjon, line 65: dont set ROLA if it is unchanged

crates/transaction/manifests/src/manifests_security_shield/transaction_manifest_apply_security_shield_kind.rs
```diff
@@ -0,0 +1,134 @@
...
+    PrimaryAndRecoveryWithExplicitConfirmation,
+
+    /// (Primary Recovery Time)
+    PrimaryAndRecoveryWithTimedAutoConfirm,
+
+    /// (Primary Confirmation)
+    PrimaryAndExplicitConfirmation,
+
+    /// (Primary Time) ‼️ REQUIRES "Dugong" ‼️
+    PrimaryWithTimedAutoConfirm,
```
Sajjon, line 28: @GhenadieVP I think Matt said we would have `Dugong` available. So we will remove this comment once Dugong is live. Maybe I should comment out `PrimaryWithTimedAutoConfirm` for now...

crates/transaction/manifests/src/manifests_security_shield/transaction_manifest_apply_security_shield_kind.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    pub fn can_set_rola_key(&self) -> bool {
+        self.can_exercise_primary_role()
+    }
+
+    pub fn input_for_initialization(
+        &self,
+        factors_and_time: &AccessControllerFactorsAndTimeInput,
+    ) -> (&'static str, Box<dyn CallMethodInput>) {
```
Sajjon, line 86: @dhedey naivly I tried returning `(&'static str, impl ResolvableArguments)` which did not compile. Then I tried `Box<dyn ResolvableArguments>` which compiled here, but did not compile when I tried passing it to `call_method` directly, so I tried `.deref()` on the box to get the "inner" `dyn ResolvableArguments` which did not compile... bah. But when I introduced this "dummy trait" (see above):

```rust
// Trickery to allow `Box<dyn ResolvableArguments>` - which is not allowed it seems,
// but this solves it, since in Scrypto they impl ResolvableArguments for ManifestEncode + ManifestSborTuple
pub trait CallMethodInput: ManifestEncode + ManifestSborTuple {}
impl<T: ManifestEncode + ManifestSborTuple> CallMethodInput for T {}
```

then everything works. It is just 2 lines of code so fine for now. But maybe you have ideas on how it can be removed? :)

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
Sajjon: @GhenadieVP indeed this must be wrong, we must ofc wait the required time before we can call `ACCESS_CONTROLLER_TIMED_CONFIRM_RECOVERY_IDENT` on the access controller - **or CANCEL** `ACCESS_CONTROLLER_CANCEL_PRIMARY_ROLE_RECOVERY_PROPOSAL_IDENT`.

I cant remember us talking about this at all so far? So probably what we wanna do is, similar to scheduling "local notifcations" in host for queued transactions (when batching), we will wanna do the same for the required time user has to wait until _confirm timed recovery_ is possible. The local notification fires, and user opens Radix Wallet and we host should present user with option to confirm/cancel. Do we have design for this yet??? Lets say user presses "confirm" we should build the small manifest which hmm, which instructions are needed in that manifest  @0xOmarA / @dhedey / @iamyulong  ? Im looking at https://docs.radixdlt.com/docs/access-controller#timed-confirm-recovery and its says:

>  Anybody can now confirm the Recovery (R) role’s recovery proposal.

Anybody? As in no specific signer needs to sign the manifest? And the manifest is (not counting "lock fee") a single instruction calling `ACCESS_CONTROLLER_TIMED_CONFIRM_RECOVERY_IDENT` on the access controller with the **new - SAME** RuleSet as set in `SCRYPTO_ACCESS_CONTROLLER_QUICK_CONFIRM_RECOVERY_ROLE_RECOVERY_PROPOSAL_IDENT`/`SCRYPTO_ACCESS_CONTROLLER_QUICK_CONFIRM_PRIMARY_ROLE_RECOVERY_PROPOSAL_IDENT` ?

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
```diff
@@ -0,0 +1,95 @@
...
+        // ACCESS_CONTROLLER_CREATE_PROOF_IDENT
+        let mut builder = TransactionManifest::produce_owner_badge(
+            ScryptoTransactionManifestBuilder::new(),
+            &securified_entity.entity,
+        );
+
+        let access_controller_address = securified_entity
+            .securified_entity_control
+            .access_controller_address;
+
```
CyonAlexRDX, line 37: I have forgotten to add the withdrawal of XRD from AccessControllera XRD vault.

I’m adding it now

crates/transaction/manifests/src/manifests_security_shield/transaction_manifest_owner_badge_producing.rs
```diff
@@ -0,0 +1,65 @@
...
+        let mut builder = builder;
+        let owner = owner.borrow();
+
+        let is_account = owner.is_account_entity();
+
+        match owner.security_state() {
+            EntitySecurityState::Securified { value } => {
+                builder = builder.call_method(
+                    value.access_controller_address.scrypto(),
+                    ACCESS_CONTROLLER_CREATE_PROOF_IDENT,
```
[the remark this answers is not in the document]
Sajjon: Maybe this question does not make any sense... but I thought we always needed to invoke an instruction which triggers requirement to "auth". but maybe the INITIATE recovery methods will do that...?

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
Sajjon: Right we need to add the locking of fee against XRD vault of AccessController **later** (like the top up) - since we dont know how big a fee to lock. Ive added `modify_manifest_add_lock_fee_against_xrd_vault_of_access_controller`. Take a look at it https://github.com/radixdlt/sargon/pull/355/commits/4d8fc6df92d4f69936cbd557a637a0f7842de592

crates/transaction/manifests/src/manifests_security_shield/lock_fee_against_xrd_vault_of_access_controller.rs
```diff
@@ -0,0 +1,52 @@
...
+        fee: Decimal192,
+        // TODO: remove `entity_applying_shield`, this should be read out from the manifest in a throwing function, `manifest.get_address_of_entity_applying_shield()` or similar which Omar need to provide us with, oh well we need the account here, so elsewhere, in SargonOS where we have access to Profile we would call `manifest.get_address_of_entity_applying_shield` and then lookup the entity.
+        entity_applying_shield: AnySecurifiedEntity,
+    ) -> TransactionManifest {
+        let mut builder = ManifestBuilder::with_manifest(manifest);
+        let access_controller_address = entity_applying_shield
+            .securified_entity_control
+            .access_controller_address;
+        // Lock fee against XRD vault of the access controller
+        builder = builder.call_method(
```
Sajjon, line 39: @0xOmarA here I'm appending this instruction **last**. Is that OK? Or must this be **prepended**, index 0 of the manifest? if so I need to do a lot of trickery to do that. Since I cannot create the instruction in rust code "manually" and since calling `buider.call_method` will put it last...

@GhenadieVP alternatively we just lock a large amount. Which ""should work"" for most shields... and if we do that we can lock fee in `apply_security_shield_for_securified_entity` at index 0  https://github.com/radixdlt/sargon/blob/ac/wallet_interaction_applying_shield_abw-4056_part2/crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs#L37

crates/transaction/manifests/src/manifests_security_shield/confirm_timed_recovery.rs
```diff
@@ -0,0 +1,27 @@
...
+
+pub trait TransactionManifestConfirmTimedRecovery {
+    /// TBD
+    /// TODO: Figure out how to do this...
+    /// since `AccessControllerTimedConfirmRecoveryInput` need the input of
+    /// the factors and time which was used to start recovery - which could not
+    /// be quick confirmed. We need to figure out how we best represent this
+    /// in Profile. Perhaps a new variant on ProvisionalSecurityConfig? Something
+    /// like:
+    /// `WaitingForTimedRecovery(SecurityStructureOfFactorInstances)`
```
Sajjon, line 16: @GhenadieVP good thing we kept `ProvisionalSecurifiedConfig` an enum! I think we need a second variant - `WaitingForTimedRecovery(SecurityStructureOfFactorInstances)`

crates/profile/models/security-structures/src/roles_matrices_structures/matrices/abstract_matrix_builder_or_built.rs
```diff
@@ -39,13 +39,14 @@ pub struct AbstractMatrixBuilderOrBuilt<
     pub(crate) confirmation_role:
         AbstractRoleBuilderOrBuilt<{ ROLE_CONFIRMATION }, MODE_OF_ROLE, FACTOR>,
 
-    pub number_of_days_until_auto_confirm: u16,
+    pub number_of_days_until_timed_confirmation_is_callable: u16,
```
Sajjon, line 42: Breaking profile change! But ok since only Radix Wallet team devs are affected.

`numberOfDaysUntilAutoConfirm` made it sound like confirmation would happen automagically. Which perhaps even I believed they would when I named this variable...

crates/profile/models/security-structures/src/roles_matrices_structures/matrices/builder/error.rs
```diff
@@ -17,8 +17,8 @@ pub enum MatrixRolesInCombinationBasicViolation {
     #[error("The factor source was not found in any role")]
     FactorSourceNotFoundInAnyRole,
 
-    #[error("The number of days until auto confirm must be greater than zero")]
-    NumberOfDaysUntilAutoConfirmMustBeGreaterThanZero,
+    #[error("The number of days until timed confirm is callable must be greater than zero")]
+    NumberOfDaysUntilTimeBasedConfirmationMustBeGreaterThanZero,
```
Sajjon, line 21: Breaking Profile format change number 2 - @sergiupuhalschi-rdx  / @danvleju-rdx  / @matiasbzurovski might get affected in host

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
```diff
@@ -0,0 +1,113 @@
...
+        let (init_method, init_input) =
+            kind.input_for_initialization(factors_and_time_input);
+        builder = builder.call_method(
+            access_controller_address.scrypto(),
+            init_method,
+            (init_input.deref(),),
+        );
+
+        // QUICK CONFIRM RECOVERY - Only if we can exercise the confirmation role explicitly.
+        if let Some((confirm_method, confirm_input)) =
```
Sajjon, line 65: Only call "quick confirm" if we can exercise confirmation role explicitly. Else we skip quick confirm and we will LATER (`number_of_days_until_timed_confirmation_is_callable` later) make another transaction.

crates/transaction/manifests/src/manifests_security_shield/transaction_manifest_apply_security_shield_kind.rs
Sajjon: Thanks!

crates/transaction/manifests/src/manifests_security_shield/transaction_manifest_apply_security_shield_kind.rs
```diff
@@ -0,0 +1,194 @@
...
+/// A "selector" of which combination of Roles we can exercise with used
+/// to build different flavours of TransactionManifest for the Security Shield
+/// update.
+///
+/// Each combination of roles allows us to skip signing with certain factors
+/// and still be able to recover + confirm the AccessController update.
+#[derive(Clone, Debug, PartialEq, Eq, Hash, enum_iterator::Sequence)]
+pub enum TransactionManifestApplySecurityShieldKind {
+    /// (Primary Recovery Confirmation)
+    PrimaryAndRecoveryWithExplicitConfirmation,
```
dhedey, line 23: Under the two of three model, lots of these don't make sense to me.

I'd expect something like:
* `InitiateWithPrimaryCompleteWithRecovery` 
* `InitiateWithPrimaryCompleteWithConfirmation` 
* `InitiateWithRecoveryCompleteWithConfirmation` 
* `InitiateWithPrimaryDelayedCompletion`  // Either Timed or (in future) external factors
* `InitiateWithRecoveryDelayedCompletion`  // Either Timed or (in future) external factors

And then in future, there might be modes where the user can't initiate themselves and needs to send a request to an external source (e.g. a friend or custodian) to do, which might look like:
* `ExternalInitiateWithPrimary`
* `ExternalInitiateWithRecovery`

CyonAlexRDX: Yes good point. 

Thanks! 

I answered on Slack too:  https://rdxworks.slack.com/archives/C031A0V1A1W/p1737728574375569?thread_ts=1737709759.091139&channel=C031A0V1A1W&message_ts=1737728574.375569

crates/uniffi/uniffi_SPLIT_ME/src/profile/mfa/security_structures/matrices/decl_matrix_macro.rs
```diff
@@ -25,7 +25,7 @@ macro_rules! matrix_conversion {
                 pub recovery_role: #recovery_role_type,
                 pub confirmation_role: #confirmation_role_type,
 
-                pub number_of_days_until_auto_confirm: u16,
+                pub number_of_days_until_timed_confirmation_is_callable: u16,
```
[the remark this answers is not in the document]
CyonAlexRDX: Changed type to use TimePeriod

crates/transaction/manifests/src/manifests_security_shield/transaction_manifest_apply_security_shield_kind.rs
CyonAlexRDX: @dhedey I used your suggested variants :)?

crates/transaction/manifests/src/manifests_security_shield/roles_exercisable_in_transaction_manifest_combination.rs
```diff
@@ -0,0 +1,224 @@
...
+                | Self::InitiateWithPrimaryDelayedCompletion
+        )
+    }
+
+    /// If we can set the ROLA key for this combination of roles.
+    pub fn can_set_rola_key(&self) -> bool {
+        self.can_exercise_primary_role()
+    }
+
+    fn initiate_recovery_with_primary(&self) -> bool {
```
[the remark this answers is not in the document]
CyonAlexRDX: It is more `is_specified_to_use_primary_as_initiating_role`

So I think this is good, or the more verbose above

crates/transaction/manifests/src/manifests_security_shield/confirm_timed_recovery.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn confirm_timed_recovery() -> Result<TransactionManifest>;
+    // {
+    //     builder.call_method(
+    //          access_controller,
+    //          SCRYPTO_ACCESS_CONTROLLER_TIMED_CONFIRM_RECOVERY_IDENT,
+    //         ScryptoAccessControllerTimedConfirmRecoveryInput::from(
+    //             factors_and_time,
+    //         ),
```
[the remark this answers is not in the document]
Sajjon: I think best done in a follow-up PR ? Or do you prefer the same PR?

crates/transaction/manifests/src/manifests_security_shield/transaction_manifest_apply_security_shield_kind.rs
CyonAlexRDX: Fixed

crates/transaction/manifests/src/manifests_security_shield/top_up_access_controller_xrd_vault.rs
```diff
@@ -36,14 +33,69 @@ pub trait TransactionManifestAccessControllerXrdVaultToppingUp {
...
+                let is_unable_to_top_up_xrd_vault =
+                    payer_is_entity_applying_shield
+                        && cannot_exercise_primary_role;
+                if is_unable_to_top_up_xrd_vault {
+                    // The payer is the entity applying the shield, but the manifest is not classified as
+                    // to be able to exercise the primary role. Thus we will not be able to
+                    // top up the XRD vault of the access controller.
+                    return Err(CommonError::Unknown); // TODO: Add error
+                }
+            }
```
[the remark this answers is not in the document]
GhenadieVP: Sure, but when we do build the WalletInteraction we will always build a manifest variant that initiates the recovery with primary, that is the manifest we will show to the user. 

Now when user slides to sign, Sargon internally will try to create the other variants, where it should not include instructions requiring primary role for combination such as `InitiateWithRecoveryCompleteWithConfirmation`

crates/transaction/manifests/src/manifests_security_shield/confirm_timed_recovery.rs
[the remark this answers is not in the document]
GhenadieVP: can be a follow up, just mentioned it

crates/transaction/manifests/src/manifests_security_shield/confirm_timed_recovery.rs
dhedey: You could also read the provisional security config from the network (i.e. gateway); but it's possibly safer to save it in the profile...
We should probably at least validate our saved config against the network state before confirming a timed recovery (in case e.g. someone has cancelled the recovery or something)

crates/transaction/manifests/src/manifests_security_shield/lock_fee_against_xrd_vault_of_access_controller.rs
```diff
@@ -0,0 +1,62 @@
...
+            access_controller_address.scrypto(),
+            SCRYPTO_ACCESS_CONTROLLER_LOCK_RECOVERY_FEE_IDENT,
+            ScryptoAccessControllerLockRecoveryFeeInput {
+                amount: ScryptoDecimal192::from(fee),
+            },
+        );
+
+        // ... then append all instructions from the original manifest
+        builder = builder.extend_builder_with_manifest(manifest);
```
dhedey, line 55: (This use is safe because we haven't created any new named manifest objects)

crates/transaction/manifests/src/manifests_security_shield/top_up_access_controller_xrd_vault.rs
```diff
@@ -36,14 +33,69 @@ pub trait TransactionManifestAccessControllerXrdVaultToppingUp {
...
+            let payer_is_entity_applying_shield =
+                entity_applying_shield.address() == payer.address().into();
+            if payer_is_entity_applying_shield {
+                let cannot_exercise_primary_role =
+                    !manifest.explicitly_references_primary_role();
+                let is_unable_to_top_up_xrd_vault =
+                    payer_is_entity_applying_shield
+                        && cannot_exercise_primary_role;
+                if is_unable_to_top_up_xrd_vault {
+                    // The payer is the entity applying the shield, but the manifest is not classified as
```
dhedey, line 91: I guess there's an implicit assumption here that:

Because we select manifests wit primary roles first, if we haven't selected a manifest with a primary role, it must be that we don't have access to the primary role, and therefore can't access the account to withdraw some XRD.

Is that right?

crates/transaction/manifests/src/manifests_security_shield/transaction_manifest_unsecurified_entity_owner_badge_into_bucket_putting.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// "securify" and put said owner badge in a Bucket and returns the bucket.
+    fn put_owner_badge_in_bucket(
+        builder: ScryptoTransactionManifestBuilder,
+        unsecurified_entity: &AnyUnsecurifiedEntity,
+    ) -> (ScryptoTransactionManifestBuilder, ScryptoManifestBucket) {
+        let is_account = unsecurified_entity.entity.is_account_entity();
+        let owner_badge_bucket_name = "owner_badge_bucket";
+        let owner_badge = if is_account {
+            SCRYPTO_ACCOUNT_OWNER_BADGE
+        } else {
```
dhedey, line 21: MINOR: Minor robustness smell here using a boolean - I'd suggest a `enum WalletEntityType { Account, Identity }` instead - so that if this ever changed in future, it'd be a compiler error here.

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
```diff
@@ -0,0 +1,273 @@
...
+        if should_set_rola_key {
+            if kind.can_set_rola_key() {
+                builder = TransactionManifest::set_rola_key(
+                    builder,
+                    &security_structure_of_factor_instances
+                        .authentication_signing_factor_instance,
+                    &entity_address,
+                );
+            } else {
+                return None; // Nothing has "failed" really, but we cannot proceed with this combination.
```
dhedey, line 91: I'm not sure I understand what this failure case is - why would `kind.can_set_rola_key()` fail?

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
```diff
@@ -0,0 +1,273 @@
...
+            }
+        }
+
+        let manifest = TransactionManifest::sargon_built(
+            builder,
+            securified_entity.network_id(),
+        );
+
+        // N.B.
```
dhedey, line 100: This is fine, although it's generally better to lock a fee of `0` and then edit it later. This provides a more accurate previewed fee.

dhedey (approved):
Nice! :+1: 

Didn't see any obvious logical issues, but had a few minor comments regarding rust coding norms which I noticed during the review and thought were worth mentioning.

crates/profile/models/security-structures/src/roles_matrices_structures/matrices/builder/matrix_builder.rs
```diff
@@ -603,14 +606,14 @@ impl MatrixBuilder {
...
-    /// 4. Number of days until auto confirm is greater than zero
+    /// 4. Number of days until timed confirm is callable is greater than zero
     fn validate_combination(&self) -> MatrixBuilderMutateResult {
         self.validate_if_primary_has_single_it_must_not_be_used_by_any_other_role()?;
         self.validate_primary_cannot_have_multiple_devices()?;
         self.validate_no_factor_may_be_used_in_both_primary_threshold_and_override()?;
         self.validate_no_factor_may_be_used_in_both_recovery_and_confirmation(
         )?;
-        self.validate_number_of_days_until_auto_confirm()?;
+        self.validate_time_until_delayed_confirmation_is_callable()?;
```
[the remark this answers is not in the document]
CyonAlexRDX: Yeah this code has become a bit messy since requirements has changed 4 times since implementation. I believe @sergiupuhalschi-rdx @danvleju-rdx are in top of things.

crates/system/os/factors/src/test_instances_provider.rs
```diff
@@ -1632,7 +1632,7 @@ async fn securify_personas_when_cache_is_half_full_single_factor_source() {
                 [],
                 [bdfs.clone()],
             ),
-            1,
+            TimePeriod::with_days(1),
```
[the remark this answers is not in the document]
CyonAlexRDX: Agreed 100

crates/transaction/manifests/src/manifest_builder_from_manifest.rs
```diff
@@ -2,7 +2,35 @@ use radix_transactions::prelude::ManifestBuilder;
...
+    }
+}
+
+impl BuilderExtendWithInstructions for ManifestBuilder {
+    fn extend_builder_with_instructions(
+        self,
+        instructions: impl IntoIterator<Item = ScryptoInstruction>,
+    ) -> ManifestBuilder {
+        instructions.into_iter().fold(self, |builder, instruction| {
+            builder.add_instruction_advanced(instruction).0
```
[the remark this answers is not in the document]
CyonAlexRDX: We only insert instruction in middle outside of ManifestBuilder - for TX Guarantees - which we can do since we can unit that instruction directly.

And when we [must](https://rdxworks.slack.com/archives/C071C6DB8TG/p1737463675178439?thread_ts=1737463301.189909&channel=C071C6DB8TG&message_ts=1737463675.178439) use builder we only insert at index 0 - eg lock fee against AccessControllers XRD Vault

Or we append instruction last - eg contribute to XRD vault of an AC.

So in all those scenarios we should be Ok, I think.

crates/transaction/manifests/src/manifests_security_shield/roles_exercisable_in_transaction_manifest_combination.rs
```diff
@@ -0,0 +1,224 @@
...
+    // User can't initiate themselves and needs to send a request to an external source (e.g. a friend or custodian)
+    // ExternalInitiateWithPrimary
+    // ExternalInitiateWithRecovery
+}
+
+impl Default for RolesExercisableInTransactionManifestCombination {
+    fn default() -> Self {
+        // we default to a combination containing Primary since it allows
+        // use to top up XRD vault of AccessController
+        Self::InitiateWithPrimaryCompleteWithRecovery
```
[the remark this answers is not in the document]
CyonAlexRDX: The to be displayed manifest for TX review. Sure can be named ctor instead and will switch to CompleteWithConfirmation

Cc @GhenadieVP

crates/transaction/manifests/src/manifests_security_shield/roles_exercisable_in_transaction_manifest_combination.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub fn all() -> IndexSet<Self> {
+        enum_iterator::all::<Self>().collect()
+    }
+
+    /// If this combination of roles references the `Primary` role or not.
+    pub fn can_exercise_primary_role(&self) -> bool {
+        matches!(
+            self,
+            Self::InitiateWithPrimaryCompleteWithRecovery
+                | Self::InitiateWithPrimaryCompleteWithConfirmation // | Self::InitiateWithPrimaryDelayedCompletion
```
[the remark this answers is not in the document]
CyonAlexRDX: Requires Dugong which is not live yet?

crates/transaction/manifests/src/manifests_security_shield/confirm_timed_recovery.rs
CyonAlexRDX: We need the actual HDKeys which network does not have. So need to use Profile models

crates/transaction/manifests/src/manifests_security_shield/top_up_access_controller_xrd_vault.rs
CyonAlexRDX: Yeah, for the manifest variant built from `init_R_complete_C` for example we should not modify the manifest so that it suddenly requires Primary role

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
CyonAlexRDX: Setting owner_keys hashes (to the hash of the authentication_signing_factor_instance aka ROLA key) can only be done by Primary role is my understanding? Is that assumption incorrect?

So any manifest which sets the ROLA key will require Primary Role. Thus a `init_R_complete_C` manifest variant which sets ROLA would not be signable since it does not exercise Primary.

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
CyonAlexRDX: Harder to modify later

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
dhedey: Ahh right ... I think this could do with a better comment?

We could work around this limitation a few ways:
* Signing with the new primary after recovery, if possible, in order to create the proof and update the ROLA badge.
* If doing a quick-confirm, then we can start by recovering to all ALLOW_ALL, then update the ROLA key, then recover to the correct framework.

Probably the easiest/more sane thing to to do is require a signature from the new primary as part of confirmation (either quick or delayed) and update the ROLA at that point?

crates/profile/models/security-structures/src/roles_matrices_structures/matrices/builder/matrix_builder.rs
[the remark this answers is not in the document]
GhenadieVP: @dhedey this validation is up to date, and note that we would allow users to build Weak shields which will not follow the rules you specified

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
GhenadieVP: >This provides a more accurate previewed fee.

@dhedey Wallet has logic to include the cost of the lock fee instruction in the total transaction fee

crates/profile/models/security-structures/src/roles_matrices_structures/matrices/builder/matrix_builder.rs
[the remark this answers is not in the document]
dhedey: :+1: OK, so this is the new minimal ruleset; and there's the "recommended" ruleset applied somewhere else?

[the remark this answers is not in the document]
GhenadieVP: minimal is the recommended, unless you mean something else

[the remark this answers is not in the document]
sergiupuhalschi-rdx: This function is no longer used as the minimal ruleset but as the recommended one. The minimal ruleset is enforced [here](https://github.com/radixdlt/sargon/blob/main/crates/profile/models/security-structures/src/roles_matrices_structures/roles/builder/roles_builder.rs#L508) when attempting to add a factor to a role.

crates/transaction/manifests/src/manifests_security_shield/manifests_securify_shield_securified_entity.rs
CyonAlexRDX: Changed to this:

```rust
pub fn order_of_instruction_setting_rola(
    &self,
    security_structure_of_factor_instances: &SecurityStructureOfFactorInstances,
    entity: &AnySecurifiedEntity,
) -> OrderOfInstructionSettingRolaKey {

// unindented body to fit better in Github
if security_structure_of_factor_instances
    .authentication_signing_factor_instance
    == entity.current_authentication_signing_factor_instance()
{
    // The factor of ROLA key is the same as the current factor of the entity
    // => we can skip setting the ROLA key.
    return OrderOfInstructionSettingRolaKey::NotNeeded;
}

if self.can_quick_confirm() {
    // we can quick confirm so we can always set the ROLA key using
    // the NEW factors - disregarding of which role initiated recovery.
    OrderOfInstructionSettingRolaKey::AfterQuickConfirm
} else {
    match self.role_initiating_recovery() {
        RoleInitiatingRecovery::Primary => {
            // We cannot quick confirm and set ROLA key using new factors in
            // the same TX, but fortunately we can set the ROLA key and auth with
            // th old factors if we place the SET_METADATA instruction
            // **before** initiating recovery instruction.
            OrderOfInstructionSettingRolaKey::BeforeInitRecovery
        }
        RoleInitiatingRecovery::Recovery => {
            // We cannot set ROLA key in this TX, because we are not using the
            // quick confirm - so we must do it in a future TX.
            OrderOfInstructionSettingRolaKey::MustSetInFutureTxForConfirmRecovery
        }
    }
}
// ^ unindented body to fit better in Github ^
}

```

sergiupuhalschi-rdx (approved):

PR #356, opened 2025-01-24
Authorization Interactor

micbakos-rdx:
A new interactor responsible for asking the user through the host to authorise an operation. Such operation is called `AuthorizationPurpose` and it currently supports variants for creating one or batch accounts, or one or batch personas.If the user rejects authorization, then such operation does not mutate the profile, nor does consume the cache.

The user can either `AuthorizationResponse::Authorized` or `AuthorizationResponse::Rejected`.

Note that, I also changed the name of the error used to be called `SigningRejected` to `HostInteractionAborted`. It can be thrown by hosts when answering to the interactor callbacks, and returned back to the method that initiated the interaction. It is a "common result" that hosts receive and can just reset the UI for the user to try again.

### DIFF
--- a/crates/system/os/os/src/sargon_os_personas.rs
+++ b/crates/system/os/os/src/sargon_os_personas.rs
@@ -1301,7 +1301,7 @@ mod tests {
         let interactors =
             Interactors::new_from_clients_and_authorization_interactor(
                 &clients,
-                Arc::new(TestAuthorizationInteractor::new_rejecting_only(
+                Arc::new(TestAuthorizationInteractor::stubborn_rejecting_specific_purpose(
                     AuthorizationPurpose::CreatingPersona,
                 )),
             );
@@ -1344,7 +1344,7 @@ mod tests {
         let interactors =
             Interactors::new_from_clients_and_authorization_interactor(
                 &clients,
-                Arc::new(TestAuthorizationInteractor::new_rejecting_only(
+                Arc::new(TestAuthorizationInteractor::stubborn_rejecting_specific_purpose(
                     AuthorizationPurpose::CreatingPersonas,
                 )),
             );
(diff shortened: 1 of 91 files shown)


crates/system/os/os/src/sargon_os_personas.rs
```diff
@@ -371,17 +385,29 @@ impl SargonOS {
...
+
+        let authorization = self
+            .authorization_interactor()
+            .request_authorization(AuthorizationPurpose::CreatingPersonas)
+            .await;
+
+        match authorization {
+            AuthorizationResponse::Rejected => {
+                debug!("User rejected authorization, aborting.");
+                return Err(CommonError::HostInteractionAborted);
```
CyonAlexRDX, line 398: Good! We auth before consuming instances in Cache, perfect

CyonAlexRDX: Do we have any unit test that fails if you would move up consume to be called before auth? 

Otherwise I prefer it if you can add such a unit test :) 

Typically create an account, Alice 

Try creating account and abort/reject.

Try crating account, Bob, again this time Auth.

Assert that Bobs VECI derivation entity Index is that of Alice+1 and not +2.

Should do it right? 

And then analogously for Persona

CyonAlexRDX (approved):
LGTM, just missing test which fails If we were to refactor order of things

crates/system/os/os/src/sargon_os_personas.rs
micbakos-rdx: Take a look at the last commits. I had to change a bit the API of the `TestAuthorizationInteractor`.

CyonAlexRDX: Not pushed yet?

micbakos-rdx: No, sorry waiting for pre-commit

micbakos-rdx: You can check now

CyonAlexRDX: Looks great

matiasbzurovski (approved):
Looks great!

micbakos-rdx:
Testing the integration with host, will merge when I verify the interaction.

GhenadieVP (approved):

PR #359, opened 2025-01-28
Move app/signing into factors/signatures-collector

micbakos-rdx:
Signing crate includes signatures-collector . It could be inside `factors` instead of `app` as is keys-collector, since it is an internal api.

Created a separate PR not to litter any other PR with many changes.


CyonAlexRDX (approved):
LGTM - in future prefer bumping version **after** PR approval, makes review a bit easier :)

GhenadieVP (approved):

PR #360, opened 2025-01-28
Create WalletToDappInteractionAuthProof on Sargon

matiasbzurovski:
## Changelog
Move to Sargon the creation of `WalletToDappInteractionAuthProof` (used when signing auth) so that it isn't Hosts responsibility to decide which `curve` to use.


Sajjon (commented):
Lgtm

CyonAlexRDX (approved):

PR #361, opened 2025-01-29
Remove SecurityQuestions & TrustedContact from UniFFI & Hosts

matiasbzurovski:
## Changelog
Given the `SecurityQuestions` & `TrustedContact` factor sources aren't yet supported, we are removing every reference to it from UniFFI exported code. Same as we currently do for `EntitySecurityState`, Sargon internal code will keep logic for the mentioned factor sources, but the exported lib used by hosts won't.

The mentioned change involves:
- Removing the variants from `FactorSource`
- Removing the variants from `FactorSourceKind`
- Removing the `FactorSourceId::Address` variant (only used by `TrustedContact`, which is no longer available).
- Removing `SecurityQuestionsNotProductionReadyFactorSource` & `TrustedContactFactorSource`.


CyonAlexRDX (approved):
host side is good. and then post-switched-to single line comment in UniFFI rust all good. preemptive approve

PR #362, opened 2025-01-29
Set default security shield API

giannis-rdx:
This PR is part of the [ABW-4146](https://radixdlt.atlassian.net/browse/ABW-4146).
It adds an API to change the default security shield

[ABW-4146]: https://radixdlt.atlassian.net/browse/ABW-4146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/system/os/factors/src/sargon_os_security_structures.rs
```diff
@@ -152,6 +157,65 @@ impl OsSecurityStructuresQuerying for SargonOS {
...
+            if let Some(current_default_shield) = &current_default_shield {
+                p.app_preferences
+                    .security
+                    .security_structures_of_factor_source_ids
+                    .update_with(current_default_shield.metadata.id, |s| {
+                        s.metadata.remove_flag(SecurityStructureFlag::Default)
+                    });
+            }
+
+            Ok(p.app_preferences
```
CyonAlexRDX, line 199: Prefer creating a method for this inside Profile logic crate.

Better to delegate as much logic relating to profile changes as possible to Profile logic crate. And let SargonOS be a as dumb as possible "coordinator"

giannis-rdx: [done](https://github.com/radixdlt/sargon/pull/362/commits/5c9af2b9ed6da20d6d1c2267c286b225127f1e58) ?

crates/profile/logic/logic_SPLIT_ME/src/logic/profile_update.rs
```diff
@@ -338,6 +338,46 @@ impl ProfileFactorSourceUpdating for Profile {
...
+        shield_id: &SecurityStructureID,
+    ) -> Result<bool>;
+}
+
+impl ProfileSecurityStructuresUpdating for Profile {
+    fn update_security_structure_remove_flag_default(
+        &mut self,
+        shield_id: &SecurityStructureID,
+    ) -> Result<bool> {
+        Ok(self
```
Sajjon, line 358: why return `Result`?

Sajjon (commented):
I mean you should do this - favour composition of many small **resusable** functions, instead of a bigger function which is not reausable. 

The idea of SargonOS is to **not have Profile model logic**. So it should delegate that logic to `profile-logic` crate. 

```rust
// In os-factors crate
async fn set_default_security_structure(
    &self,
    shield_id: SecurityStructureID,
) -> Result<()> {
    let updated_ids = self.update_profile_with(|p| Ok(p.set_default_security_structure(shield_id)))?;

    // Emit event
    self.event_bus
        .emit(EventNotification::profile_modified(
            EventProfileModified::SecurityStructuresUpdated {
                ids: updated_ids,
            },
        ))
        .await;

    Ok(())
}

// In Profile Logic crate
impl AppPreferences {
    fn get_default_scurity_structure(&self) -> Option<SecurityStructureOfFactorSourceIds> {
        self
            .security
            .security_structures_of_factor_source_ids
            .iter()
            .find(|s| s.metadata.is_default());
    }
}
impl Profile {
    fn get_default_Security_structure(&self) -> Option<SecurityStructureOfFactorSourceIds> {
        self
            .app_preferences.get_default_security_structure()
    }

    /// Returns the list of IDs of updated SecurityShields - either one or two elements,
    /// depending on if shield identified by `shield_id` already was "default" shield or not.
    fn set_default_security_structure(
        &self,
        shield_id: SecurityStructureID,
    ) -> Result<<Vec<SecurityStructureID>> {

        let current_default_shield = self.get_default_Security_structure();

        let updated_ids = match current_default_shield {
            Some(ref current_default_shield) => {
                vec![current_default_shield.metadata.id, shield_id]
            }
            None => vec![shield_id],
        };

        if let Some(current_default_shield) = &current_default_shield {
            self.update_security_structure_remove_flag_default(
                &current_default_shield.metadata.id,
            )
        }
        p.update_security_structure_add_flag_default(&shield_id)

        updated_ids
    }
}
```

crates/profile/logic/logic_SPLIT_ME/src/logic/profile_update.rs
Sajjon: If ID not found. Right. 

So do you map to correct Error if not found?

crates/profile/models/security-structures/src/roles_matrices_structures/security_structure_flags.rs
```diff
@@ -0,0 +1,34 @@
...
+
+/// Flags which describe a certain state a Security Shield might be in, primarily used
+/// to mark which Security Shield is the **default** one.
+#[derive(
+    Serialize, Deserialize, Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord,
+)]
+#[serde(rename_all = "camelCase")]
+pub enum SecurityStructureFlag {
+    /// Used to mark a Security Shield as "default". We can only have one.
+    Default,
```
matiasbzurovski, line 11: should we call it `Main` as we do on `FactorSourceFlag` (despite in UI it will be treated as "default")?

If so, I would update the name of every related function as well

crates/profile/models/security-structures/src/roles_matrices_structures/security_structure_metadata.rs
```diff
@@ -25,31 +29,52 @@ impl SecurityStructureMetadata {
...
         Self::with_details(
             SecurityStructureID::from(Uuid::new_v4()),
             display_name,
             now(),
             now(),
+            flags,
         )
     }
+
+    pub fn set_flag(&mut self, flag: SecurityStructureFlag) {
```
matiasbzurovski, line 56: `insert_flag` ?

crates/profile/logic/logic_SPLIT_ME/src/logic/profile_update.rs
Sajjon: as I wrote in https://github.com/radixdlt/sargon/pull/362#pullrequestreview-2583236391 there should be an update method on AppPreferences , which should actually should delegate to `Security` which should have this impl, but you need to change the impl to **actually return an Error if not found** - which you are not doing now. So `Result<()>` is correct, but now you are incorrectly returning `Bool` which you **unconditionally** wrap in a `Result`.

```rust
impl Security {
  /// # Throws
  /// Throws `CommonError:InvalidSecurityStructureID` if the structure identified by `shield_id` does not exist.
  fn update_security_structure_remove_flag_default(
		&mut self,
		shield_id: &SecurityStructureID,
	) -> Result<()> {
		self
			.security_structures_of_factor_source_ids
			.try_update_with(shield_id, |s| {
				s.metadata.remove_flag(SecurityStructureFlag::Default)
			}))
			.map(|_| CommonError::InvalidSecurityStructureID { bad_value: shield_id.to_string() })
	}
}
```

Sajjon: and then ofc unit test all the way :)

crates/system/os/factors/src/sargon_os_security_structures.rs
```diff
@@ -334,6 +390,111 @@ mod tests {
...
+        .await
+        .unwrap();
+
+        // ASSERT
+        assert!(event_bus_driver.recorded().iter().any(|e| e.event
+            == Event::ProfileModified {
+                change: EventProfileModified::SecurityStructuresUpdated {
+                    ids: vec![structure_ids_sample_other.metadata.id()]
+                }
+            }));
```
matiasbzurovski, line 495: need to check `ProfileSaved` was emitted as well

Also, we should check that `SecurityStructuresUpdated` is emitted with two ids when there was an existent main shield, and with one id when there isn't

matiasbzurovski (changes requested):

crates/profile/logic/logic_SPLIT_ME/src/logic/profile_update.rs
giannis-rdx: I understand why you want reusable functions and not having profile logic in SargonOS, but I don't understand at all where am I supposed to place the new functions. The way the profile-logic is structured doesn't really help me.

Therefore we need the following functions:
- `update_security_structure_remove_flag_default`: where this is supposed to go? in (profile crate) `models/app_preferences`, I see there is a query fun there `has_gateway_with_url`. But it doesn't seem the right place to me. Should it be in the `profile_update`? Should I create a `security_structures_logic` file under `profile/logic`?
- `update_security_structure_set_flag_default`: same as above
- `get_default_scurity_structure`: a new file `query_app_preferences` under `profile/logic`? or in the `models/app_preferences`?
- `set_default_security_structure`: it looks like it has to to `profile_update` or something else?

crates/profile/models/security-structures/src/roles_matrices_structures/security_structure_metadata.rs
giannis-rdx: ✅

crates/profile/models/security-structures/src/roles_matrices_structures/security_structure_flags.rs
giannis-rdx: Well, it’s not exactly the same. “Main” indicates the primary or most important option, while “default” refers to something preselected. That’s why in UI, it is (correctly) defined as “default.”

Based on that, do you really think we should treat it as “main” in the code? 🤔

matiasbzurovski: I think they are both (Device Factor Sources & Security Shields) kind of the same. Nothing prevents using a non-main factor source/security shield, but host apps will use the main one when either creating/securifying an entity.

CyonAlexRDX: Yeah `main` in code makes sense to call it

giannis-rdx: We have democracy, so if the majority agrees, I will change it 😄

crates/profile/logic/logic_SPLIT_ME/src/logic/app_preferences/security_structures_update.rs
```diff
@@ -0,0 +1,97 @@
...
+            }
+            None => vec![*shield_id],
+        };
+
+        if let Some(current_main_shield) = &current_main_shield {
+            self.update_security_structure_remove_flag_main(
+                &current_main_shield.metadata.id,
+            )?;
+        }
+        self.update_security_structure_add_flag_main(shield_id)?;
```
Sajjon, line 31: You must do this FIRST (set new main) and THEN remove from old(if any). Otherwise if the shield by shield_id does not exist, you have already removed the default flag of the old.

So add a test for this first - which should fail - and then update code and the test should succeed :)

giannis-rdx: oh right, good point.

giannis-rdx: done ✅

CyonAlexRDX (approved):
LGTM

crates/profile/logic/logic_SPLIT_ME/src/logic/app_preferences/security_structures_update.rs
```diff
@@ -0,0 +1,98 @@
+use crate::prelude::*;
+
+pub trait MainSecurityStructureUpdating {
+    /// Returns the list of IDs of updated SecurityShields - either one or two elements,
+    /// depending on if shield identified by `shield_id` already was "main" shield or not.
```
matiasbzurovski, line 5: this comment seems incorrect. It returns one element if there wasn't any existing main SecurityShield, while it returns two elements if there was a previous main Security Sehidl whose flag was removed

matiasbzurovski (approved):
LGTM, one comment that seems incorrect and good to go

crates/profile/logic/logic_SPLIT_ME/src/logic/app_preferences/security_structures_update.rs
giannis-rdx: fixed ✅

PR #364, opened 2025-01-30
Expose make interaction for applying shields function

micbakos-rdx:
* Method is exposed in uniffi
* For now `CannotSecurifyEntityHasProvisionalSecurityConfig` is not returned when provisional shield is in place since this error should be thrown when a provisional shield exists for that entity but there is also a pending transaction in the queue.
* Rola key metadata is set after the owner badge is set to the auth zone. 

Not authorizing the set metadata update ends in the following error
```
SystemModuleError(AuthError(Unauthorized(Unauthorized { failed_access_rules: RoleList([(RoleKey { key: "metadata_setter" }, [Protected(BasicRequirement(Require(NonFungible(ResourceAddress(9a4c...6318c6):[5167...1250]))))])]), fn_identifier: PackageAddress(0d90....6318c6):<Metadata>:"set" })))
```


crates/system/os/factors/src/apply_security_shield/sargon_os_apply_shield.rs
```diff
@@ -84,18 +84,23 @@ impl OsShieldApplying for SargonOS {
...
-            })
-        {
-            return Err(
-                CommonError::CannotSecurifyEntityHasProvisionalSecurityConfig,
-            );
-        }
+        // TODO change when queue is introduced
+        // `CannotSecurifyEntityHasProvisionalSecurityConfig` should be returned when
+        // 1. entity's shield is in provisional state
+        // 2. interaction is in the queue.
```
CyonAlexRDX, line 90: When 1. AND 2. Right? I think an update to the comment stating the Boolean AND make it more clear

crates/system/os/factors/src/apply_security_shield/sargon_os_apply_shield.rs
```diff
@@ -181,11 +186,11 @@ impl OsShieldApplying for SargonOS {
         InstancesInCacheConsumer,
         FactorInstancesProviderOutcome,
     )> {
-        if !entities.iter().all(|a| a.get_provisional().is_none()) {
-            return Err(
-                CommonError::CannotSecurifyEntityHasProvisionalSecurityConfig,
-            );
-        }
+        // if !entities.iter().all(|a| a.get_provisional().is_none()) {
```
CyonAlexRDX, line 189: I would copy paste the comment above here too, explaining why code is commented out

CyonAlexRDX (approved):
Nice, some minor comments

PR #366, opened 2025-01-31
Remove duplicate fsi assert from veci

sergiupuhalschi-rdx:
Remove the **assert** of address and factor instance derivation path discrepancies from `VirtualEntityCreatingInstance` ctor non-debug code. If we identify a discrepancy, we log it with `error!`.


crates/profile/models/supporting-types/src/veci.rs
```diff
@@ -65,6 +68,32 @@ impl VirtualEntityCreatingInstance {
...
+    ///
+    /// However, even though the bug was fixed after less than 5 months, we have
+    /// end users which are in this bad state (shared instances between Personas and Acccounts).
+    ///
+    /// Thus we can't assert the discrepancy.
+    ///
+    /// [identpr]: https://github.com/radixdlt/sargon/pull/254/files#r1860748013
+    /// [badpr]: https://github.com/radixdlt/babylon-wallet-android/pull/1042
+    /// [goodpr]: https://github.com/radixdlt/babylon-wallet-android/pull/1256
+    fn check_duplicate_factor_source_instances(
```
Sajjon, line 87: This fn has an incorrect name. It sgould be `check_for_derivation_path_discrepancies` or similar

Sajjon: Furthemore you can maybe put a `debug_assertion` gated behind #[cfg(test)] in the body

Sajjon: fn check_for_derivation_path_discrepancies(
        factor_instance: &HierarchicalDeterministicFactorInstance,
        address: &AddressOfAccountOrPersona,
    ) {
       let discrepancy_found = factor_instance.derivation_path().get_entity_kind()
            == address.get_entity_kind();
let error_msg = "Discrepancy! Address and DerivationPath of FactorInstances have different entity kinds.";
        if discrepancy_found
        {
            error!(error_msg);

        }
            #[cfg(test]
            debug_assert!(!discrepancy_found, error_msg)
    }

sergiupuhalschi-rdx: Although I was close to getting it right, I think your comment saved me from another round of changes :)) Thanks!

crates/profile/models/supporting-types/src/veci.rs
```diff
@@ -65,6 +68,36 @@ impl VirtualEntityCreatingInstance {
...
+    /// [identpr]: https://github.com/radixdlt/sargon/pull/254/files#r1860748013
+    /// [badpr]: https://github.com/radixdlt/babylon-wallet-android/pull/1042
+    /// [goodpr]: https://github.com/radixdlt/babylon-wallet-android/pull/1256
+    fn check_for_derivation_path_discrepancies(
+        factor_instance: &HierarchicalDeterministicFactorInstance,
+        address: &AddressOfAccountOrPersona,
+    ) {
+        let discrepancy_found =
+            factor_instance.derivation_path().get_entity_kind()
+                == address.get_entity_kind();
```
Sajjon, line 93: it should be `!=` and not `==` right?.

Missing unit tests as well :)

sergiupuhalschi-rdx: Isn't [this](https://github.com/radixdlt/sargon/pull/366/files#diff-6a594aba60de270108c80de9cb62109335ba5b8834864e6320b10e888a581a36R167) unit test enough?

Sajjon: I realize it was ME who accidentally gave you `==` :D :D sorry about that

and yes, sorry I was accidentally checking only last commit... that test is fine!

CyonAlexRDX (approved):

crates/profile/models/supporting-types/src/veci.rs
sergiupuhalschi-rdx: No worries, I should have double-checked. Thanks for spotting this!

PR #367, opened 2025-01-31
Securify Entity classification

micbakos-rdx:
* Securify classification in transaction manifest.
> [!note]
> RET does not provide a direct classification for securifying entities. So the code that detects it is temporary. 
* A function to mark an entity as securified. This function can be invoked by hosts when an securify transaction succeeds.
* Refactored the code a bit to generalise how we retrieve global ancestor addresses from one vault. This logic is reused when we detect that an account is deleted. The same code is needed now to retrieve the access controller of the newly securified entity.
* Initially developed a sync method that tries to retrieve information about entities on ledger and apply that in profile. This method is WIP and called `wip_sync_entities_state_on_ledger`. This method could be used in the future as a blueprint on how we could sync on ledger state with profile. On that note I renamed the trait from `OsSyncAccountsDeletedOnLedger` to `OsSyncEntitiesStateOnLedger` to accommodate the wip function but to also host sync operation on entities and not only accounts.


crates/crypto/addresses/src/address/non_fungible_local_id.rs
```diff
@@ -118,6 +118,19 @@ impl std::fmt::Display for NonFungibleLocalId {
     }
 }
 
+impl From<AddressOfAccountOrPersona> for NonFungibleLocalId {
+    fn from(value: AddressOfAccountOrPersona) -> Self {
+        match value {
+            AddressOfAccountOrPersona::Account(account_address) => {
+                From::<AccountAddress>::from(account_address)
```
CyonAlexRDX, line 125: Ive never seen `From::<AccountAddress>::from` before, it is kind of non standard syntax I guess. Dont you just want `Self::from(account_address` ?

crates/profile/logic/logic_SPLIT_ME/src/logic/entity/mark_entity_as_securified.rs
```diff
@@ -0,0 +1,44 @@
+use crate::prelude::*;
+
+pub trait ProfileMarkEntityAsSecurified {
+    fn mark_entity_as_securified(
+        &mut self,
+        access_controller_address: AccessControllerAddress,
```
CyonAlexRDX, line 6: and with my PR this method needs one more arg: `xrd_vault_address: VaultAddress` 

https://github.com/radixdlt/sargon/pull/349

but yeah that might go away... not approved by Ghenadie yet. Just a heads up.

crates/profile/logic/logic_SPLIT_ME/src/logic/entity/mark_entity_as_securified.rs
```diff
@@ -0,0 +1,44 @@
...
+        entity_address: AddressOfAccountOrPersona,
+    ) -> Result<AccountOrPersona> {
+        let mut entity = self.entity_by_address(entity_address)?;
+        let transaction_signing = entity
+            .security_state()
+            .as_unsecured()
+            .ok_or(CommonError::SecurityStateSecurifiedButExpectedUnsecurified)
+            .map(|security_state| security_state.transaction_signing.clone())?;
+
+        let security_structure_of_factor_instances = entity
```
CyonAlexRDX, line 24: I think we best split this into:

```rust
let provisional_config = entity
    .get_provisional()
    .ok_or(CommonError::EntityHasNoProvisionalSecurityConfigSet)?;

let security_structure_of_factor_instances = provisional_config.into_factor_instances_derived().ok_or(CommonError::ProvisionalConfigInWrongStateExpectedInstancesDerived)?; // 🔮 NEW Error kind
```

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/entities/sync/support/entity_sync_report.rs
```diff
@@ -0,0 +1,8 @@
+use crate::prelude::*;
+use sargon::EntitySyncReport as InternalEntitySyncReport;
+
+/// The report that gathers the different actions performed on profile after sync completes.
+#[derive(Clone, Debug, InternalConversion, uniffi::Record)]
+pub struct EntitySyncReport {
```
CyonAlexRDX, line 6: I actually like "Report" but I think "Outcome" is suitable here too, and it is the naming we have used for other situations. What do you think? Should ofc only use "outcome" if you agree it is similar with e.g. `SignaturesCollectorOutcome` etc.

micbakos-rdx: Sure

crates/system/os/transaction/src/tx/sargon_os_transaction_analysis.rs
```diff
@@ -248,6 +251,23 @@ impl OsExecutionSummary for SargonOS {
             });
         }
 
+        execution_summary.classify_securify_entity_if_present(
+            |entity_address| {
+                self.entity_by_address(entity_address)
+                    .and_then(|entity| {
+                        entity
+                            .get_provisional()
+                            .and_then(|p| p.into_factor_instances_derived().ok())
```
CyonAlexRDX, line 260: [my comment here](https://github.com/radixdlt/sargon/pull/367/files#r1952691477) applies here too

crates/system/os/transaction/src/tx/sargon_os_transaction_analysis.rs
```diff
@@ -187,6 +188,7 @@ impl OsExecutionSummary for SargonOS {
             .await?;
 
         Self::extract_execution_summary(
```
CyonAlexRDX, line 190: change this to `self.extract_execution_summary` instead

crates/system/os/transaction/Cargo.toml
```diff
@@ -17,6 +17,7 @@ profile = { workspace = true }
 # profile-state-holder = { workspace = true }
```
CyonAlexRDX, line 17: not your code: can you delete dead code `# profile-state-holder = { workspace = true }`

and sort https://rdxworks.slack.com/archives/C031A0V1A1W/p1739369746796369?thread_ts=1739369619.909279&cid=C031A0V1A1W

crates/system/os/factors/src/sargon_os_mark_as_securified.rs
```diff
@@ -0,0 +1,436 @@
...
+#[async_trait::async_trait]
+impl OsMarkAsSecurified for SargonOS {
+    async fn mark_entities_as_securified(
+        &self,
+        entity_addresses: IndexSet<AddressOfAccountOrPersona>,
+    ) -> Result<()> {
+        let network_id = self.current_network_id()?;
+
+        let gateway_client =
+            GatewayClient::new(self.http_client.driver.clone(), network_id);
```
CyonAlexRDX, line 22: now it is most def worth doing a:

```rust
impl SargonOS {
 /// instantiates a GatewayClient against the current network using the `http_client_driver`
    pub fn gateway_client(&self) -> GatewayClient { // Arc here? 
                  let network_id = self.current_network_id()?;

       
            GatewayClient::new(self.http_client.driver.clone(), network_id)
    }
}
```

and search for **all usages of `"GatewayClient::new(self.http_client"` and replace with that` (and then `cargo clippy --all -- -D warnings` might complain about some `network_id` variables not being used?)

crates/gateway/client-and-api/src/methods/state_methods.rs
```diff
@@ -57,80 +58,142 @@ impl GatewayClient {
             .map(|x| x.unwrap_or(Decimal192::zero()))
     }
 
+    /// Fetches the ancestor global address for each account by querying the vault location
+    /// of the owner badge of that account.
+    pub async fn fetch_owning_vault_global_ancestor_address_for_entities(
+        &self,
+        network_id: NetworkID,
+        entity_addresses: impl IntoIterator<Item = AddressOfAccountOrPersona>,
+    ) -> Result<HashMap<AddressOfAccountOrPersona, Option<Address>>> {
```
CyonAlexRDX, line 67: Can we use `IndexMap` instead of `HashMap` ?

CyonAlexRDX: Change return type from `Address` to `VaultAddress` please

CyonAlexRDX: resolved on Slack.

crates/system/os/factors/src/sargon_os_mark_as_securified.rs
micbakos-rdx: I started with this, but most of the methods that require a GW client also require a network id later down the road which is already queried here. So we have 3 options

1. Instantiate the gw client as you suggest. And then retrieve the network id from the GW client (`gateway_client.gateway.network.id`) so we don't have to query it again from profile.
2. Make this function return a tuple. A client and the network id under it was created. Too much IMHO.
3. Just query the network id wherever needed.

CyonAlexRDX: I like 1 and 2

CyonAlexRDX: `pub fn gateway_client_on(&self) -> (GatewayClient, NetworkID)`
`pub fn gateway_client(&self) -> GatewayClient { self.gateway_client_on().0 }`

perhaps?

CyonAlexRDX: and when you need network use `gateway_client_on` else `gateway_client`

GhenadieVP (approved):

CyonAlexRDX (approved):
Good job

PR #370, opened 2025-02-04
[DO-000] use specific artifact name to download it

duje-begonja-rdx:


.github/workflows/release-kotlin-desktop.yml
```diff
@@ -74,8 +74,14 @@ jobs:
...
 
-      - uses: RDXWorks-actions/download-artifact@main
+      - uses: RDXWorks-actions/download-artifact-v4@main
         with:
+          name: aarch64-apple-darwin
+          path: artifacts
+
+      - uses: RDXWorks-actions/download-artifact-v4@main
+        with:
+          name: x86_64-unknown-linux-gnu
```
Sajjon, line 84: should this really be `x86_64-unknown-linux-gnu` but the one above is `aarch64-apple-darwin` ? looks wrong with the assymetry?

micbakos-rdx (approved):

micbakos-rdx:
In summary:

The upload-artifact job was uploading the artifacts and was renaming them to the target triple's name:
Upload from:
```
path: target/x86_64-unknown-linux-gnu/release/libsargon_uniffi.so
name: x86_64-unknown-linux-gnu
```
AND
```
path: target/aarch64-apple-darwin/release/libsargon_uniffi.dylib
name: aarch64-apple-darwin
```

This still works as it used to, but,
In the past the `download-artifact` job we used, was downloading each artifact and when unwrapping it was also naming it with the name of the directory it was saved into. So
```
// It used to be like this
artifacts/x86_64-unknown-linux-gnu/libsargon_uniffi.so
artifacts/aarch64-apple-darwin/libsargon_uniffi.dylib
```
```
// Now changed to
artifacts/libsargon_uniffi.dylib
artifacts/libsargon_uniffi.so
```

So gradle's scripts could not pick up the correct files. It seems that the latest main of the download job is unwrapping in place without extracting the binaries in their respective directory

PR #372, opened 2025-02-05
[ABW-4051] API to perform Spot Check

matiasbzurovski:
## Changelog
- Adds a new `fn perform_spot_check(&self, input: SpotCheckInput) -> bool` implemented by `FactorSourceIDFromHash`, `FactorSource` & `FactorSourceID`. This determines whether the associated id was created with the same `input` than provided.
- Adds a new function on `SargonOS` to trigger a Factor Source spot check. This will use a new host interactor that will be responsible for presenting the corresponding UI for user to perform the check. To do so, it will use the function described on previous point.

### DIFF
--- a/crates/factors/factors/src/factor_source_id_spot_check.rs
+++ b/crates/factors/factors/src/factor_source_id_spot_check.rs
@@ -1,0 +1,227 @@
+use crate::prelude::*;
+
+/// An enum with the input to perform a spot check for a given `FactorSourceID`.
+/// This is, to validate that the `FactorSourceID` was created with the same input that has been provided.
+#[derive(Clone, Debug, PartialEq, Eq, std::hash::Hash)]
+pub enum SpotCheckInput {
+    /// The user retrieved the id of a Ledger device.
+    /// Used for the identification of `LedgerHardwareWalletFactorSource`.
+    Ledger { id: Exactly32Bytes },
+
+    /// The user retrieved the `FactorSourceIdFromHash` that identified an Arculus card.
+    /// /// Used for the identification of `ArculusCardFactorSource`.
+    ArculusCard { id: FactorSourceIDFromHash },
+
+    /// The user retrieved a `MnemonicWithPassphrase`.
+    /// Used for the identification of any software `FactorSource`.
+    Software {
+        mnemonic_with_passphrase: MnemonicWithPassphrase,
+    },
+}
+
+pub trait FactorSourceIDSpotCheck {
+    /// Performs a spot check and returns whether the `FactorSourceID` was created with the same input that has been provided.
+    fn perform_spot_check(&self, input: SpotCheckInput) -> bool;
+}
+
+impl FactorSourceIDSpotCheck for FactorSourceID {
+    fn perform_spot_check(&self, input: SpotCheckInput) -> bool {
+        let Some(id_from_hash) = self.as_hash().cloned() else {
+            panic!("Address FactorSourceID not supported for spot check")
+        };
+        let kind = id_from_hash.kind;
+        match input {
+            SpotCheckInput::Ledger { id } => {
+                if kind != FactorSourceKind::LedgerHQHardwareWallet {
+                    return false;
+                }
+                let built_id = FactorSourceIDFromHash::new(kind, id);
+                built_id == id_from_hash
+            }
+            SpotCheckInput::ArculusCard { id } => {
+                if kind != FactorSourceKind::ArculusCard {
+                    return false;
+                }
+                id == id_from_hash
+            }
+            SpotCheckInput::Software {
+                mnemonic_with_passphrase,
+            } => {
+                if !kind.is_software() {
+                    return false;
+                }
+                let built_id =
+                    FactorSourceIDFromHash::from_mnemonic_with_passphrase(
+                        kind,
+                        &mnemonic_with_passphrase,
+                    );
+                built_id == id_from_hash
+            }
+        }
+    }
+}
+
+impl FactorSourceIDSpotCheck for FactorSource {
+    fn perform_spot_check(&self, input: SpotCheckInput) -> bool {
+        self.id().perform_spot_check(input)
+    }
+}
+
+impl FactorSourceIDSpotCheck for FactorSourceIDFromHash {
+    fn perform_spot_check(&self, input: SpotCheckInput) -> bool {
+        FactorSourceID::from(*self).perform_spot_check(input)
+    }
+}
+
+impl FactorSourceKind {
+    fn is_software(&self) -> bool {
+        match self {
+            FactorSourceKind::Device
+            | FactorSourceKind::OffDeviceMnemonic
+            | FactorSourceKind::Password
+            | FactorSourceKind::SecurityQuestions => true,
+            FactorSourceKind::LedgerHQHardwareWallet
+            | FactorSourceKind::ArculusCard
+            | FactorSourceKind::TrustedContact => false,
+        }
+    }
+}
+
+#[cfg(test)]
+#[allow(non_snake_case)]
+mod tests {
+    use super::*;
+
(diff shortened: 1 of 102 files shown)


CyonAlexRDX (approved):

micbakos-rdx (approved):

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/sargon_os_factors.rs
```diff
@@ -216,4 +216,15 @@ impl SargonOS {
...
             .into_result()
     }
+
+    /// Triggers the spot check for the given factor source, and returns whether the spot check was successful.
+    pub async fn trigger_spot_check(
+        &self,
+        factor_source: FactorSource,
+    ) -> Result<bool> {
+        self.wrapped
+            .trigger_spot_check(factor_source.into_internal(), false)
```
matiasbzurovski, line 226: When the spot check is triggered by the hosts, it will never allow skip

crates/factors/factors/src/factor_source_id_spot_check.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+impl FactorSourceIDSpotCheck for FactorSourceIDFromHash {
+    fn perform_spot_check(&self, input: SpotCheckInput) -> bool {
+        FactorSourceID::from(*self).perform_spot_check(input)
+    }
+}
+
+impl FactorSourceKind {
+    fn is_software(&self) -> bool {
```
Sajjon, line 77: Iwould like a doc here. this is to say that **derivation** takes place in Radix controlled software? (sargon / hosts)?

because it is not about STORAGE

matiasbzurovski: I will add some docs and rename to something more descriptive, but this is just to validate that the `FactorSourceKind` is one with which we should use `SpotCheckInput::Software` to validate its id.

I am actually wondering if these checks (lines `35`, `42` & `50`) are actually worthy.. because the `FactorSourceIDFromHash` built when using incorrect combination of `SpotCheckInput` & `FactorSourceKind` would actually result in a different `FactorSourceIDFromHash` than the original one (unless the original one was incorrectly built)

CyonAlexRDX (approved):

crates/factors/factors/src/factor_source_id_spot_check.rs
```diff
@@ -0,0 +1,234 @@
...
+/// An enum with the input to perform a spot check for a given `FactorSourceID`.
+/// This is, to validate that the `FactorSourceID` was created with the same input that has been provided.
+#[derive(Clone, Debug, PartialEq, Eq, std::hash::Hash)]
+pub enum SpotCheckInput {
+    /// The user retrieved the id of a Ledger device.
+    /// Used for the identification of `LedgerHardwareWalletFactorSource`.
+    Ledger { id: Exactly32Bytes },
+
+    /// The user retrieved the `FactorSourceIdFromHash` that identified an Arculus card.
+    /// /// Used for the identification of `ArculusCardFactorSource`.
```
GhenadieVP, line 12: Double `///`

crates/factors/factors/src/factor_source_id_spot_check.rs
```diff
@@ -0,0 +1,234 @@
...
+    fn perform_spot_check(&self, input: SpotCheckInput) -> bool {
+        let Some(id_from_hash) = self.as_hash().cloned() else {
+            panic!("Address FactorSourceID not supported for spot check")
+        };
+        let kind = id_from_hash.kind;
+        match input {
+            SpotCheckInput::Ledger { id } => {
+                if kind != FactorSourceKind::LedgerHQHardwareWallet {
+                    return false;
+                }
```
GhenadieVP, line 37: It would be good to have an error of sort to identify this scenario, as it would be a programmer error. Maybe a CommonError would be useful.

crates/factors/factors/src/factor_source_id_spot_check.rs
```diff
@@ -0,0 +1,234 @@
...
+
+impl FactorSourceIDSpotCheck for FactorSource {
+    fn perform_spot_check(&self, input: SpotCheckInput) -> bool {
+        self.id().perform_spot_check(input)
+    }
+}
+
+impl FactorSourceIDSpotCheck for FactorSourceIDFromHash {
+    fn perform_spot_check(&self, input: SpotCheckInput) -> bool {
+        FactorSourceID::from(*self).perform_spot_check(input)
```
GhenadieVP, line 72: Maybe instead move the whole code starting from line 32 here? since that is specific to FactorSourceIDFromHash.

crates/factors/factors/src/factor_source_id_spot_check.rs
```diff
@@ -0,0 +1,234 @@
...
+    /// Returns whether the kind expects a `Software` input for spot check.
+    fn expects_software_spot_check_input(&self) -> bool {
+        match self {
+            FactorSourceKind::Device
+            | FactorSourceKind::OffDeviceMnemonic
+            | FactorSourceKind::Password
+            | FactorSourceKind::SecurityQuestions => true,
+            FactorSourceKind::LedgerHQHardwareWallet
+            | FactorSourceKind::ArculusCard
+            | FactorSourceKind::TrustedContact => false,
```
GhenadieVP, line 86: This should be software I think.

crates/system/interactors/src/spot_check_interactor.rs
```diff
@@ -0,0 +1,27 @@
...
+    async fn spot_check(
+        &self,
+        factor_source: FactorSource,
+        allow_skip: bool,
+    ) -> Result<SpotCheckResponse>;
+}
+
+/// An enum indicating the result of a spot check.
+///
+/// Note that there isn't a failure case since user never fails a spot check: it either succeeds,
```
GhenadieVP, line 18: For clarification - abort will be error thrown from the host side, HostInteractionAborted is the error

crates/system/interactors/src/testing/test_spot_check_interactor.rs
```diff
@@ -0,0 +1,45 @@
...
+            SpotCheckUser::Failure(common_error) => Err(common_error),
+            SpotCheckUser::Valid => Ok(SpotCheckResponse::Valid),
+            SpotCheckUser::Skipped => Ok(SpotCheckResponse::Skipped),
+        }
+    }
+}
+
+#[derive(Clone)]
+pub enum SpotCheckUser {
+    Failure(CommonError),
```
Sajjon, line 24: `Skipped` is past tense of a **verb**, `Valid` and `Failure` are not verbs. Pick a part of speech kind and use for all variants? :)

crates/factors/factors/src/factor_source_id_spot_check.rs
matiasbzurovski: FactorSourceIDFromAddress is not supported for spot check so doesn't make a difference

crates/system/interactors/src/spot_check_interactor.rs
matiasbzurovski: updating docs

crates/factors/factors/src/factor_source_id_spot_check.rs
GhenadieVP: Sure, but still it is not hardware kind.

GhenadieVP (approved):

crates/factors/factors/src/factor_source_id_spot_check.rs
matiasbzurovski: notice that the function has been renamed a few commits ago into `expects_software_spot_check_input`, which for `TrustedContact` isn't true since it doesn't expect any spot check input, but I can change if you prefer

PR #373, opened 2025-02-06
RET manfiest classification update

GhenadieVP:
Update to use new RET manifest summary models.

For now just mapped new models to current Sargon structure with the intent to reduce the number of changes as much as possible.

A followup PR will add additional fields/models to Sargon's ManifestSummary/ExecutionSummary to make use of new RET summary surfaced information.


crates/crypto/addresses/src/address/resource_address.rs
```diff
@@ -38,6 +38,20 @@ impl ResourceAddress {
     }
 }
 
+impl TryFrom<(ScryptoManifestResourceAddress, NetworkID)> for ResourceAddress {
+    type Error = CommonError;
+    fn try_from(
+        value: (ScryptoManifestResourceAddress, NetworkID),
```
Sajjon, line 44: an alternative spelling is: `(address, network): (ScryptoManifestResourceAddress, NetworkID)` which I've started to use

crates/transaction/models/src/low_level/manifest_encountered_component_address.rs
```diff
@@ -10,20 +10,19 @@ address_union!(
...
+                let (address, network_id) = value;
 
                 $(
-                    if let Ok(address) = <$address_type>::try_from((global_address, network_id)) {
+                    if let Ok(address) = <$address_type>::try_from(value) {
                         return Ok(ManifestEncounteredComponentAddress::$variant(address));
                     }
                 )*
-
                 Err(CommonError::FailedToCreateAddressFromGlobalAddressAndNetworkID {
```
Sajjon, line 24: Maybe rename the error to `FailedToCreateAddressFromManifestAddressAndNetworkID { address: address.to_string() }` or `to_hex`

Cargo.toml
```diff
@@ -370,6 +370,7 @@ iso8601-timestamp = { version = "0.2.16", default-features = false, features = [
 ] }
 
 itertools = { version = "0.12.0", default-features = true }
+either = { version = "1.13.0", default-features = false }
```
CyonAlexRDX, line 373: Im really not a fan of `Either` as `::Left` gives zero glue as to what it is. Can we create a custom enum instead?

crates/transaction/models/src/low_level/execution_summary/execution_summary.rs
```diff
@@ -30,6 +30,23 @@ where
         .collect::<HashMap<L, U>>()
 }
 
+pub(crate) fn filter_try_to_hashmap_network_aware_key<K, V, L, U>(
+    values: impl IntoIterator<Item = (K, V)>,
+    network_id: NetworkID,
+) -> HashMap<L, U>
```
CyonAlexRDX, line 36: should we return `IndexMap` instead for stable sorting helpful when debugging?

crates/transaction/models/src/low_level/execution_summary/execution_summary.rs
```diff
@@ -79,7 +96,7 @@ pub struct ExecutionSummary {
 
     /// The set of instructions encountered in the manifest that are reserved
     /// and can only be included in the manifest by the wallet itself.
-    pub reserved_instructions: Vec<ReservedInstruction>,
+    pub reserved_instructions: IndexSet<ReservedInstruction>,
```
CyonAlexRDX, line 99: 👍 ah nice with `IndexSet`! post break out UniFFI allows us to do this, excellent!

Cargo.toml
GhenadieVP: I am using Either as it is part of iter `partition_map` API, might need to revisit then that implementation. Either is good for temporary types, similar to tuples.

crates/transaction/models/src/low_level/execution_summary/execution_summary.rs
GhenadieVP: yeah right, may be a good time to use IndexMap instead of HashMap where possible

crates/transaction/models/src/low_level/transaction_classes/detailed_manifest_class.rs
```diff
@@ -146,142 +154,214 @@ impl From<(RetDetailedManifestClass, NetworkID)> for DetailedManifestClass {
...
+                        ResourceAddress::try_from((resource, n));
+                    let update = ResourcePreferenceUpdate::from(v);
+                    if let (Ok(account_address), Ok(resource_address)) =
+                        (account_address, resource_address)
+                    {
+                        acc.entry(account_address)
+                            .or_default()
+                            .insert(resource_address, update);
+                    };
+                    acc
```
CyonAlexRDX, line 320: if any is err. we effectively filter it out without error? should be log at least?

crates/transaction/models/src/low_level/transaction_classes/detailed_manifest_class.rs
```diff
@@ -146,142 +154,214 @@ impl From<(RetDetailedManifestClass, NetworkID)> for DetailedManifestClass {
...
+        ) = output
+            .authorized_depositor_updates
+            .into_iter()
+            .filter_map(|((account, manifest_resource), op)| {
+                let account_address =
+                    AccountAddress::try_from((account, n)).ok()?;
+                let resource =
+                    ResourceOrNonFungible::try_from((manifest_resource, n))
+                        .ok()?;
+                Some(((account_address, resource), op))
```
CyonAlexRDX, line 336: log if err?

CyonAlexRDX (approved):
Great job! nothing major at all

crates/transaction/models/src/low_level/transaction_classes/detailed_manifest_class.rs
GhenadieVP: This can fail only if there are named manifest addresses, which now are simply ignored.

PR #374, opened 2025-02-06
Add shield for display API

giannis-rdx:
This PR is the last part of the [ABW-4146](https://radixdlt.atlassian.net/browse/ABW-4146).

It adds an API to 
- get entities linked to a given security structure
- get shields for display

[ABW-4146]: https://radixdlt.atlassian.net/browse/ABW-4146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/profile/logic/logic_SPLIT_ME/src/logic/profile_network/profile_network_entities_linked_to_security_structure.rs
```diff
@@ -0,0 +1,44 @@
...
+    ) -> Result<EntitiesLinkedToSecurityStructure>;
+}
+
+impl ProfileNetworkQueryEntitiesLinkedToSecurityStructure for ProfileNetwork {
+    /// Returns the entities linked to a given `SecurityStructure` on the current `ProfileNetwork`.
+    fn entities_linked_to_security_structure(
+        &self,
+        shield_id: SecurityStructureID,
+    ) -> Result<EntitiesLinkedToSecurityStructure> {
+        let accounts = self
```
CyonAlexRDX, line 16: nit: does this work:

```rust
fn filter<E>(entity: &E) -> bool where E:  HasSecurityState { entity.is_linked_to_security_structure(shield_id) }
```
and then `.filter(filter)` in the four places below.

or does compiler complain about `shield_id` being captured?

can you do `let filter = |entity: &(impl HasSecurityState)| { .. }; ? or does that not work?

mostly curiuos...

crates/profile/models/base-entity/src/base_entity/has_security_state.rs
```diff
@@ -52,4 +52,37 @@ pub trait HasSecurityState: HasFactorInstances + IsSecurityStateAware {
...
+            EntitySecurityState::Securified { value } => {
+                if let Some(
+                    ProvisionalSecurifiedConfig::FactorInstancesDerived {
+                        value: provisional_security_structure,
+                    },
+                ) = &value.provisional_securified_config
+                {
+                    return provisional_security_structure
+                        .security_structure_id
+                        == shield_id;
```
CyonAlexRDX, line 82: undefined behaviour here... if the provisional is different, then we return `false` here, even though the actual ("committed") **is** linked. So you need to doc this method what is it suppose to do? what does `linked` mean? I think linked means "provisionally_referenced_or_de_facto_in_use" ?

crates/profile/logic/logic_SPLIT_ME/src/logic/profile_network/profile_network_entities_linked_to_security_structure.rs
matiasbzurovski: [this ](https://github.com/radixdlt/sargon/blob/main/crates/profile/logic/logic_SPLIT_ME/src/logic/profile_network/profile_network_entities_linked_to_factor_source.rs#L18-L24) might help as guide

crates/uniffi/uniffi_SPLIT_ME/src/profile/supporting_types/shield_for_display.rs
danvleju-rdx, line 1: Missing module declaration in mod.rs file

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/sargon_os_entities_linked_to_security_structure.rs
danvleju-rdx, line 1: Missing module declaration in mod.rs file

crates/system/os/factors/src/sargon_os_security_structures.rs
```diff
@@ -185,6 +193,63 @@ impl OsSecurityStructuresQuerying for SargonOS {
...
+                let profile_network = specific_profile
+                    .networks
+                    .get_id(NetworkID::Mainnet)
+                    .ok_or(CommonError::Unknown)?;
+                profile_network.entities_linked_to_security_structure(shield_id)
+            }
+        }
+    }
+
+    async fn get_shields_for_display(&self) -> Result<ShieldsForDisplay> {
```
danvleju-rdx, line 217: Should be UNIFFI exported, right?

crates/profile/logic/logic_SPLIT_ME/src/logic/profile_network/profile_network_entities_linked_to_security_structure.rs
CyonAlexRDX: DRY 

But only truly helps IMO if one can pass `.filter(filter)` which is possible in Swift, but trickier in Rust since the fn or closure captures the `shield_id` variable

crates/profile/models/base-entity/src/base_entity/has_security_state.rs
Sajjon: I would actually split this into three methods:

```rust
trait HasSecurityState {
fn is_currently_or_provisionally_securified_with(&self, shield_id: SecurityStructureID) {
    self.is_currently_securified_with(shield_id) || self.is_provisionally_securified_with(shield_id)
}

fn is_currently_securified_with(&self, shield_id: SecurityStructureID) {
    self.security_state().as_securified().map(|s| s.shield_id == shield_id).unwrap_or(false)
}

fn is_provisionally_securified_with(&self, shield_id: SecurityStructureID) {
    self.provisional_security_config().map(|s| s.shield_id() == shield_id).unwrap_or(false)
}
}
```

```
impl ProvisionalSecurityConfig {
    fn shield_id(&self) -> bool {
       match self { ... }
    }
}
```

crates/profile/models/app-preferences/src/shield_for_display.rs
```diff
@@ -0,0 +1,96 @@
...
+    Eq,
+    Clone,
+    Hash,
+    derive_more::Display,
+)]
+#[serde(rename_all = "camelCase")]
+#[display("{}", self.metadata.display_name)]
+pub struct ShieldForDisplay {
+    pub metadata: SecurityStructureMetadata,
+    pub number_of_linked_accounts: usize,
```
Sajjon, line 45: this can probably be put in `profile-logic` crate. it is not a model used by **the Profile**.

crates/profile/models/base-entity/src/base_entity/has_security_state.rs
Sajjon: @giannis-rdx furthermore, this should be in `profile-logic` crate. **All Profile logic goes into `profile-logic` crate** :)

crates/profile/models/profile_SPLIT_ME/src/v100/entities_linked_to_security_structure/entities_linked_to_security_structure.rs
```diff
@@ -0,0 +1,72 @@
+use crate::prelude::*;
+
+/// This is the result of checking what entities are linked to a given `SecurityStructure`.
+#[derive(Clone, Debug, PartialEq)]
+pub struct EntitiesLinkedToSecurityStructure {
```
Sajjon, line 5: Not **Profile model** so move to `profile-logic` crate I think.

Sajjon: Furthermore, this model does not contain **which Shield ID**, which looks like an oversight! If you include it the code later can be more functional.

crates/system/os/factors/src/sargon_os_security_structures.rs
```diff
@@ -185,6 +193,63 @@ impl OsSecurityStructuresQuerying for SargonOS {
...
+        let mut shields_for_display = ShieldsForDisplay::new();
+        for security_structure in security_structures.iter() {
+            let linked_entities = self
+                .entities_linked_to_security_structure(
+                    security_structure.id(),
+                    ProfileToCheck::Current,
+                )
+                .await?;
+
+            shields_for_display.insert(ShieldForDisplay {
```
Sajjon, line 230: Should use constructor `ShieldForDisplay::new` also prefer functional:

```rust
let security_structures = self.security_structures_of_factor_source_ids()?;
let shields_for_display = security_structures
    .iter()
    .map(|shield| shield.metadata) 
    .map(|metadata| self.entities_linked_to_security_structure(metadata))
    .map(ShieldForDisplay::new)
    .collect::<ShieldsForDisplay>();

Ok(shields_for_display)
```

which I think is much more clean

Sajjon (commented):
Great first start! Some crate admin needed. Logic into logic :)

crates/profile/models/base-entity/src/base_entity/has_security_state.rs
GhenadieVP: > if the provisional is different, then we return false here, even though the actual ("committed") is linked

That is the behaviour we want, or at least agreed to - if an account has a provisional shield set, then that is shown throughout the app. What you are proposing implies showing users that an account is linked to two shields at the same time, no?

crates/profile/logic/logic_SPLIT_ME/src/logic/profile_network/profile_network_entities_linked_to_security_structure.rs
```diff
@@ -0,0 +1,44 @@
...
+pub trait ProfileNetworkQueryEntitiesLinkedToSecurityStructure {
+    fn entities_linked_to_security_structure(
+        &self,
+        shield_id: SecurityStructureID,
+    ) -> Result<EntitiesLinkedToSecurityStructure>;
+}
+
+impl ProfileNetworkQueryEntitiesLinkedToSecurityStructure for ProfileNetwork {
+    /// Returns the entities linked to a given `SecurityStructure` on the current `ProfileNetwork`.
+    fn entities_linked_to_security_structure(
```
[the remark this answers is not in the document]
giannis-rdx: but it is in the profile-logic crate, isn't it? 
It is in the same directory with the `profile_network_entities_linked_to_factor_source`

crates/profile/models/app-preferences/src/shield_for_display.rs
giannis-rdx: I agree with you on that. 
However, the **reasons** I placed it there are as follows:
1. there is already an existing `account_for_display` which also is not used by the `Profile`. For consistency, I added `shield_for_display` here.
2. I am not sure where exactly I should place it in the `profile-logic` crate. We do not have a `model` directory there... Oops! model ≠ logic :/

Sajjon: `account_for_display` should be moved to `profile-logic` too. :)  The "logic" in `profile-logic` does not mean "cannot contain structs or enums". It means "here is code that DOES stuff using models from Profile models crate OR using models declared in this crate". Where as profile model crates should as much as possible just be the Serde impl of Profile model. Without logic. Why? Well since we **might need to migrate v100 Profile to v101 Profile someday** and then we need to copy paste all profile models and do some `trait IntoLatest` and impl that for v100 models. But we **should not** duplicate logic, neither should the logic operate on `v100` models then, only `v101` models.

giannis-rdx: ok where do you think we should place it in `profile-logic`?
under `app_preferences`?
or create a new directory `security_structures` and place it there?

crates/profile/models/profile_SPLIT_ME/src/v100/entities_linked_to_security_structure/entities_linked_to_security_structure.rs
giannis-rdx: >Not Profile model so move to profile-logic crate I think.

now that i see this then better to have a `security_structures` in profile-logic crate and place the `EntitiesLinkedToSecurityStructure` and `ShieldForDisplay` there.

right?

crates/profile/models/base-entity/src/base_entity/has_security_state.rs
Sajjon: @GhenadieVP my point is that `is_linked` is too broad a term. so the method should be renamed and it should it should be documented.

crates/profile/logic/logic_SPLIT_ME/src/logic/profile_network/profile_network_entities_linked_to_security_structure.rs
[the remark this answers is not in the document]
Sajjon: oh my bad, yes in correct crate :) I

crates/profile/models/app-preferences/src/shield_for_display.rs
Sajjon: fine to have it in `app_preferences` I think, since that is where the shield ids are stored.

crates/profile/models/profile_SPLIT_ME/src/v100/entities_linked_to_security_structure/entities_linked_to_security_structure.rs
Sajjon: sure

giannis-rdx: [done](https://github.com/radixdlt/sargon/pull/374/commits/f1daea94e2b0a784135f74d55cc18b6a0f1f81cf)

crates/profile/models/app-preferences/src/shield_for_display.rs
giannis-rdx: [done](https://github.com/radixdlt/sargon/pull/374/commits/f1daea94e2b0a784135f74d55cc18b6a0f1f81cf)

crates/system/os/factors/src/sargon_os_security_structures.rs
giannis-rdx: although I updated the `EntitiesLinkedToSecurityStructure` struct by adding the  `metadata: SecurityStructureMetadata`, this functional code doesn't work because `map` doesn't support `async`

Unless there is another way that I am not aware of.

crates/profile/models/profile_SPLIT_ME/src/v100/entities_linked_to_security_structure/entities_linked_to_security_structure.rs
giannis-rdx: >Furthermore, this model does not contain which Shield ID, which looks like an oversight! If you include it the code later can be more functional.
>If you include the SecurityStructureMetadata ...

if I include the `SecurityStructureMetadata` in the `EntitiesLinkedToSecurityStructure`, I need:
- **either** update the API 
```
pub async fn entities_linked_to_security_structure(
        &self,
        metadata: SecurityStructureMetadata, // from shield_id: SecurityStructureID
        profile_to_check: ProfileToCheck,
    )
// so I can pass the metadata all the way down to build the `EntitiesLinkedToSecurityStructure` struct.
```
- **or** leave the API as it is with `shield_id: SecurityStructureID` and internally fetch the `SecurityStructureMetadata`.

I think the latter is more appropriate. 
what do you think @CyonAlexRDX ?

crates/system/os/factors/src/sargon_os_security_structures.rs
Sajjon: @giannis-rdx you should be able to use `futures::future::join_all`, so:

```rust
let security_structures = self.security_structures_of_factor_source_ids()?;
let shields_for_display = security_structures
    .iter()
    .map(|shield| shield.metadata) 
    .map(|metadata| self.entities_linked_to_security_structure(metadata, ProfileToCheck::Current))
    .map(ShieldForDisplay::new)

let shields_for_display = futures::future::join_all(shields_for_display)
     .await
    .collect::<ShieldsForDisplay>();

Ok(shields_for_display)
```

I think that should work.

See [join_all](https://docs.rs/futures/latest/futures/future/fn.join_all.html), which does preserve order:
> The returned future will drive execution for all of its underlying futures, collecting the results into a destination Vec<T> in the same order as they were provided.

giannis-rdx: no it doesn't work. I get a couple of errors:
```
function is expected to take 1 argument, but it takes 5 arguments
.map(ShieldForDisplay::new);

Vec<_> is not an iterator
   --> crates/system/os/factors/src/sargon_os_security_structures.rs:262:14
    |
260 |           let shields_for_display = futures::future::join_all(shields_for_display)
```

giannis-rdx: the only way to make it work is this:

```rust
let security_structures = self.security_structures_of_factor_source_ids()?;

        let shields_for_display_futures = security_structures.iter().map(|security_structure| {
                let metadata = security_structure.metadata.clone();

                async {
                    let linked_entities = self
                        .entities_linked_to_security_structure(
                            metadata.id,
                            ProfileToCheck::Current,
                        )
                        .await?;

                    Ok(ShieldForDisplay::new(
                        metadata,
                        linked_entities.accounts.len(),
                        linked_entities.hidden_accounts.len(),
                        linked_entities.personas.len(),
                        linked_entities.hidden_personas.len(),
                    ))
                }
            });

        let shields_for_display =
            futures::future::join_all(shields_for_display_futures)
                .await
                .into_iter()
                .collect::<Result<ShieldsForDisplay>>()?;

        Ok(shields_for_display)
    }
```

which is quite ugly.

giannis-rdx: I think I will stick to my original solution.

CyonAlexRDX (approved):
LGTM! Hope `join_all` works as I thought, else lets skip it.

crates/system/os/factors/Cargo.toml
```diff
@@ -32,3 +32,4 @@ preinterpret = { workspace = true }
 pretty_assertions = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+futures = "0.3.31"
```
CyonAlexRDX, line 35: Use workspace, always

danvleju-rdx (approved):

PR #375, opened 2025-02-07
Expose securified security state and expose methods that relied on unsecured entity control on hosts

micbakos-rdx:
* Export a method that checks if an account is legacy olympia.
* Export a method that returns the unsecured controlling factor instance for accounts and personas.
* The BDFS error check, as discussed, is not needed so I removed the implementation I temporarily wrote in sargon.
* Expose secured entity control to hosts.


crates/profile/logic/logic_SPLIT_ME/src/logic/account/account_is_legacy.rs
```diff
@@ -0,0 +1,91 @@
+use crate::prelude::*;
+
+pub trait AccountIsLegacyOlympia {
+    fn is_legacy_olympia(&self) -> bool;
+}
+
+impl AccountIsLegacyOlympia for Account {
+    fn is_legacy_olympia(&self) -> bool {
+        let Some(unsecured_entity_control) = self.security_state.as_unsecured()
```
CyonAlexRDX, line 9: You can just `self.address.is_legacy()`

crates/profile/logic/logic_SPLIT_ME/src/logic/account/account_is_unsecured_ledger_controlled.rs
```diff
@@ -0,0 +1,80 @@
+use crate::prelude::*;
+
+pub trait EntityUnsecuredControllingFactorInstance {
+    fn unsecured_controlling_factor_instance(
+        &self,
+    ) -> Option<HierarchicalDeterministicFactorInstance>;
+}
+
+impl EntityUnsecuredControllingFactorInstance for Account {
+    fn unsecured_controlling_factor_instance(
```
CyonAlexRDX, line 10: I want to avoid this.

Why are we adding it? 

Temporarily as part of a process of transitioning?

As so said the other day, I don't want hosts to concern themselves with FactorInstances / PublicKeys **at all**

CyonAlexRDX (changes requested):

crates/profile/logic/logic_SPLIT_ME/src/logic/account/account_is_unsecured_ledger_controlled.rs
micbakos-rdx: > Temporarily as part of a process of transitioning?

Yes exactly for this reason. The refactoring effort is too much  for now.

CyonAlexRDX (approved):

crates/uniffi/uniffi_SPLIT_ME/src/profile/v100/entity_security_state/entity_security_state.rs
```diff
@@ -23,8 +27,10 @@ impl From<InternalEntitySecurityState> for EntitySecurityState {
                     value: value.into(),
                 }
             }
-            InternalEntitySecurityState::Securified { value: _ } => {
-                panic!("Securified state not yet supported in the Wallet")
+            InternalEntitySecurityState::Securified { value } => {
```
GhenadieVP, line 30: With this added you should be able to use InternalConversion macro

micbakos-rdx: Right will do

GhenadieVP (approved):

GhenadieVP (approved):

PR #377, opened 2025-02-10
Main Security Shield updates

danvleju-rdx:
- When adding a new security shield, set it as `Main` if no main shield exists
- Expose `is_Main` from `SecurityStructureMetadata` to hosts
- Expose `security_structure_of_factor_source_ids_by_security_structure_id` to hosts

### DIFF
--- a/crates/system/os/factors/src/sargon_os_security_structures.rs
+++ b/crates/system/os/factors/src/sargon_os_security_structures.rs
@@ -155,6 +155,17 @@ impl OsSecurityStructuresQuerying for SargonOS {
                 EventProfileModified::SecurityStructureAdded { id },
             ))
             .await;
+
+        if !self
+            .profile()?
+            .app_preferences
+            .security
+            .security_structures_of_factor_source_ids
+            .iter()
+            .any(|s| s.metadata.is_main())
+        {
+            self.set_main_security_structure(id).await?;
+        }
         Ok(())
     }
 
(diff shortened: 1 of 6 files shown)


crates/system/os/factors/src/sargon_os_security_structures.rs
```diff
@@ -155,6 +155,17 @@ impl OsSecurityStructuresQuerying for SargonOS {
                 EventProfileModified::SecurityStructureAdded { id },
             ))
             .await;
+
+        if !self
+            .profile()?
+            .app_preferences
+            .security
+            .security_structures_of_factor_source_ids
```
CyonAlexRDX, line 163: I think this might be useful to have as a reusable method. In `profile-logic` crate:

```rust
impl Profile { fn has_any_main_security_structure(&self) -> bool }
```
?

CyonAlexRDX: dispatching/forwarding calls to `AppPreferences` -> `Security` -> `SecurityStructure` (fn `is_main(&self)` on SecStruct).

crates/uniffi/uniffi_SPLIT_ME/src/profile/mfa/security_structures/security_structure_metadata.rs
```diff
@@ -40,3 +40,10 @@ pub fn new_security_structure_metadata_named(
     )
     .into()
 }
+
+#[uniffi::export]
+pub fn security_structure_is_main(
```
matiasbzurovski, line 45:
```suggestion
pub fn security_structure_metadata_is_main(
```

crates/profile/logic/logic_SPLIT_ME/src/logic/profile_security_structure.rs
```diff
@@ -0,0 +1,17 @@
+use crate::prelude::*;
+
+pub trait ProfileHasAnyMainSecurityStructure {
+    /// Returns true if the profile has any main security structure.
+    fn has_any_main_security_structure(&self) -> bool;
+}
+
+impl ProfileHasAnyMainSecurityStructure for Profile {
```
matiasbzurovski, line 8: I believe @CyonAlexRDX  idea is to implement the logic on `Security`, then also on `AppPreferences` (which will call `Security` impl) and finally on `Profile` (which will call `AppPreferences` impl)


This way you have it on all 3 levels "for free". See this as [ref](https://github.com/radixdlt/sargon/blob/main/crates/profile/logic/logic_SPLIT_ME/src/logic/app_preferences/security_structures_update.rs)

matiasbzurovski (approved):

crates/profile/logic/logic_SPLIT_ME/src/logic/profile_security_structure.rs
giannis-rdx, line 1: imo this can go under the `app_preferences/security_structures/`

danvleju-rdx: agree 👍

giannis-rdx (approved):

PR #380, opened 2025-02-12
Adding new factor source

sergiupuhalschi-rdx:
> [!NOTE]  
> Adding doc & tests is WIP

Defined a `uniffi::Object` called `DeviceFactorAddingManager` (name alternatives are welcome), exposing functionality required for the new UI flow for adding a device factor source. We can define other objects similarly for other factor source kinds.


crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,308 @@
...
+struct FactorIdentification {
+    mnemonic_with_passphrase: MnemonicWithPassphrase,
+    factor_source: DeviceFactorSource,
+    mnemonic_words: Vec<BIP39Word>,
+    confirmation_indices: Vec<u8>,
+}
+
+impl std::hash::Hash for FactorIdentification {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.mnemonic_with_passphrase.hash(state);
```
CyonAlexRDX, line 19: Why manually Impl Hash, why not derive?

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,308 @@
...
+    os_ref: Arc<dyn OsNewFactorAdding>,
+    factor_identification: RwLock<Option<FactorIdentification>>,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+struct FactorIdentification {
+    mnemonic_with_passphrase: MnemonicWithPassphrase,
+    factor_source: DeviceFactorSource,
+    mnemonic_words: Vec<BIP39Word>,
+    confirmation_indices: Vec<u8>,
```
CyonAlexRDX, line 14: Indices should be `usize`. Duplicates not allowed right? So should be IndexSet<usize>?

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,308 @@
...
+impl FactorIdentification {
+    const NUMBER_OF_CONFIRMATION_WORDS: u8 = 4;
+
+    fn new(mnemonic: Mnemonic, host_info: HostInfo) -> Self {
+        let mnemonic_words = mnemonic.clone().words;
+        let number_of_words = mnemonic_words.clone().len() as u8;
+        let mnemonic_with_passphrase =
+            MnemonicWithPassphrase::new(mnemonic.clone());
+        let factor_source = DeviceFactorSource::babylon(
+            false,
```
CyonAlexRDX, line 78: Can you `decl_bool_type` for this bool? Is it if it is Main?

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,308 @@
...
+            ),
+        }
+    }
+
+    fn generate_confirmation_indices(number_of_words: u8) -> Vec<u8> {
+        let max_confirmation_index = number_of_words - 1;
+        let mut csprng = OsRng;
+        let mut confirmation_indices: HashSet<u8> = HashSet::new();
+
+        while (confirmation_indices.len() as u8)
```
CyonAlexRDX, line 98: Instead of this you can use:

`generate_bytes::<3>().into_iter().map(|b| b as usize).collect::<IndexSet<_>>()`

And done.

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,308 @@
...
+            Some(FactorIdentification::new(mnemonic, host_info))
+    }
+}
+
+impl DeviceFactorAddingManager {
+    /// Returns true if the mnemonic `word` at the given `index` is correct.
+    pub fn is_word_at_index_correct(
+        &self,
+        word: impl AsRef<str>,
+        index: u8,
```
CyonAlexRDX, line 190: Indices  should be  usize

CyonAlexRDX: Also index is ambiguous in context of BIP39 I think. Call it `index_in_mnemonic` (as opposed to index in BIP39 wordlist)

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
sergiupuhalschi-rdx: That's right. I just wanted to avoid conversion when exposing through `uniffi`, as all this code is meant to be used only by the hosts and not internally. The benefit of having a more suitable type compared to the conversion effort seemed not worth it.

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
sergiupuhalschi-rdx: I think this will uncover the case when the same value is generated, with the `while` we're sure we'll end up with 3 distinct values.

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
sergiupuhalschi-rdx: Agreed. I can change it to be `usize`, but I'd like your opinion on [this](https://github.com/radixdlt/sargon/pull/380#discussion_r1953976672) first, as it's kind of the same thing, we'll have to convert at `uniffi` level, although this is much less effort, but still..

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,298 @@
+use crate::prelude::*;
+use rand::{rngs::OsRng, Rng};
+
+pub struct DeviceFactorAddingManager {
+    os_ref: Arc<dyn OsNewFactorAdding>,
+    factor_identification: RwLock<Option<FactorIdentification>>,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+struct FactorIdentification {
```
GhenadieVP, line 10: So this will keep the internal state as user progresses through the flow?

sergiupuhalschi-rdx: The `DeviceFactorAddingManager` itself keeps the internal state and this is meant to reflect only the "factor identification" part of the flow as shown in the [UI design](https://www.figma.com/design/6MaKVYek0m9GKkd1xujB8P/Babylon-Wallet?node-id=28609-243055&t=mlJaacUOmEg3HLDK-0).

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
CyonAlexRDX: right, so lets do `generate_bytes::<100>().into_iter().map(|b| b as usize).take(3).collect::<IndexSet<_>>()` that will statistically never fail.

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,298 @@
...
+        let other_factor_identification = other
+            .factor_identification
+            .read()
+            .expect("Failed to read other factor identification");
+
+        *factor_identification == *other_factor_identification
+    }
+}
+
+impl Eq for DeviceFactorAddingManager {}
```
CyonAlexRDX, line 32: does the manafer need to be `Eq`?

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,298 @@
...
+pub struct DeviceFactorAddingManager {
+    os_ref: Arc<dyn OsNewFactorAdding>,
+    factor_identification: RwLock<Option<FactorIdentification>>,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+struct FactorIdentification {
+    mnemonic_with_passphrase: MnemonicWithPassphrase,
+    factor_source: DeviceFactorSource,
+    mnemonic_words: Vec<BIP39Word>,
```
CyonAlexRDX, line 13: no need to store `mnemonic_words` they are already stored in `mnemonic_with_passphrase`. Use a method instead :)

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,298 @@
...
+    confirmation_indices: Vec<u8>,
```
CyonAlexRDX, line 14: I prefer `words_to_confirm_index_in_mnemonic` over `confirmation_indices`

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,298 @@
+use crate::prelude::*;
+use rand::{rngs::OsRng, Rng};
+
+pub struct DeviceFactorAddingManager {
+    os_ref: Arc<dyn OsNewFactorAdding>,
```
CyonAlexRDX, line 5: I would name this `factor_adder` or `factor_into_profile_inserter`.

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,298 @@
...
+            .read()
+            .expect("Failed to read other factor identification");
+
+        *factor_identification == *other_factor_identification
+    }
+}
+
+impl Eq for DeviceFactorAddingManager {}
+
+impl Clone for DeviceFactorAddingManager {
```
CyonAlexRDX, line 34: do you need `Clone`? if not avoid it. If `os_ref` is a **mutable** ref to SargonOS we want as few copies around as possible - due to deadlock capability with its interior mutability (`RwLock<ProfileState>`).

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,298 @@
...
+        let factor_identification = self
+            .factor_identification
+            .read()
+            .expect("Failed to read factor identification");
+        factor_identification.hash(state);
+    }
+}
+
+impl FactorIdentification {
+    const NUMBER_OF_CONFIRMATION_WORDS: u8 = 4;
```
CyonAlexRDX, line 59: `NUMBER_OF_WORDS_OF_MNEMONIC_USER_NEED_TO_CONFIRM`

CyonAlexRDX: Prefer: 
```rust
const NUMBER_OF_WORDS_OF_MNEMONIC_USER_NEED_TO_CONFIRM_EXCL_CHECKSUM: u8 = 3;
const NUMBER_OF_WORDS_OF_MNEMONIC_USER_NEED_TO_CONFIRM: u8 = NUMBER_OF_WORDS_OF_MNEMONIC_USER_NEED_TO_CONFIRM_EXCL_CHECKSUM + 1; // +1 for checksum word (last one)
```

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
CyonAlexRDX: and when https://github.com/radixdlt/sargon/pull/380#discussion_r1954308146

it becomes:

```rust
generate_bytes::<100>()
    .into_iter()
    .map(|b| b as usize)
    .take(Self::NUMBER_OF_WORDS_OF_MNEMONIC_USER_NEED_TO_CONFIRM_EXCL_CHECKSUM)
    .collect::<IndexSet<_>>()
```

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
sergiupuhalschi-rdx: Generally speaking, I agree. The difference here might be that this code is meant to be used only by the hosts directly (similar to `AccountForDisplay`), so it might make sense to have the types prepared to be easily exported as they are, especially if the benefit is negligible compared to the conversions overhead. But anyway, I'll make the changes.

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/mod.rs
```diff
@@ -1,3 +1,4 @@
+mod add_factor;
```
CyonAlexRDX, line 1: `factor` is verboten! `add_factor_source` **everywhere** in types too. `DeviceFactorSourceAdding` etc etc - everywhere :)

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,151 @@
...
+        .await
+    }
+}
+
+#[uniffi::export]
+impl DeviceFactorAddingManager {
+    pub fn is_word_at_index_correct(
+        self: Arc<Self>,
+        word: String,
+        index: u8,
```
CyonAlexRDX, line 131: `index_in_mnemonic` and I would use `BIP39Word` here - not string.

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
sergiupuhalschi-rdx: Do we need to `uniffi` export `Eq` or only `Hash` is enough to make it ready to use within a TCA shared state? (I hope I don't mess up the iOS specific terms). @danvleju-rdx might now better since he made the changes to the `SecurityShieldBuilder` and I just applied them here as well.

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,151 @@
...
+        self.get(|manager| manager.get_confirmation_indices())
+    }
+}
+
+// ====================
+// ===== MUTATION =====
+// ====================
+#[uniffi::export]
+impl DeviceFactorAddingManager {
+    pub fn set_factor_name(self: Arc<Self>, name: String) -> Arc<Self> {
```
CyonAlexRDX, line 99: prefer `DisplayName` over `String`

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/add_factor/device_factor_adding_manager.rs
```diff
@@ -0,0 +1,151 @@
...
+impl DeviceFactorAddingManager {
+    pub fn is_word_at_index_correct(
+        self: Arc<Self>,
+        word: String,
+        index: u8,
+    ) -> bool {
+        self.wrapped.is_word_at_index_correct(word, index)
+    }
+
+    pub fn get_incorrect_confirmation_words(
```
CyonAlexRDX, line 136: `get_disconfirmed_words` ?

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/add_factor/device_factor_adding_manager.rs
sergiupuhalschi-rdx: The UI doesn't provide us a BIP39Word, but a String which might not be a valid bip39 word. This is a function to be called from the UI to validate what the user has introduced. Isn't it simpler to just compare a `String` with the mnemonic words at the specified index which would imply both that the word is a valid bip39 word and is also correct within the mnemonic?

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
danvleju-rdx: I think it’s required for `Hash`. Try removing `impl Eq`, and it should fail with an error related to unsatisfied trait bounds

sergiupuhalschi-rdx: Seems to be working fine after removing `impl Eq` both here and for the `SecurityShieldBuilder` as well.

danvleju-rdx: Compiling error [here](https://github.com/radixdlt/sargon/blob/main/crates/profile/models/security-structures/src/roles_matrices_structures/security_shield_builder.rs#L915) for me. 

>The function or associated item `from_iter` exists for struct `HashSet<SecurityShieldBuilder>`, but its trait bounds were not satisfied

crates/system/os/factors/src/add_factor_source/device_factor_source_adding_manager.rs
```diff
@@ -0,0 +1,360 @@
...
+            .add_new_factor_source(
+                self.get_factor_source_identification()
+                    .mnemonic_with_passphrase,
+                display_name,
+            )
+            .await
+    }
+}
+
+struct MockOsFactorAdder {
```
CyonAlexRDX, line 254: Annotate test only in **this crate** types and methods with `cfg(test)`

If you need it in another crate too use `cfg(debug_assertions)`

crates/system/os/factors/src/add_factor_source/device_factor_source_adding_manager.rs
```diff
@@ -0,0 +1,360 @@
...
+            indices
+        )
+    }
+
+    #[actix_rt::test]
+    async fn get_unconfirmed_words() {
+        let sut = SUT::sample();
+
+        let words_to_confirm: HashMap<usize, String> = vec![
+            (0, "word1".into()),
```
CyonAlexRDX, line 348: Prefer real world use case tests when possible:

* abandon
* about
* device
* sample 

All valid BIP39 words

crates/system/os/factors/src/add_factor_source/device_factor_source_adding_manager.rs
```diff
@@ -0,0 +1,360 @@
+use crate::prelude::*;
+
+pub struct DeviceFactorSourceAddingManager {
```
CyonAlexRDX, line 3: Missing doc

crates/system/os/factors/src/add_factor_source/device_factor_source_adding_manager.rs
```diff
@@ -0,0 +1,360 @@
...
+    factor_adder: Arc<dyn OsFactorSourceAdder>,
+    factor_source_identification: RwLock<Option<FactorSourceIdentification>>,
+    factor_source_name: RwLock<String>,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+struct FactorSourceIdentification {
```
CyonAlexRDX, line 10: Missing doc and I would put it in a separate file

crates/system/os/factors/src/add_factor_source/device_factor_source_adding_manager.rs
```diff
@@ -0,0 +1,360 @@
...
+                factor_source_name.as_ref().to_owned(),
+            ),
+        }
+    }
+}
+
+// ====================
+// ==== GET / READ ====
+// ====================
+impl DeviceFactorSourceAddingManager {
```
CyonAlexRDX, line 135: Do all methods which are pub need to be pub ?

jvm/sargon-android/src/test/java/com/radixdlt/sargon/os/factors/AddNewFactorManagerTest.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.radixdlt.sargon.os.interactor.FakeHostInteractor
+import com.radixdlt.sargon.samples.sample
+import io.mockk.mockk
+import kotlinx.coroutines.test.StandardTestDispatcher
+import kotlinx.coroutines.test.runTest
+import okhttp3.OkHttpClient
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+class AddNewFactorManagerTest {
```
CyonAlexRDX, line 32: `AddNewFactorSourceManagerTest`

crates/system/os/factors/src/add_factor_source/sargon_os_factor_source_adder.rs
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        self.update_profile_with(|p| {
+            p.factor_sources.append(factor_source.clone());
+            Ok(())
+        })
+        .await?;
+
+        if let Err(e) = self
+            .pre_derive_and_fill_cache_with_instances_for_factor_source(
```
CyonAlexRDX, line 74: hmm would be nice with `.inspect_err` on the Result from `pre_derive_and_fill_cache_with_instances_for_factor_source` to avoid the `if let Err` but I think maybe tricky due to async...?

crates/factors/factors/src/factor_source.rs
```diff
@@ -155,6 +157,61 @@ impl FactorSource {
...
+                )))
+            }
+            FactorSourceKind::Password => {
+                let hint = PasswordFactorSourceHint::new(display_name.value());
+                Ok(FactorSource::from(PasswordFactorSource::new(
+                    id_from_hash,
+                    hint,
+                )))
+            }
+            _ => Err(CommonError::InvalidFactorSourceKind {
```
Sajjon, line 208: I think best avoid `_` and spell out the variants with `|` i.e. `FactorSourceKind::Arculus | FactorSourceKind::Ledger ` etc so that we get a compilation error here when we add a new kind, reminding us to handle that new variant.

Sajjon: Yes we have used `_` in the past for `match` but I think many of those places ought to be changed to.

crates/system/os/factors/src/add_factor_source/sargon_os_factor_source_adder.rs
sergiupuhalschi-rdx: It would have been nice, but I didn't manage to make it work. Besides `await is only allowed inside async functions and blocks` error there is also the fact that the operation we perform in case of error returns a Result which we want to propagate.

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
sergiupuhalschi-rdx: I'm not sure this is a good idea anymore, because I got the tests failing a few times so I increased the value from 100 to 1000, but just now it failed again. Do you agree to revert to the previous code to avoid the tests occasionally failing?

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
sergiupuhalschi-rdx: This cascades into many changes and I'm not sure it's worth it, WDYT?

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/add_factor_source/device_mnemonic_builder.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+}
+
+// ====================
+// ===== MUTATION =====
+// ====================
+#[uniffi::export]
+impl DeviceMnemonicBuilder {
+    /// Generates a new mnemonic
+    pub fn create_new_mnemonic_with_passphrase(self: Arc<Self>) -> Arc<Self> {
```
Sajjon, line 130: I think this should be named:

`generate_new` not `create_new` I think.

sergiupuhalschi-rdx: What about `create_mnemonic_with_passphrase_from_words`? Should it be `generate_mnemonic_with_passphrase_from_words` for consistency?

CyonAlexRDX: Generate if generating random entropy. Create If using mnemonic words

CyonAlexRDX: So your last commit looks good

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/add_factor_source/device_mnemonic_builder.rs
```diff
@@ -0,0 +1,247 @@
...
+/// Exposes functions to be called by hosts to be able to use the resulting `MnemonicWithPassphrase`.
+#[derive(Debug, Hash, PartialEq, Clone, uniffi::Object)]
+#[uniffi::export(Hash, Eq)]
+pub struct DeviceMnemonicBuilder {
+    wrapped: Arc<InternalDeviceMnemonicBuilder>,
+}
+
+/// The result of the `build` function from `DeviceMnemonicBuilder`.
+#[derive(Debug, PartialEq, uniffi::Enum)]
+pub enum DeviceMnemonicBuildResult {
```
Sajjon, line 15: I would call this `DeviceMnemonicBuildOutcome`  - we try to always let `**Result` be a `Result` type.

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/add_factor_source/device_mnemonic_builder.rs
```diff
@@ -0,0 +1,247 @@
...
+    }
+}
+
+// ====================
+// ===== MUTATION =====
+// ====================
+#[uniffi::export]
+impl DeviceMnemonicBuilder {
+    /// Generates a new mnemonic
+    pub fn generate_new_mnemonic_with_passphrase(self: Arc<Self>) -> Arc<Self> {
```
CyonAlexRDX, line 129: maybe we should drop the `_with_passphrase` here and in `create_mnemonic_with_passphrase_from_words` since we do not in fact support BIP39Passphrases to be used.

crates/uniffi/uniffi_SPLIT_ME/src/system/sargon_os/add_factor_source/device_mnemonic_builder.rs
```diff
@@ -0,0 +1,247 @@
...
+        );
+
+        pretty_assertions::assert_eq!(
+            sut.clone().get_factor_source_id(),
+            FactorSourceID::sample_device()
+        );
+
+        let indices = sut.clone().get_indices_in_mnemonic_of_words_to_confirm();
+
+        let r0 = sut.clone().build(
```
CyonAlexRDX, line 203: `r` stands for? ah `result`? lets call it `let outcome`. And Rust supports shadowing! So no need for `outcome1` you can just `let outcome = // second` and again and again. since it looks like you dont need `r0` **and** `r1`?

CyonAlexRDX (approved):
LGTM, left some last minor remarks. Good job!

crates/system/os/factors/src/add_factor/device_factor_adding_manager.rs
sergiupuhalschi-rdx: Improved this by taking only the first N **unique** values from the randomly generated ones, so now statistically it shouldn't ever fail.

crates/system/os/factors/src/add_factor_source/device_mnemonic_builder.rs
```diff
@@ -0,0 +1,446 @@
+use crate::prelude::*;
+
+/// A builder of `MnemonicWithPassphrase` required for a new `DeviceFactorSource` creation.
+/// Exposes functions to be called by hosts to be able to use the resulting `MnemonicWithPassphrase`.
+#[derive(Debug)]
+pub struct DeviceMnemonicBuilder {
+    mnemonic_with_passphrase: RwLock<Option<MnemonicWithPassphrase>>,
```
GhenadieVP, line 7: Do we actually need the RwLock here? Is it expected to have many concurrent readers/writers

sergiupuhalschi-rdx: I don't think so, at least not according to the current requirements, but I think this is a good thing to have in place for the values accessed/modified by methods called from hosts directly. Do you think we should remove the lock?

GhenadieVP: Well it is no big impact having it here, it just adds additional complexity for no apparent need. It creates a sort of wrong expectation, that the state really needs to be protected. I would remove it possible, unless it is really needed.

GhenadieVP (changes requested):

crates/system/os/factors/src/add_factor_source/sargon_os_factor_source_adder.rs
```diff
@@ -0,0 +1,400 @@
...
+        .await?;
+
+        if factor_source_kind == FactorSourceKind::Device {
+            self.secure_storage
+                .save_mnemonic_with_passphrase(
+                    &mnemonic_with_passphrase,
+                    &factor_source.id_from_hash(),
+                )
+                .await?;
+        }
```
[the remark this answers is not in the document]
sergiupuhalschi-rdx: Good point. Or we could update the profile after saving the mnemonic to storage.

crates/system/os/factors/src/add_factor_source/sargon_os_factor_source_adder.rs
```diff
@@ -0,0 +1,427 @@
...
+                    &factor_source.id_from_hash(),
+                )
+                .await?
+        }
+
+        self.update_profile_with(|p| {
+            p.factor_sources.append(factor_source.clone());
+            Ok(())
+        })
+        .await?;
```
GhenadieVP, line 75: Same thing applies here - if profile update failed, the previously added device factor source should be removed from the secure storage.
Consider using a local scope to group together this and `self.pre_derive_and_fill_cache_with_instances_for_factor_source`, something like:
```rust
let setup_result = {
    self.update_profile_with(|p| {
            p.factor_sources.append(factor_source.clone());
            Ok(())
        })
        .await?;
        
    self
        .pre_derive_and_fill_cache_with_instances_for_factor_source(
            factor_source.clone(),
        )
        .await
};

if let Err(e) = setup_result {
   // revert logic on error
}
```

GhenadieVP (approved):

PR #386, opened 2025-02-24
Expose claim_profile API

GhenadieVP:
Expose claim_profile API so it can be used from the host side, this is an attempt to fix the wrong display of Wallet Transfer screen. ref -> https://rdxworks.slack.com/archives/C05V4202J90/p1739871801878009

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,7 +4,7 @@ version = 4
 
 [[package]]
 name = "account-for-display"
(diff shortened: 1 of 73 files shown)


CyonAlexRDX (approved):
LGTM

giannis-rdx (approved):

matiasbzurovski (approved):

PR #387, opened 2025-02-24
Fix spot check during entity creation

sergiupuhalschi-rdx:
When creating a new entity, the host has to be asked for a factor source "spot check" if there are enough factor instances pre-derived or, if not, to pre-derive and fill the factor instances cache. 

- This PR fixes the condition for the spot check when creating a new entity.
- Slightly improved the factor source adding API by returning the `FactorSourceId` from `add_new_mnemonic_factor_source` method.

> The changes were tested on Android

### DIFF
--- a/crates/system/os/factors/src/add_factor_source/sargon_os_factor_source_adder.rs
+++ b/crates/system/os/factors/src/add_factor_source/sargon_os_factor_source_adder.rs
@@ -12,7 +12,7 @@ pub trait OsFactorSourceAdder {
         factor_source_kind: FactorSourceKind,
         mnemonic_with_passphrase: MnemonicWithPassphrase,
         name: String,
-    ) -> Result<()>;
+    ) -> Result<FactorSourceID>;
 }
 
 #[async_trait::async_trait]
@@ -43,8 +43,9 @@ impl OsFactorSourceAdder for SargonOS {
         factor_source_kind: FactorSourceKind,
         mnemonic_with_passphrase: MnemonicWithPassphrase,
         name: String,
-    ) -> Result<()> {
+    ) -> Result<FactorSourceID> {
         let host_info = self.resolve_host_info().await;
+        let network_id = self.current_network_id()?;
         let factor_source = FactorSource::with_mwp(
             factor_source_kind,
             mnemonic_with_passphrase.clone(),
@@ -77,6 +78,7 @@ impl OsFactorSourceAdder for SargonOS {
 
             self.pre_derive_and_fill_cache_with_instances_for_factor_source(
                 factor_source.clone(),
+                network_id,
             )
             .await
         };
@@ -103,7 +105,7 @@ impl OsFactorSourceAdder for SargonOS {
             ))
             .await;
 
-        Ok(())
+        Ok(id)
     }
 }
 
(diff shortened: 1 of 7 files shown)


crates/system/os/factors/src/add_factor_source/sargon_os_factor_source_adder.rs
```diff
@@ -77,6 +78,7 @@ impl OsFactorSourceAdder for SargonOS {
 
             self.pre_derive_and_fill_cache_with_instances_for_factor_source(
                 factor_source.clone(),
+                network_id,
```
CyonAlexRDX, line 81: I think this is wrong. This way we do NOT pre derive on mainnet if current network is `Stokenet`. I think maybe best if you make `pre_derive_and_fill_cache_with_instances_for_factor_source` accept a set of networks and prederive on each network? 

NOTE that most end user do NOT use any other network than mainnet, but for those who do we should create on mainnet and on stokenet I think. 

N.B. the FactorInstancesProvider is optimized for readable /maintainable source code, not optimized for those select few users who are using stoknet, this means it works only with one network at a time. So if you pass `[Mainnet, Stoketnet]` then it will do 2 passes to the KeysCollector, not one, i.e. one per network. Which is a bad UX, but not a problem since most users do not use Stokenet.

CyonAlexRDX (changes requested):
should probably ALWAYS derive for mainnet, and if current network is not mainnet, we should pre derive for that network as well (but also mainnet).

CyonAlexRDX (approved):
resolved during SU.

crates/system/os/factors/src/apply_security_shield/sargon_os_apply_shield.rs
```diff
@@ -1150,6 +1150,16 @@ mod tests {
...
             let network = NetworkID::Mainnet;
+
+            let main_bdfs = os.main_bdfs().unwrap();
+            // The spot check is performed only when there are enough factor instances in cache
+            os.pre_derive_and_fill_cache_with_instances_for_factor_source(
+                main_bdfs.into(),
+                NetworkID::Mainnet,
+            )
+            .await
+            .unwrap();
```
matiasbzurovski, line 1161: this should be added to `test_create_and_save_new_account_continues_when_user_skips_spot_check`

crates/system/os/os/src/sargon_os_personas.rs
```diff
@@ -1333,6 +1333,15 @@ mod tests {
...
             .await;
 
+        let main_bdfs = os.main_bdfs().unwrap();
+        // The spot check is performed only when there are enough factor instances in cache
+        os.pre_derive_and_fill_cache_with_instances_for_factor_source(
+            main_bdfs.into(),
+            NetworkID::Mainnet,
+        )
+        .await
+        .unwrap();
```
matiasbzurovski, line 1343: This should be added to `test_create_and_save_new_persona_continues_when_user_skips_spot_check` too

matiasbzurovski (approved):

crates/system/os/factors/src/add_factor_source/sargon_os_factor_source_adder.rs
sergiupuhalschi-rdx: As discussed on SU, leaving as is for now, but we might apply the proposed solution in the future in combination with other measures like lowering the number of cached factor instances.

PR #388, opened 2025-02-24
ci: run ci on pr only when RunCI label [DO-3025]

marek-karwacki-rdx:


duje-begonja-rdx (approved):

.github/workflows/test.yml
```diff
@@ -40,6 +43,9 @@ jobs:
 
   # version bump
   check-version:
+    if: >
+      ( github.ref == 'refs/heads/main' && github.event_name == 'push' ) ||
```
CyonAlexRDX, line 47: right, we have an `if` per job, makes sense, and that makes me realize that you can switch the condition for **this job only** (`check-version`) to be only for OPEN PRs (i.e. remove `label` logic for it, and exclude it always for DRAFT).

PR #393, opened 2025-02-27
Off device mnemonic & password builders

sergiupuhalschi-rdx:


crates/system/os/factors/src/add_factor_source/password_mnemonic_builder.rs
```diff
@@ -0,0 +1,229 @@
...
+        if password.as_ref().len() < Self::MINIMUM_NUMBER_OF_CHARACTERS {
+            return Err(CommonError::PasswordTooShort);
+        }
+
+        if password.as_ref() != confirm_password.as_ref() {
+            return Err(CommonError::PasswordsDoNotMatch);
+        }
+
+        let entropy_bytes =
+            PasswordBasedKeyDerivationScheme::version1().kdf(password.as_ref());
```
sergiupuhalschi-rdx, line 104: Not sure about this. Do we need something else as per David's comment [here](https://radixdlt.atlassian.net/wiki/spaces/AT/pages/3758063620/MFA+Rules+for+Factors+and+Security+Shields?focusedCommentId=3805774308)?

PR #405, opened 2025-05-22
Deprecate DiscoverRadixDapps home card

micbakos-rdx:
* Remove this card when loaded
* Avoid storing it or leaking it to wallets
* The unify exported enum does not include this card.


micbakos-rdx:
Closing since there is a better way to handle removal of an old card.

PR #413, opened 2025-08-01
Arculus set and verify pin api

GhenadieVP:
- Add `set` and `verify pin` api to allow users to change the pin for the arculus card.
- Added API to extend crypto parameters.


sergiupuhalschi-radix (approved):
LGTM!

PR #449, opened 2026-06-04
Fix custom gateway URLs ignored for submit, preview, status, and apply-shield calls

timanrebel:
## Summary

- `gateway_client_on()` now reads the current saved `Gateway` (including custom URL) instead of deriving a hardcoded default from `network_id`
- `gateway_client_with(network_id)` looks up saved gateways to find the matching custom gateway before falling back to the default
- Apply-shield builders snapshot `SavedGateways` at construction so all epoch/balance requests use the correct gateway URL
- Transaction analysis, status, and submit tests verify that the current custom gateway URL is used

## No breaking changes

All fixes are internal to Sargon. No public API signatures change; no host-side (iOS/Android) updates required.

## Test plan

- [ ] `cargo nextest run -p sargon-os` — new tests `gateway_client_uses_current_custom_gateway_url` and `gateway_client_with_uses_current_custom_gateway_for_network` pass
- [ ] `cargo nextest run -p sargon-os-transaction` — new tests for analysis, status, submit with custom gateway pass
- [ ] `cargo check --workspace` clean


timanrebel:
A follow-up PR is coming that fixes the remaining call sites: HomeCardsManager and RadixNameService. That change is a breaking change. The public uniffi constructors will take a Gateway instead of a NetworkID, so it will need companion PRs on both `babylon-wallet-ios` and `babylon-wallet-android` to pass the user's current configured gateway instead of deriving a default from the network ID. Keeping this PR separate so the non-breaking fixes can land independently.

GhenadieVP:
Thanks you @timanrebel for the contribution, closing this PR as a fix was done with https://github.com/radixdlt/sargon/pull/451

